PR #8, opened 2022-09-13
Simplify the RFC process

chadwhitacre:
The current RFC process is confusing because we have two state machines operating independently: GitHub's PR workflow, and the RFC status-based workflow.

I propose that we simplify the RFC process by editing documents directly on `main`, and only moving to PRs once people start to care about it. Of course we would require a PR to change the status of a non-informational RFC to `approved`.

- "request" = push to `main`
- "comments" = issues or PRs against the doc on `main`

If this is approved and merged then the next step would be for current PR authors to merge or close their PRs.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/cwlw/simplify-process/text/0001-workflow.md)

### DIFF
--- a/text/0001-workflow.md
+++ b/text/0001-workflow.md
@@ -31,7 +31,7 @@ There are three types of RFCs:
 
 # Workflow
 
-1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
+1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` (number it one greater than the current max) and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
 1. Modify your RFC directly on `main` until other people care about it, then switch to PRs. PRs (or Issues, etc.) are the "comments" in RFC.
 1. For non-informational RFCs, updating the RFC status to `approved` _must_ happen in a PR, with relevant parties signing off.
 
--- a/text/0001-workflow.md
+++ b/text/0001-workflow.md
@@ -31,9 +31,9 @@ There are three types of RFCs:
 
 # Workflow
 
-1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time.
-1. Modify your RFC directly on `main` or through PRs as you see fit, except that ...
-1. For non-informational RFCs, updating the RFC status to `approved` must happen in a PR, with relevant parties signing off.
+1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
+1. Modify your RFC directly on `main` until other people care about it, then switch to PRs. PRs (or Issues, etc.) are the "comments" in RFC.
+1. For non-informational RFCs, updating the RFC status to `approved` _must_ happen in a PR, with relevant parties signing off.
 
 If you are writing a DACI style RFC, read "Instructions for running this Play" (10 mins) from
 [Atlassian's Playbook](https://www.atlassian.com/team-playbook/plays/daci).  Mention informed and contributors in the RFC and have approver sign off on the `approved` PR.
--- a/text/0001-workflow.md
+++ b/text/0001-workflow.md
@@ -12,15 +12,15 @@ This RFC describes the Sentry RFC process.
 This document exists so that future RFC creators and editors have a workflow to go by.  This workflow is inspired by
 common RFC processes from the Open Source community (Rust RFCs, Python PEPs) but also our internal use of DACIs.
 
-# RFC Type
+# Type
 
-The are three kinds of RFCs:
+There are three types of RFCs:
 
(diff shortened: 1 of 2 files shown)


text/0001-workflow.md
```diff
@@ -29,32 +29,21 @@ The are three kinds of RFCs:
...
+# Workflow
 
-1. Create a new RFC document as a copy of `0000-template.md`.
-2. Name it `XXXX-short-description.md` and commit it to a new branch.
-3. Create a pull request against the `rfcs` repository.  The number of the pull request then
-   becomes the assigned RFC number filled into `XXXX`.  Zero pad it out to 4 places for better sorting.
-4. Pick an RFC type and write it down on your RFC text into the header.
+1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time.
+1. Modify your RFC directly on `main` or through PRs as you see fit, except that ...
+1. For non-informational RFCs, updating the RFC status to `approved` must happen in a PR, with relevant parties signing off.
```
markstory, line 36: For RFCs that are made on main, how would the 'comments' be done? One strength of having the entire RFC show up as a pull request is the workflow that github provides around providing feedback and incorporating that feedback.

chadwhitacre: I guess the mental model is: publishing on main is the "request", and subsequent PRs are the "comments." How about this?

```suggestion
1. Modify your RFC directly on `main` until other people care about it, then switch to PRs to track comments and changes.
1. For non-informational RFCs, updating the RFC status to `approved` _must_ happen in a PR, with relevant parties signing off.
```

chadwhitacre: > subsequent PRs

... or Issues, or Discussions, or Discord, or Discord Forums, or ... 😭

text/0001-workflow.md
```diff
@@ -29,32 +29,21 @@ The are three kinds of RFCs:
...
 
-# RFC Creation Workflow
+# Workflow
 
-1. Create a new RFC document as a copy of `0000-template.md`.
-2. Name it `XXXX-short-description.md` and commit it to a new branch.
-3. Create a pull request against the `rfcs` repository.  The number of the pull request then
-   becomes the assigned RFC number filled into `XXXX`.  Zero pad it out to 4 places for better sorting.
-4. Pick an RFC type and write it down on your RFC text into the header.
+1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time.
```
chadwhitacre, line 34:
```suggestion
1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
```

text/0001-workflow.md
chadwhitacre: Right now PRs that add a new RFC are requests for comments about requests for comments. I propose we unnest. :)

chadwhitacre:
Rebased post-#9, addressed some review feedback as well.

mitsuhiko:
These changes look good to me and I would like to accept these.

mitsuhiko:
I'm actually not sure now if this is a good model as it also means we no longer have the RFC number allocation which currently comes from the PR number. More importantly it also I believe is more complex in practice as you might now have multiple states in which the whole RFC is. Does the process actually get simpler by editing on `main`?

chadwhitacre:
> RFC number allocation which currently comes from the PR number.

This would simply be current_max + 1, yes? Race condition is unlikely and easy to recover from (bump one of them).

> More importantly it also I believe is more complex in practice as you might now have multiple states in which the whole RFC is.

Not sure I understand. What would the multiple states be? The state on `main` would be the only state, no?

text/0001-workflow.md
```diff
@@ -29,32 +29,21 @@ The are three kinds of RFCs:
...
 
-# RFC Creation Workflow
+# Workflow
 
-1. Create a new RFC document as a copy of `0000-template.md`.
-2. Name it `XXXX-short-description.md` and commit it to a new branch.
-3. Create a pull request against the `rfcs` repository.  The number of the pull request then
-   becomes the assigned RFC number filled into `XXXX`.  Zero pad it out to 4 places for better sorting.
-4. Pick an RFC type and write it down on your RFC text into the header.
+1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
```
chadwhitacre, line 34:
```suggestion
1. Create a new RFC document as a copy of `0000-template.md`. Name it `XXXX-short-description.md` (number it one greater than the current max) and commit it directly to the `main` branch. Anyone can do this at any time. Publishing a new doc to `main` is the "request" in RFC.
```

chadwhitacre:
> The current RFC process is confusing because we have two state machines operating independently: GitHub's PR workflow, and the RFC status-based workflow.

Alternative solution would be to drop the status field from the RFC template and use GH PR state exclusively. If it's merged it's accepted, if it's not merged it's still in draft, if it's closed it's withdrawn.

Not sure what to do with info type in that case (real talk tho how are "info RFCs" different than "documentation"?).

mitsuhiko:
We can remove the state entirely except for `replaced`. I want to ensure that someone who looks at an RFC in the repo knows if it still is relevant or not.

chadwhitacre:
Okay, in that case I will close this PR and try to get around to opening another one if I can. :)

PR #9, opened 2022-09-14
list the rfc types in the template

asottile-sentry:
the "see below" was lost it seems -- adding the types here makes this slightly easier

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -1,5 +1,5 @@
 * Start Date: YYYY-MM-DD
-* RFC Type: see below
+* RFC Type: feature / decision / informational
 * RFC PR: <link>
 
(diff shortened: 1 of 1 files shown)


asottile-sentry:
it looks like #8 also addresses this, so perhaps that is the better PR

markstory (commented):
I like this, the 'see below' part was not clear to me. :+1:

chadwhitacre (approved):

chadwhitacre:
Ditto, happy to see this land asap as #8 is more complex and this is an obvious win.

asottile-sentry:
I'll land this now then -- sorry for the conflict @chadwhitacre !

Issue #1, opened 2022-07-21
RFC Workflow Discussion


mitsuhiko:
This is an experimental repository to see if we can make our DACI process public. Currently we use internal notion documents to track a lot of our decisions. We are an open source company and really there shouldn't be a reason why the majority of these decisions can't be out in the open.

This repository is not yet active and no decisions has been made about this approach. This issue is a meta issue to discuss how we can structure this process so that it's as helpful as possible for all associated parties.

I added `0001-workflow` not through a PR and as such I would propose we discuss this item here instead.

mwarkentin:
@mitsuhiko I linked here from the README as I don't think it was too clear where the discussion was on why we are considering moving here from our existing DACI documents.

I love the idea - I've really enjoyed browsing some of the [UK Government's ADRs](https://github.com/alphagov/govuk-aws/tree/main/docs/architecture/decisions) on their open-source AWS infra repo.

Along those lines it would probably be interesting to port over a bunch of our existing DACI documents.. it would probably help figure out what sort of documentation (if any) is likely to cause challenges with exposing publicly.

chadwhitacre:
Why is something like https://github.com/getsentry/rfcs/pull/6 an RFC vs. an issue in https://github.com/getsentry/sentry-cocoa? Seems like RFCs should be for cross-cutting or quite complex problems.

bruno-garcia:
@chadwhitacre the options affect not only the SDK, like ingestion/Relay and we'd like to get more ppl involved. Hence the decision to write here. Should re revert the PR?

chadwhitacre:
Ah, okay. Do we want that context in about the impact beyond the SDK the PR or the RFC itself? It's surely obvious to those in the know but I think part of the exercise here with public RFCs is getting better at documenting things in a way that scales beyond small tribal knowledge groups.

PR #10, opened 2022-09-17
Introduce additional types of exceptions next to `unhandled` exceptions

ueman:
This RFC suggests a feature which introduces additional types of exceptions next to `unhandled`.

Currently, exception which cause the running software to exit are marked as `unhandled: true`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit. The aim of this RFC is to fix that.

[Rendered RFC](https://github.com/ueman/sentry-rfcs/blob/main/text/exception-types.md)

### DIFF
--- a/text/exception-types.md
+++ b/text/exception-types.md
@@ -23,7 +23,7 @@ Based on the problem stated above, I propose to introduce the types of `manually
 
 - `manually caught`: The exception was recorded by a developer. May or may not be visually indicated by the Sentry user interface.
 - `uncaught`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception however didn't cause the software to quite, and the software will continue to be executed. This should be visualized in the Sentry user interface.
-- `software crashed`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface.
+- `software crashed`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface. This is currently done by the unhandled flag in the [exception mechanism](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism).
 
 This enables the consideration in the `session health` metric.
 
--- a/text/exception-types.md
+++ b/text/exception-types.md
@@ -1,0 +1,44 @@
+* Start Date: 2022-09-14
+* RFC Type: feature
+* RFC PR: -
+* RFC Status: draft
+
+# Summary
+
+This RFC suggests a feature which introduces additional types of exceptions next to `unhandled`.
+
+# Motivation
+
+Currently, exception which cause the running software to exit are marked as `unhandled: true`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit.
+
+Sentry shows exceptions which aren't caught but also do not cause the software to exit in the same way as exceptions which are manually caught by a developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
+
+This issue impacts, for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
+
+Another issue is, that excpetions which don't cause the application to exit but are uncaught, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
+
+# Proposal
+
+Based on the problem stated above, I propose to introduce the types of `manually caught`, `uncaught`, `software crashed` (this is the same as the current `unhandeld`, but rephrased to avoid confusion). I'm open for better phrasing of those types, but I'll stick to those names for the rest of the RFC. The meaning of those types is as follows:
+
+- `manually caught`: The exception was recorded by a developer. May or may not be visually indicated by the Sentry user interface.
+- `uncaught`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception however didn't cause the software to quite, and the software will continue to be executed. This should be visualized in the Sentry user interface.
+- `software crashed`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface.
+
+This enables the consideration in the `session health` metric.
+
+I'm guessing this affects data ingestion layer, but since I'm not familiar with that part, I can't comment on the impact theses changes would have on that.
+
+# Other options considered
+
+Mark both `uncaught` and `software crashed` as the same type.
+That would however make it impossible to differentiate between those exception types in the `session health` metric.
+
+# Approaches taken by other monitoring tools
+
+- Crashlytics does not differentiate between `uncaught` and `software crashed`.
+
+# Unresolved Questions
+
+- Which SDKs would profit from this RFC? Is it the majority, a good chunk of it, or just the minory?
+- Are there any other exception types next to the ones metioned in this RFC?
--- a/text/exception-types.md
+++ b/text/exception-types.md
@@ -13,30 +13,61 @@ Currently, exception which cause the running software to exit are marked as `unh
 
 Sentry shows exceptions which aren't caught but also do not cause the software to exit in the same way as exceptions which are manually caught by a developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
 
-This issue impacts, for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
+This issue impacts for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
 
 Another issue is, that excpetions which don't cause the application to exit but are uncaught, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
 
 # Proposal
 
-Based on the problem stated above, I propose to introduce the types of `manually caught`, `uncaught`, `software crashed` (this is the same as the current `unhandeld`, but rephrased to avoid confusion). I'm open for better phrasing of those types, but I'll stick to those names for the rest of the RFC. The meaning of those types is as follows:
+Based on the problem stated above, I propose to introduce the types of `handled`, `unhandled`, `process termination` (this is the same as the current `unhandeld`, but rephrased to avoid confusion). I'm open for better phrasing of those types, but I'll stick to those names for the rest of the RFC. The meaning of those types is as follows:
 
-- `manually caught`: The exception was recorded by a developer. May or may not be visually indicated by the Sentry user interface.
-- `uncaught`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception however didn't cause the software to quite, and the software will continue to be executed. This should be visualized in the Sentry user interface.
-- `software crashed`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface. This is currently done by the unhandled flag in the [exception mechanism](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism).
+- `handled`: The exception was recorded by a developer via `Sentry.capture*` method. May or may not be visually indicated by the Sentry user interface.
+- `unhandled`: Indicates whether the exception was recorded automatically by Sentry through the use of a global exception handler or similar. This exception however didn't cause the software to exit, and the software will continue to be executed. This should be visualized in the Sentry user interface.
+- `process termination`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface. This is currently done by the `unhandled` flag in the [exception mechanism](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism).
 
-This enables the consideration in the `session health` metric.
+In order to propagate those exception types, the exception mechanism needs to be adapted:
+
+```json
+{
+  "exception": {
+    "values": [
+      {
+        "type": "Error",
+        "value": "An error occurred",
+        "mechanism": {
+          "type": "generic",
+          "handled": true,
+          "process_terminated": false // <--- newly introduced field
+        }
+      }
+    ]
+  }
+}
+```
+
+In order to achieve backwards compatibility, in the absence of the `process_termination` flag, the current behavior stays as is.
+As soon as the `process_terminated` flag is present the bavior is as follows:
+
+- `handled = true` and `process_terminated = true`: Software was gracefully shut down after an handled exception
+- `handled = true` and `process_terminated = false`: Exception is not handled by the user but didn't cause the software to terminate. Same as `unhandled` in the list above
+- `handled = false` and `process_terminated = true`: Software terminated after an unhandled exception. Same as `process termination` in the list above
+- `handled = true` and `process_terminated = false`: Exception was reported via `Sentry.capture*()` method. Same as `handled` in the list above.
+
+In the absence of the `handled` or its value being null, it's assumed to be `handled = true`.
+
+
+The introduction of the `process_terminated` flag enables the consideration of such exception types in the `session health` metric.
 
 I'm guessing this affects data ingestion layer, but since I'm not familiar with that part, I can't comment on the impact theses changes would have on that.
 
 # Other options considered
 
-Mark both `uncaught` and `software crashed` as the same type.
+Unhandled exceptions, which don't cause a process termination, are considered like exceptions which cause the process to terminate.
 That would however make it impossible to differentiate between those exception types in the `session health` metric.
 
 # Approaches taken by other monitoring tools
 
-- Crashlytics does not differentiate between `uncaught` and `software crashed`.
+- Crashlytics just differentiates between manually caught exceptions and unhandled exceptions, regardless of wether they cause the process to terminate.
 
 # Unresolved Questions
 


text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
+* Start Date: 2022-09-17
+* RFC Type: feature
+* RFC PR: 0010
+* RFC Status: draft
+
+# Summary
+
+This RFC suggests a feature which introduces additional types of exceptions next to `unhandled`.
```
marandaneto, line 8:
```suggestion
This RFC suggests a feature which introduces additional types of exceptions next to `mechanism.handled`.
```

Flag `unhandled` does not exist so it's confusing, I've updated the description and issue's title as well.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Motivation
+
+Currently, exception which cause the running software to exit are marked as `unhandled: true`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit.
```
marandaneto, line 12:
```suggestion
Currently, exception which cause the running software to exit are marked as `handled: false`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit.
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+Sentry shows exceptions which aren't caught but also do not cause the software to exit in the same way as exceptions which are manually caught by a developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
```
marandaneto, line 14:
```suggestion
Sentry shows exceptions which aren't caught by the developer but also do not cause the software to exit in the same way as exceptions which are manually caught by the developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+This issue impacts for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
+
+Another issue is, that excpetions which don't cause the application to exit but are uncaught, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
```
marandaneto, line 18:
```suggestion
Another issue is, that excpetions which don't cause the application to exit but are unhandled, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
Currently, the session would be marked as `errored` instead of `crashed`.
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Proposal
+
+Based on the problem stated above, I propose to introduce the types of `handled`, `unhandled`, `process termination` (this is the same as the current `unhandeld`, but rephrased to avoid confusion). I'm open for better phrasing of those types, but I'll stick to those names for the rest of the RFC. The meaning of those types is as follows:
```
marandaneto, line 22:
```suggestion
Based on the problem stated above, I propose to introduce the types of `handled`, `unhandled`, `process termination` (this is the same as the current `handled`, but rephrased to avoid confusion). I'm open for better phrasing of those types, but I'll stick to those names for the rest of the RFC. The meaning of those types is as follows:
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+- `handled`: The exception was recorded by a developer via `Sentry.capture*` method. May or may not be visually indicated by the Sentry user interface.
+- `unhandled`: Indicates whether the exception was recorded automatically by Sentry through the use of a global exception handler or similar. This exception however didn't cause the software to exit, and the software will continue to be executed. This should be visualized in the Sentry user interface.
+- `process termination`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface. This is currently done by the `unhandled` flag in the [exception mechanism](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism).
```
marandaneto, line 26:
```suggestion
- `process termination`: The exception was recorded automatically by Sentry through the use of a exception handler or similar. The exception caused the software to terminate the execution. This should be visualized in the Sentry user interface. This is currently done by the `handled` flag in the [exception mechanism](https://develop.sentry.dev/sdk/event-payloads/exception/#exception-mechanism).
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+      }
+    ]
+  }
+}
+```
+
+In order to achieve backwards compatibility, in the absence of the `process_termination` flag, the current behavior stays as is.
+As soon as the `process_terminated` flag is present the bavior is as follows:
+
+- `handled = true` and `process_terminated = true`: Software was gracefully shut down after an handled exception
```
marandaneto, line 51: I believe this will never happen, if it's handled, the application won't terminate.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+- `handled = true` and `process_terminated = false`: Exception is not handled by the user but didn't cause the software to terminate. Same as `unhandled` in the list above
```
marandaneto, line 52:
```suggestion
- `handled = false` and `process_terminated = false`: Exception is not handled by the user but didn't cause the software to terminate. Same as `unhandled` in the list above
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+- `handled = false` and `process_terminated = true`: Software terminated after an unhandled exception. Same as `process termination` in the list above
+- `handled = true` and `process_terminated = false`: Exception was reported via `Sentry.capture*()` method. Same as `handled` in the list above.
+
+In the absence of the `handled` or its value being null, it's assumed to be `handled = true`.
```
marandaneto, line 56: That's already the current behavior.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+The introduction of the `process_terminated` flag enables the consideration of such exception types in the `session health` metric.
+
+I'm guessing this affects data ingestion layer, but since I'm not familiar with that part, I can't comment on the impact theses changes would have on that.
```
marandaneto, line 61: I don't think so, because clients (SDKs) are responsible for sending the `status` of the sessions, the SDKs check the type of the exception and mark them as such.
The question is if we want to introduce a new category for release health, which differentiates `handled`, `unhandled`, and `application exist`.
Right now there are only 2 categories, `errored` and `crashed`.
`crashed` meaning `process termination`.
`errored` meaning `handled` and `unhandled`.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Other options considered
+
+Unhandled exceptions, which don't cause a process termination, are considered like exceptions which cause the process to terminate.
+That would however make it impossible to differentiate between those exception types in the `session health` metric.
```
marandaneto, line 66: Some SDKs do that, but they are languages/platforms that never cause the application to exist, such as the Browser.
Unhandled JS events are marked as `handled: false` so sessions are `crashed`, but the browser never exit.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Approaches taken by other monitoring tools
+
+- Crashlytics just differentiates between manually caught exceptions and unhandled exceptions, regardless of wether they cause the process to terminate.
```
marandaneto, line 70: The problem here is that you cannot give a higher priority to events that cause the application to exit.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Unresolved Questions
+
+- Which SDKs would profit from this RFC? Is it the majority, a good chunk of it, or just the minory?
```
marandaneto, line 74: Likely every language/platform/framework that either does not exit at all (such as the Browser SDK), or when they swallow the exception such as Flutter and a few other technologies.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+- Are there any other exception types next to the ones metioned in this RFC?
```
marandaneto, line 75: I guess this RFC covers the 3 known types.
Manually captured by the dev.
Automatically captured via error handlers, not causing the app to exit.
Automatically captured via error handlers, causing the app to exit.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+
+Currently, exception which cause the running software to exit are marked as `unhandled: true`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit.
+
+Sentry shows exceptions which aren't caught but also do not cause the software to exit in the same way as exceptions which are manually caught by a developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
+
+This issue impacts for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
+
+Another issue is, that excpetions which don't cause the application to exit but are uncaught, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
+
+# Proposal
```
marandaneto, line 20: I'd add 2 more items to the proposal:

A way to filter events on the issues page or discover, such as:

`handled:yes`, `handled:no` are already possible.
Introducing a new one which filters the `process termination` type, eg `handled:termination` or a brand new tag, that's just an example.

The other 2 is the UI label highlight for `unhandled` events and `process termination`.
Currently, there's an `unhandled` label on the issue's page but it's only highlighted for `process termination` errors.

text/0010-exception-types.md
```diff
@@ -0,0 +1,75 @@
...
+- `handled = true` and `process_terminated = false`: Exception was reported via `Sentry.capture*()` method. Same as `handled` in the list above.
+
+In the absence of the `handled` or its value being null, it's assumed to be `handled = true`.
+
+
+The introduction of the `process_terminated` flag enables the consideration of such exception types in the `session health` metric.
+
+I'm guessing this affects data ingestion layer, but since I'm not familiar with that part, I can't comment on the impact theses changes would have on that.
+
+# Other options considered
```
marandaneto, line 63: Usually, an RFC consists of different options, one of them being the recommended one.

So I'd just rename them to `Option #1 (recommended), `Option #2`, etc...

marandaneto, line 63: I'd add an option which is to deprecate the `handled` flag and introduce a new one which is an `enum` with the 3 known types.
We'd send the `handled` flag still for a while.
Ingestion, SDKs and UI would use this new flag in case it exists otherwise it uses the current `handled` flag because of back compatibility.

marandaneto: I'd add another option which is setting `handled: false` for unhandled errors that don't cause the application to crash.
Sessions would be marked as `crashed`.
No changes would be required on the ingestion and UI, but obviously, you'd not be able to differentiate such events.
Maybe that would be better than now, but I agree it's not the best case.

marandaneto:
Just for context: that were my notes about this feature back then.
This RFC addresses all of them.

Screenshot 2022-09-18 at 09 40 23

text/0010-exception-types.md
ueman: I just added this to cover all 4 possible cases. I guess if it happens, it's the developer setting it manually.

marandaneto: Gotcha, but it's an invalid use case, I'd rename it as "invalid" or even remove it, otherwise it's just more information to digest that does not represent a real use case.

text/0010-exception-types.md
ueman: I added a note for that.

text/0010-exception-types.md
ueman: Removed this sentence

text/0010-exception-types.md
ueman: Added a list of known SDKs which would profit from this RFC

text/0010-exception-types.md
ueman: Added ✅

text/0010-exception-types.md
ueman: Done ✅

ueman: Added ✅

text/0010-exception-types.md
```diff
@@ -0,0 +1,115 @@
...
+- Crashlytics just differentiates between manually caught exceptions and unhandled exceptions, regardless of wether they cause the process to terminate.
+
+# List of SDK to which this applies
+
+This list might be incomplete
+
+- [Flutter](https://github.com/getsentry/sentry-dart/issues/456)
+- Browser SDKs
+- Unity
+- React Natice
```
marandaneto, line 111:
```suggestion
- React Native
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,115 @@
...
+        }
+      }
+    ]
+  }
+}
+```
+
+If the currently available `handled` flag is also present, the `exception_type` flag takes precedence. The `handled` flag however should become deprecated.
+
+The introduction of the `exception_type` flag enables the consideration of such exception types in the `session health` metric.
```
marandaneto, line 93: The release health protocol would need to change as well, because right now we only have `crashed` and `errored`
https://docs.sentry.io/product/releases/health/#session-status
but we need to represent 3 states.

marandaneto:
Relates to https://github.com/getsentry/rfcs/pull/15 as well.

mattjohnsonpint:
LGTM.  I like option 1 because of the backwards-compatibly.

Definitely applies for .NET and .NET MAUI as well.

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+
+# Summary
+
+This RFC suggests a feature which introduces additional types of exceptions next to `mechanism.handled`.
+
+# Motivation
+
+Currently, exception which cause the running software to exit (the process died/hard crash) are marked as `handled: false`. This isn't enough for SDKs where an exception can be unhandled but at the same time doesn't cause the software to exit.
+
+Sentry shows exceptions which aren't caught by the developer (unhandled) but also do not cause the software to exit in the same way as exceptions which are manually caught by the developer. This seems rather unintuitive and makes exceptions seem less severe than they are.
```
bruno-garcia, line 14: this seem inaccurate to me. Or maybe I'm not understanding.

Sentry doesn't show exceptions that are not captured by the user (unhandled) in the same way it shows exceptions  exceptions manually captured (handled). One has the red banner:
image

Do you mean that Sentry shows all unhandled errors in the same way, independent if they caused the app to crash?

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+
+This issue impacts for example the [Dart/Flutter SDK](https://github.com/getsentry/sentry-dart/issues/456), the Unity SDK, the React Native SDK and possibly more.
+
+Another issue is, that excpetions which don't cause the software to exit but are unhandled, are not considered in the [`session health` metric](https://develop.sentry.dev/sdk/sessions/).
```
bruno-garcia, line 18: What do you mean by 'they are not considered'?

My understanding is that we mark a session as crashed for all unhandled errors. For example, on Browser, we finish the session as crashed if there's an unhandled error, even though there was no crash.

Is this referring to Flutter where in the event of a Dart unhandled exception we don't close the session as crashed? But if there's a native unhandled error (which terminates the app) we do?

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+Currently, the session would be marked as `errored` instead of `crashed`.
+
+The attribute `thread.errored` was added in the past for similar reasons, but it got [reverted](https://github.com/getsentry/relay/pull/306) for some reasons.
```
bruno-garcia, line 21:
```suggestion
The attribute `thread.errored` was added in the past for similar reasons, but it got [reverted on Relay](https://github.com/getsentry/relay/pull/306) and [Android](https://github.com/getsentry/sentry-android/pull/187).
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+          "type": "generic",
+          "exception_type": "handled|unhandled|process_termination", // <--- newly introduced field
+        }
+      }
+    ]
+  }
+}
+```
+
+If the currently available `handled` flag is also present, the `exception_type` flag takes precedence. The `handled` flag however should become deprecated.
```
bruno-garcia, line 95: This flag is how users current filter and set alerts. Deprecating this feel like a big change

text/0010-exception-types.md
bruno-garcia: It seems to me we don't really need to change Sessions. We can just go from errroed to crashed in case process_exit=true instead of handled:false

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+The introduction of the `exception_type` flag enables the consideration of such exception types in the `session health` metric.
+
+# Option 3
+
+Unhandled exceptions, which don't cause a process termination, are considered like exceptions which cause the process to terminate and are marked `handled: false`
+That would however make it impossible to differentiate between those exception types in the `session health` metric, filters, alerts, etc.
+
+# Approaches taken by other monitoring tools
+
+- Crashlytics just differentiates between manually caught exceptions and unhandled exceptions, regardless of wether they cause the process to terminate.
```
bruno-garcia, line 106:
```suggestion
- Crashlytics just differentiates between manually caught exceptions and unhandled exceptions, regardless of whether they cause the process to terminate.
```

text/0010-exception-types.md
```diff
@@ -0,0 +1,119 @@
...
+
+# List of SDK to which this applies
+
+This list might be incomplete
+
+- [Flutter](https://github.com/getsentry/sentry-dart/issues/456)
+- Browser SDKs
+- Unity
+- React Native
```
bruno-garcia, line 115:
```suggestion
- React Native
- .NET (UnobservedTaskException)
```

text/0010-exception-types.md
ueman: Unhandled exceptions in Flutter don't have that red banner because they don't crash the application. I'm not sure how similar behavior is implemented in other SDKs.

That means there's no visual differentiation between handled and unhandled exceptions as long as they don't cause the software to exit.

> Do you mean that Sentry shows all unhandled errors in the same way, independent if they caused the app to crash?

No, unhandled exceptions which cause the software to exit are shown visually different.

text/0010-exception-types.md
ueman: > My understanding is that we mark a session as crashed for all unhandled errors. For example, on Browser, we finish the session as crashed if there's an unhandled error, even though there was no crash.

That probably wouldn't be the case for Flutter, if it had sessions, based on the current behavior. Even for Flutter for Web.

> Is this referring to Flutter where in the event of a Dart unhandled exception we don't close the session as crashed? But if there's a native unhandled error (which terminates the app) we do?

Yes, exactly.

Based on your questions, I'm assuming that `crashed`/`unhandled` mean different things across different SDKs. That's one more reason to align across SDKs.

text/0010-exception-types.md
marandaneto: Right now, Handled and Unhandled exceptions are `handled: null` or `handled: true` so there's no red banner, no way to filter or alert them differently.
Only when it crashes on the Native side (C code or Native threads), the flag is `handled: false`.
So yes, it's bad and the RFC tries to fix this.

text/0010-exception-types.md
marandaneto: > Based on your questions, I'm assuming that crashed/unhandled mean different things across different SDKs. That's one more reason to align across SDKs.

Yes, this RFC tries to align this inconsistency as well @ueman .

Sessions are only marked as crashes when it cashes on the Native side.

lobsterkatie:
@ueman Thank you for writing this up! It's been on my list for a long time, because it's also decidedly broken in JS, in a number of different ways.

On the browser side of things:
- True crashes (the kind that land you [here](chrome://crash/)) are pretty rare. That's fortunate, because at that point the JS engine has gone down in flames (taking Sentry along with it) so we have no good way to record what's happened. (There's no publicly-available native browser API to capture such an event the way you can with, say, a minidump.)
- Nonetheless, we mark some errors as unhandled, specifically those which bubble up to the [global handlers](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/globalhandlers.ts#L245). While it's true they haven't been handled (unless the user has add a global `onerror` handler themselves, which is not something we currently try to detect), they're also not crashes. (They show up in release health metrics that way, though.)
- Errors which are caught by our auto-instrumentation ([1](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/trycatch.ts#L117), [2](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/trycatch.ts#L138), [3](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/trycatch.ts#L164), [4](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/trycatch.ts#L224), [5](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/integrations/trycatch.ts#L243)) are marked as handled, even though they have not, in fact, been handled by the user.
- Events which _are_ handled by the user are correctly [marked as being handled](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/browser/src/eventbuilder.ts#L157).

So the data is wrong in many cases, but it's also kind of arbitrary and divorced from reality, in both directions: It's totally possible for some browser extension the user has installed to be blowing up the console with unhandled errors, causing all of the sessions in the main web app to be marked as crashed, even though the user is blissfully unaware that anything's wrong (and the erroring code has nothing to do with the app integrating sentry). It's also possible for, say, a click handler to error, the site to therefore become unresponsive to at least some user interactions, and for our auto-instrumentation to catch it and mark it handled (and therefore a healthy session), even though the user's experience is of an app which has frozen up.

On the node side, we have both the same (too much `handled`) and the opposite (too much `unhandled`) problem:
- True crashes (anything which causes the main process to exit) do happen, but [only sometimes trigger the `beforeExit` event](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/sdk.ts#L262-L274). We also sometimes [purposefully exit the process](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/integrations/utils/errorhandling.ts#L11-L37).
- Similar to the browser SDK, the node SDK marks [unhandled errors](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/integrations/onuncaughtexception.ts#L83) and [rejections](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/integrations/onunhandledrejection.ts#L54) which make it to the global handler as unhandled, but depending on `Sentry.init` settings, this [may](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/integrations/onuncaughtexception.ts#L58-L67) or [may not ](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/integrations/onunhandledrejection.ts#L74-L79) result in the process being force-exited. Also similar to browser-land, all other errors [are marked handled](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/node/src/eventbuilder.ts#L54-L57), whether they were caught by the user or by auto-instrumentation.
- If the node SDK is being run inside the serverless SDK, though, [all errors are considered unhandled](https://github.com/getsentry/sentry-javascript/blob/584c4cbebde327123e9d97728b696a4872a6d248/packages/serverless/src/utils.ts#L12-L18), regardless of whether or not the user catches the error.

Ultimately, there are really three problems/challenges/considerations/whatever that we're dealing with:

- We'd like some way for the data model to distinguish between true crashes and unhandled errors (the main subject of this RFC).
- As has been alluded to in other comments, making changes here has consequences for both the UI (the red 💀 badge) and downstream data (session statuses and the resulting crash-free rate for a given release).
- We should be emitting values which better reflect reality (by itself not a hard change, but one which needs the above two issues to get figured out first).

marandaneto:
Thanks for the detailed message @lobsterkatie 
@smeubank and Daniel Khan will discuss this in the next planning.

text/0010-exception-types.md
```diff
@@ -0,0 +1,120 @@
...
+
+A user of Sentry should be able to 
+
+- filter events on the issues page or discover for the newly introduces exception types (3 categories).
+- highlight (similar to the unhandled label) events of the type unhandled and process termination.
+- get alerted for events of the type unhandled and process termination separately.
+
+Currently, there's an unhandled label on the issue's page but it's only highlighted for process termination errors (if `handled: false`).
+
+In order to propagate those exception types, the exception mechanism needs to be adapted:
```
antonpirker, line 39: I like this option more, because it is backwards compatible. 
But I would not add another boolean, the new "process_terminated" variable needs to be added because "handled" is a boolean and thus not very future proof. 
So I would add a number/enum kind of type called something more general like `error_state` or such that can then have values like `process_terminated`, `timeout`, `deadlock`, `caught-fire`, ...

antonpirker:
One general question: How can we send data to sentry, when the process terminates? Is this even possible? (can we store the error fast enough to disk to send on next start?)

antonpirker:
And @lobsterkatie can this new thing catch real browser crashes: https://web.dev/reporting-api/ maybe?

text/0010-exception-types.md
marandaneto: > One general question: How can we send data to sentry, when the process terminates? Is this even possible? (can we store the error fast enough to disk to send on next start?)

yes

marandaneto:
> One general question: How can we send data to sentry, when the process terminates? Is this even possible? (can we store the error fast enough to disk to send on next start?)

yes

lobsterkatie:
> And @lobsterkatie can this new thing catch real browser crashes: [web.dev/reporting-api](https://web.dev/reporting-api/) maybe?

Not really. Or, rather, technically yes, but [not in all browsers](https://web.dev/reporting-api/#browser-support), and not with really any useful data. The problem is, it's not a hook like `onerror` is, it's just the ability to set a URL to receive a set data payload. (It's the same way [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri) reports work.) And in the case of a crash, that data payload you get is pretty underwhelming:

image

(You can test it out by using [this demo site](https://reports-endpoint.glitch.me/).)

text/0010-exception-types.md
```diff
@@ -0,0 +1,120 @@
...
+
+- [Flutter](https://github.com/getsentry/sentry-dart/issues/456)
+- Browser SDKs
+- Unity
+- React Native
+- .NET (UnobservedTaskException)
+
+# Unresolved Questions
+
+- Are there any other exception types next to the ones metioned in this RFC?
```
marandaneto, line 120:
```suggestion
- Are there any other exception types next to the ones metioned in this RFC?
- When there's an unhandled exception, we'd automatically finish the transaction bound to the scope, but only if the app crashes (process termination)? See [issue](https://github.com/getsentry/develop/issues/443).
```

smeubank:
transfering questions from Eran Arkin

A few questions:
1. What splitting (crashed/not crashed) will allow the user to achieve? I understand it will impact the health metrics, but will it let them fix issues more easily? To triage issues quicker? Alert the right people?
2. Do we have an estimate on what % of our customer base this impacts split by free/paid? Are there workarounds today that people are using that we can gauge the interest level?
3. I’m not sure what’s the cost of piping this change to the UI (i.e., filters, search, alerts, etc.). So that’s something we will need to budget.

Points 1 & 3 I think developers here can answer from experience working with and on Sentry

point 2, % of customer based would be for all platforms where the boolean doesn't really apply. In theory they would all be impacted

marandaneto:
> transfering questions from Eran Arkin
> 
> A few questions:
> 
> 1. What splitting (crashed/not crashed) will allow the user to achieve? I understand it will impact the health metrics, but will it let them fix issues more easily? To triage issues quicker? Alert the right people?
> 2. Do we have an estimate on what % of our customer base this impacts split by free/paid? Are there workarounds today that people are using that we can gauge the interest level?
> 3. I’m not sure what’s the cost of piping this change to the UI (i.e., filters, search, alerts, etc.). So that’s something we will need to budget.
> 
> Points 1 & 3 I think developers here can answer from experience working with and on Sentry
> 
> point 2, % of customer based would be for all platforms where the boolean doesn't really apply. In theory they would all be impacted

> 1. What splitting (crashed/not crashed) will allow the user to achieve? I understand it will impact the health metrics, but will it let them fix issues more easily? To triage issues quicker? Alert the right people?

This is answered in the RFC already, but I can elaborate more:
You'd be able to prioritize issues better `crashes > unhandled > hadled` and ignore less important issues (cut noise).
Because of the importance of issues, you can triage quicker and alert only for the important ones, saving you time to be alerted for something not so important or not being alerted at all.

> 3. I’m not sure what’s the cost of piping this change to the UI (i.e., filters, search, alerts, etc.). So that’s something we will need to budget.

The changes on the SDKs should be very quick, but I can't estimate filters, search, alerts etc, it's not part of my scope.

ueman:
> transfering questions from Eran Arkin
> 
> A few questions:
> 
> 1. What splitting (crashed/not crashed) will allow the user to achieve? I understand it will impact the health metrics, but will it let them fix issues more easily? To triage issues quicker? Alert the right people?

Yep, currently the problem is that our app basically never crashes (as in the app quits, that's how the platform works), so there's no easy way to differentiate between exception which we caught ourselves vs those that were unhandled but didn't crash (as in quitting the app) the application. And those are the one we're interested in fixing.

> 2. Do we have an estimate on what % of our customer base this impacts split by free/paid? Are there workarounds today that people are using that we can gauge the interest level?

No clue, since I'm not working for Sentry but for a company which is a paying enterprise customer. Workarounds are filtering based on the mechanism with which the exception was reported, but that requires that the SDK does add a mechanism in those cases and it also requires a probably unrealistic good understanding of the SDK and UI. I just happen to have contributed a lot of code to Sentry SDKs, so I'm not a good example.

lobsterkatie:
> 1. What splitting (crashed/not crashed) will allow the user to achieve? I understand it will impact the health metrics, but will it let them fix issues more easily? To triage issues quicker? Alert the right people?

I'd echo everything @marandaneto  and @ueman's said, and add that for JS, what the customer gains is data which isn't totally bogus, which feels like a good thing in and of itself. I/we would have fixed the data long ago if there weren't UI/product consequences, but even then it would have been only semi-accurate, because the reality is that there really is a distinction between crashes and unhandled errors. If we were to fix the booleans to be the correct booleans tomorrow, people's crash rates on their releases would skyrocket, because all sorts of things we're incorrectly marking as handled would get marked as unhandled and therefore count as crashing the app, even though that's not reflective of reality.  (There's the added complication I detailed above of what exactly _does_ count as a crash in JS, especially on the browser side, but at least if there were three options, there'd be the _possibility_ of reporting accurate data, and then it'd just be on us SDK devs to figure out what to send when.)

> 2. Do we have an estimate on what % of our customer base this impacts split by free/paid? Are there workarounds today that people are using that we can gauge the interest level?
> point 2, % of customer based would be for all platforms where the boolean doesn't really apply. In theory they would all be impacted

There's no real workaround here. As for impact, I don't know the total numbers, but the mere fact that it affects JS SDK users already means we're talking about a plurality of our customers, both paid and unpaid. I wouldn't be surprised if adding in all of the other affected platforms pushed us past 50%.

szokeasaurusrex:
I agree in general with your idea of having three separate categories to split errors recorded manually by users, errors recorded by Sentry auto-instrumentation, and errors which crash the program. 

However, I think we should be taking a step back here to reconsider our terminology and how we indicate error severity to users. The current handled/unhandled terminology is somewhat misleading in my opinion, since at least in the Python SDK that I am most familiar with, it does not actually indicate whether the exception gets caught, but rather whether the exception was reported manually by the user or by the SDK itself via an instrumentation. Furthermore, the UI visually indicates "unhandled" exceptions as more severe than "handled" ones, and I question whether this is necessarily the case. 

For this reason, I propose changing the handled/unhandled terminology to something more meaningful, and I also propose creating a separate measure of error severity. 

### Capturing mechanism

For the capturing mechanism, we should have two categories to separate manually and automatically captured errors. These would replace the current labels of "handled" and "unhandled," respectively.

- **Manually captured**: Errors which users record via `Sentry.capture*` methods.
- **Automatically captured**: Errors which the Sentry SDK recorded automatically. 

Unlike the current “handled” and “unhandled” categories, automatically captured errors **would not necessarily be marked as higher severity** than manually captured ones. 

### Severity levels

To indicate error severity, we should have a separate concept of error severity. I propose three error severity levels as follows, but am open to other suggestions:

- **Low**: Errors which are less likely to be caused by a bug, and more likely to be caused by user error, for example a 404 error returned by a web server. This category would be the default for manually captured errors, but users would be able to set a high severity on a per-event basis, instead. 
- **High**: Errors which likely indicate a bug in the code or some other more serious issue that developers need to fix, for example a 500 error returned by a web server. This category would likely be the default for most types of automatically captured errors, but users would be able to set a low severity on a per-event or per-error-type basis, instead. 
- **Critical**: Errors which crashed the process. This category would be reserved for crashes; i.e. users would be unable to manually mark an error as critical, and errors which crash the process would always be marked as critical regardless of configurations. 

This separate severity level mechanism would be more flexible, since users would have the option to set a different severity level for certain errors or configure a different default level for certain classes of errors, instead of tying the severity level to how the error was captured.

Lms24:
I generally like the idea of moving away from handled/unhandled. Two things:

> For the capturing mechanism, we should have two categories to separate manually and automatically captured errors. These would replace the current labels of "handled" and "unhandled," respectively.

For the mechanism, we already have the `type` field. Maybe the way to go is to standardize what values we set for the respective cases and leave `handled` unset by default.

> we should have a separate concept of error severity

I guess severity levels are hard to define well across different platforms. For example, in Browser JS we can’t detect a full crash and it basically never happens. Also besides Http error responses (which not all of our SDKs capture), it’s probably hard to distinguish between high/low reliably. 

Whatever we do here, if we want to solve this for the entire product, we need to coordinate this change with how we set the session status and, based on that, how we calculate release health. Maybe the key here is adapting the calculation to a certain platform, so that it makes (somewhat) sense for web, mobile and backend/server projects respectively.

InterstellarStella:
https://sentry.zendesk.com/agent/tickets/146342

PR #11, opened 2022-09-20
0011: gocd succeeds Freight as our CD solution

joshuarli:
To meet the growing demands of continuous deployment (CD) at Sentry, this RFC proposes gocd as Freight's successor and the foundation upon which scalable pipelines for CD are built internally.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/0042/text/0042-gocd-succeeds-freight-as-our-cd-solution.md)

### DIFF
--- a/text/0042-gocd-succeeds-freight-as-our-cd-solution.md
+++ b/text/0042-gocd-succeeds-freight-as-our-cd-solution.md
@@ -1,0 +1,62 @@
+* Start Date: 2022-09-20
+* RFC Type: feature
+* RFC PR: https://github.com/getsentry/rfcs/pull/11
+* RFC Status: draft
+
+# Summary
+
+To meet the growing demands of continuous deployment (CD) at Sentry, this RFC proposes [gocd](https://www.gocd.org/) as [Freight’s](https://www.notion.so/Freight-and-the-Deployment-Pipeline-7a7541decb8c4c1a8fb3741f49d3018b) successor and the foundation upon which scalable pipelines for CD are built internally. Freight can only run one sequential pipeline per deployment, and has over many years accumulated enough complexity hindering feature development, maintenance, and scaling. While the ability to customize the UI is greatly limited with an out-of-the-box solution, we (the Dev Infra team) believe gocd’s UI will meet the majority of needs with minimal workarounds.
+
+# Motivation
+
+In its current state Freight represents very primitive pipelining technology. Deployment of a service (e.g. sentry, snuba, relay) is limited to a single, sequential pipeline. Pipelines, and tasks within pipelines can’t execute asynchronously and with dependency relationships. There isn’t even functionality to pass values throughout tasks in a pipeline. Further, as Freight is not a task scheduler and instead is designed to run on a single host, CD cannot be horizontally scaled.
+
+Freight has been around for many years, and as primary contributors are no longer here it has been in maintenance mode for a long time. In short, it’s a collection of docker containers running outdated software tied to a lot of messy disk state on an outdated host machine. The bespoke software and processes that support it, from `dockerctl` to `freight-cli` to `getsentry-deploy` to `fab`, represent a lot of brittle and error-prone tech debt.
+
+A more specific example among many: Freight uses postgres extensively and alembic as its SQL driver at the app-level - no one knows what alembic is and this actually resulted in some downtime as a recent Freight migration had to be figured out adhoc. As you can imagine, working on Freight locally (so as to not incur downtime) is difficult as well.
+
+Here’s a non-exhausative list of some recent feature requests as the engineering organization grows:
+
+- canary deployments
+- gradual/slower/segmented rolling deploys
+- in-flight and rapid rollbacks
+- self-service customization of sensitive deploy configurations
+- adaptation to single-tenant and future hybrid cloud deployment CD workflows (deploying to GKE clusters in separate google projects)
+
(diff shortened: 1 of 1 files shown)


mitsuhiko:
I changed the number of the RFC to the pull request number as intended. Should be changed on merge.

text/0042-gocd-succeeds-freight-as-our-cd-solution.md
```diff
@@ -0,0 +1,62 @@
...
+3. **[Tekton](https://tekton.dev/)**. Like gocd, Tekton is general-purpose pipelining tech. It’s a young project (2018) donated by Google to cd.foundation. Tekton is deployed in a k8s cluster, and as such Tekton executors have k8s abilities builtin, but are generic otherwise; it is not strictly a k8s controller like ArgoCD. Compared to gocd, it’s a complex system with moving parts in k8s. The UI, Tekton Dashboard, isn’t particularly purposeful or focused and feels like a thin CRUD layer on top of Tekton’s k8s CRDs (don’t think normal users would want to edit something like ClusterTriggerBindings).
+4. Argo Workflows. The same company that made ArgoCD, also made this. It’s pretty much analogous to Tekton except the UI is more confusing and hard to navigate.
```
tonyo, line 39: For the sake of additional context:
We've been using Argo Workflows in the Quality Engineering team for almost a year now (there's a deployed version [here](https://run.testa.getsentry.net/), with a bunch of workflows definitions [here](https://github.com/getsentry/test-factory/tree/main/k8s/services/workflows)). It's a general-purpose workflow engine (closer to Jenkins in its universality), and can be combined with all other Argo products (Rollouts, Events, CD, etc.), and that's probably its main curse: you can implement pretty much any kind of pipeline/workflow there, but if you want to scale those (as in, have more than a couple pipelines), you better invest in some helpers, wrappers, and custom scripts to tailor that 

Other impressions:
* UI is not super fancy, but after a few uses you don't really notice it anymore. There are a few UI bugs/annoyances here and there, but generally those that are reported get fixed in next releases. K8S complexity can be abstracted out e.g. by configuring workflows with a limited number of inputs, basically the same thing that we have in Freight:

image

* If there are too many stages and dependencies in the workflow, the execution graph can get pretty messy, without any straightforward way to get it rendered nicely.

* The tool is K8S native, for better or worse. Since we also use `sentry-kube` for the QE env at the moment, we basically pre-render all resources before applying, which is slightly annoying. At the same time, ordinary users are not directly exposed to all K8S horrors: there's a button for viewing logs, and that's usually enough for people who "just use" the system.

* Build artifacts management is pretty useful: test results, service configurations, HTML reports can be registered as artifacts and then used as inputs for other pipeline stages, or downloaded/viewed. Probably not that important for the vanilla CD use case, but generally things like an HTML report (e.g. with links to logs, or some analytics) could be a nice addition to every deploy we do, so wonder if GoCD supports anything like that.

* Poor support for Google Identity-Aware Proxy for auth (or more like, lack of it). This one is a bit of a bummer, given that tools like Grafana solve it pretty nicely (most importantly, transparently). Does GoCD have good auth options?

tonyo:
Is there maybe a deployed version of GoCD somewhere, to get a feel of it?

text/0042-gocd-succeeds-freight-as-our-cd-solution.md
joshuarli: Oh cool, I wasn't aware that we were already using Argo Workflows somewhere. I'm definitely not a fan of the UI either but also rejected it because I wasn't smart enough to write some basic working YAML for it. Now that I know we have some prior art I'm more open to trying again, but for now we've already invested into some more serious exploration into gocd. Good news is a lot of the work being done to port Freight and getsentry-deploy functionality into standalone scripts is generic and reusable if gocd falls through and we pivot to Argo Workflows (which IMO now, is the next best candidate).

joshuarli: Thanks for chiming in with that amount of detail Anton, I appreciate it. To answer some questions:

- gocd has first class support for HTML reports which we plan on using to do things like provide links to the corresponding sentry release: https://docs.gocd.org/current/configuration/managing_artifacts_and_reports.html
- there's [a gocd plugin](https://github.com/gocd-contrib/google-oauth-authorization-plugin) which I believe @mattgauntseo-sentry tested and it Just Works? But it's only for auth. I'm unsure if it associates an identity with the gmail login that can be used for https://docs.gocd.org/current/configuration/managing_users.html, I don't think we've looked into that yet. But that plugin would certainly be easier than putting Google IAP in front of it.

mattgauntseo-sentry:
There isn't a deployed version that I'm aware, but running it locally has been fairly painless for us, this should be all you need https://www.gocd.org/test-drive-gocd.html

text/0042-gocd-succeeds-freight-as-our-cd-solution.md
mattgauntseo-sentry: @joshuarli the plugin will do the oauth flow for access and create users in the gocd instance which you can use to assign roles. It'll also use the email to show who is performing actions in gocd (i.e. showing who kicked off a deployment or paused a deployment).

joshuarli:
> There isn't a deployed version that I'm aware, but running it locally has been fairly painless for us, this should be all you need https://www.gocd.org/test-drive-gocd.html

We're closeish to an official deployment, I would just wait for a while.

mitsuhiko:
Since we agreed on GoCD we should be merging this RFC.

mitsuhiko (approved):

PR #13, opened 2022-09-22
Propose new Portable PDB related protocol fields

Swatinem:
This RFC proposes to add new fields to the event protocol to support symbolication of .NET stack frames that use Portable PDB debug files.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/portable-pdb/text/0013-portable-pdb.md)

### DIFF
--- a/text/0013-portable-pdb.md
+++ b/text/0013-portable-pdb.md
@@ -40,33 +40,38 @@ for each frame in the stack trace:
 
 ## Stack Trace Interface
 
-We propose to add the following to the [Stack Trace Interface Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes):
+We propose to use/add the following fields to the
+[Stack Trace Interface Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes)
+that should be used to facilitate symbolication using Portable PDB files:
 
-**`addr_mode: "rel:$idx"`**, _existing_
+**`addr_mode: "rel:$idx"`**, _existing_, _required_
 
 This re-uses the existing `addr_mode` field using relative indexing.
 The index points into `debug_meta.images` protocol field and references a debug file
 that contains the debug data necessary to symbolicate this stack frame.
 
-**`instruction_addr: HexValue`**, _existing_
+**`instruction_addr: HexValue`**, _existing_, _required_
 
 This re-uses the existing `instruction_addr` field and represents the
 function-relative IL Offset and corresponds to the .NET [`StackFrame.GetILOffset`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stackframe.getiloffset) method.
 
-**`function_id: HexValue`**, _new_
+**`function_id: HexValue`**, _new_, _required_
 
 A new field that represents a unique identifier for this function _inside the referenced image_.
 This is derived from the .NET [`MemberInfo.MetadataToken`](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.metadatatoken).
 This `MetadataToken` is encodes the metadata table in its most significant byte and the index into that table as the lower 3 bytes (24 bits).
+The SDK should filter specifically for `MethodDef` tokens and only transmit the lower 3 bytes when it in fact is a `MethodDef` token.
 
 ### Open Questions:
 
 - ~~How does WASM symbolication work? How is the concept of a _function index_ being used in WASM?~~
   - WASM has the context of `function_index`, though it is not useful for symbolication as the module relative instruction offset is enough to do DWARF lookups.
   - The WASM `function_index` would be useful to look up the function metadata in WASM format, which we do not use.
+- ~~Should the `function_id` be a `HexValue`, or completely free-form?~~
+  - We will restrict this to a `HexValue ` initially. We are free to relax that to `String` in the future if needed.
+- ~~Should we decode the `MetadataToken` on the client? (filter for only `MethodDef`, and mask away the table id?)~~
+  - Yes please. Since this is the functions ID, it should validate/decode that in the SDK.
 - Should we introduce another `addr_mode` to call out "function-relative" addressing explicitly?
-- Should the `function_id` be a `HexValue`, or completely free-form?
-- Should we decode the `MetadataToken` on the client? (filter for only `MethodDef`, and mask away the table id?)
 
 ### Existing implementation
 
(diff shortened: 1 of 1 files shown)


text/0013-portable-pdb.md
```diff
@@ -0,0 +1,86 @@
...
+# Options Considered
+
+## Stack Trace Interface Extensions
+
+We propose to add the following to the [Stack Trace Interface Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes):
+
+- `instruction_offset: Number`, as it is currently used by the `sentry-dotnet` SDK.
+- `method_index: Number`.
+
+The `instruction_offset` is defined as a JSON `Number`.
```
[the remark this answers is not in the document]
Swatinem: https://github.com/dotnet/runtime/blob/main/docs/design/specs/PortablePdb-Metadata.md#sequence-points-blob
This defines that:

> IL Offset is within range [0, 0x20000000)

I believe the ECMA 335 spec also has tight limits on what the IL offset can be.
This should be well in the range what JS/ON `Number` can safely hold.

If we want to change this, this would also need a breaking change in the `sentry-dotnet` SDK. CC @mattjohnsonpint

mitsuhiko:
We have discussed this previously but the entries in `debug_meta` today are a bit of a weird hybrid. An entry usually refers to a single module and then describes both the object file and the debug companion. However the `type` is almost exclusively referring to the file type of the object file. So we talk about `pe`, `macho`, `elf` and so forth. The exception to this rule today I believe is only the `proguard` entry which describes the debug companion only.

With the portable PDB situation it's odd now because the `pe` (which already is portable in the name) is more or less the same for .NET and native. While technically true that there is a difference as the .NET PE really is just a dummy, it does carry a lot of useful PE information still.

The proposal here to introduce a `portable-pe` (portable, portable executable :P) is understandable and probably not a terrible idea, but I wonder if that is actually how we should think about these debug files. An alternative world would be to allow a `debug_meta` entry to just refer to a debug file instead. That way you could also potentially associate more than one debug companion with a module.

Swatinem:
I would also prefer to describe the debug companion (`portable-pdb` in this case), that would make a lot more sense to me.

text/0013-portable-pdb.md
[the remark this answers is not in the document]
mattjohnsonpint: We currently send the `instruction_addr` in hex format and the `instruction_offset` as a long integer.  We could easily change both to be sent in hex format without requiring a "breaking" change - assuming the back end will accept either format.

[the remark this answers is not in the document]
mitsuhiko: I think even if it's unlikely that this values goes out of the currently defined 32bit range, better safe than sorry and implement as hex field. SDKs can keep sending integers if they are capable of fitting it into their JSON int.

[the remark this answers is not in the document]
mattjohnsonpint: Looks like `instruction_offset` isn't retained anymore, after https://github.com/getsentry/sentry/pull/4892?

Swatinem:
Fun fact: PE files can indeed have more than one CodeView record, for example here is a hex view of `System.Reflection.Metadata.dll`:

grafik

That file does not seem to have a PDB checksum however.

It seems like a lot of the `System.XXX` files have two codeview records.

text/0013-portable-pdb.md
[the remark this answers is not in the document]
Swatinem: Oh, that is interesting, looks like `instruction_offset` meant something entirely different previously: The offset from the start of the symbol? (just going by the description of the linked PR)

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+- `SymbolChecksum` Header: https://github.com/dotnet/symstore/blob/8a7c47ac74302510f839cc2361ca591b6f4df542/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs#L112
+
+# Options Considered
+
+## Stack Trace Interface Extensions
+
+We propose to add the following to the [Stack Trace Interface Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes):
+
+- `instruction_offset: Number`, as it is currently used by the `sentry-dotnet` SDK.
+- `method_index: Number`.
```
jan-auer, line 57: The protocol does not refer to methods but functions. I would ask to use consistent language and call this either `function_index` or potentially even `function_id` if we want to make this more generic.

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Debug Meta Interface Extensions
+
+We propose to add the following to the [Debug Images Interface](https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images):
+
+- A new debug image type similar to the existing native image types.
+- It should have a `type: "portable-pe"` attribute.
+- A `debug_id: String` attribute with a `DebugId` that is parsed from the [CodeView Debug Directory Entry](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#codeview-debug-directory-entry-type-2).
+- A `debug_file: String` attribute with the corresponding PDB file path parsed from the same CodeView Entry.
+- A `debug_checksum: String` attribute that is parsed from the [PDB Checksum Debug Directory Entry](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#pdb-checksum-debug-directory-entry-type-19). The format for this attribute should be `${algorithm}:${hex-bytes}`, for example `SHA256:xxxx`.
```
jan-auer, line 79: Please also add `code_file` and `code_id` like for non-portable pdbs.

jan-auer, line 79: Do these fields need to be validated or normalized during ingestion? What are valid algorithms (e.g. `\w+`)?

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+- Key Format: https://github.com/dotnet/symstore/blob/aa44862e5028cb7595bbd474da1d63e8c24bf718/src/Microsoft.SymbolStore/KeyGenerators/KeyGenerator.cs#L166-L177
+- `SymbolChecksum` Header: https://github.com/dotnet/symstore/blob/8a7c47ac74302510f839cc2361ca591b6f4df542/src/Microsoft.SymbolStore/SymbolStores/HttpSymbolStore.cs#L112
+
+# Options Considered
+
+## Stack Trace Interface Extensions
+
+We propose to add the following to the [Stack Trace Interface Frame Attributes](https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes):
+
+- `instruction_offset: Number`, as it is currently used by the `sentry-dotnet` SDK.
```
jan-auer, line 56: WASM uses the `instruction_addr` along with an `addr_mode` to achieve a similar alternative addressing schema. We are applying a different approach for a similar problem here, and I would ask to align.

There are two options:
1. Use the same protocol as WASM: `instruction_addr` and a new `addr_mode`. Note that in WASM "relative" indicates that the address is in the address space of the entire image/module, so we would need a new mode, for example `"func_rel"`. That address mode declares that the address is relative to a function (defined below).
2. Introduce a new `instruction_offset` field, but then also get rid of `addr_mode` for WASM and have a dedicated field there.

jan-auer, line 56: `instruction_offset` is a good generic name for such a field in principle. Given that the old fields for this have been long removed and were inactive, there is no risk of collisions.

If we want to be more specific, we could also name this something like `function_offset`, which would further differentiate the semantics of this field to WASM's addressing mode.

text/0013-portable-pdb.md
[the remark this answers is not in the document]
jan-auer: The protocol generally accepts both `0x`-prefixed hex and numeric representations for address-like fields. As such, you do not have to consider it a breaking change if you switch the SDK from one to another representation. 

I agree with @mitsuhiko's ask to make this an `Addr` type to allow for large values. The SDK may still send numbers as long as it can guarantee that they do not exceed 53 bit significance.

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+- `method_index: Number`.
+
```
jan-auer, line 58: Please call out _required_ fields that already exist in the protocol as those are relevant for lookup in the next section. It looks like this is just the `addr_mode`.

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+- The current `type: "pe"` image is referencing a PDB file via its `debug_file` and `debug_id` attributes.
+- An alternative spelling could be `"portable-pdb"`, `"portablepdb"` or `"ppdb"`.
+- The `debug_checksum` attribute name has a similar reason. It is the checksum of the referenced debug file.
+
+# Drawbacks
+
+This increases the complexity and surface area of the Event Payload Interface. However all the proposed fields are
+eventually necessary to offer symbolication for .NET stack traces based on debug data included in Portable PDB files.
+
+# Implementation guidelines
```
jan-auer, line 93: Please describe a high-level algorithm for using these protocol fields to perform lookups and symbolications. This helps to illustrate use cases for these protocol fields.

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+
+WASM stack frames can also refer to the concept of a function index. An alternative would be to use a property named `function_index` for both.
+
+## Debug Meta Interface Extensions
+
+We propose to add the following to the [Debug Images Interface](https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images):
+
+- A new debug image type similar to the existing native image types.
+- It should have a `type: "portable-pe"` attribute.
+- A `debug_id: String` attribute with a `DebugId` that is parsed from the [CodeView Debug Directory Entry](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#codeview-debug-directory-entry-type-2).
```
jan-auer, line 77: The DebugID is stricter than string fields. Unfortunately, there is no RFC, so you can either link to [develop docs](https://develop.sentry.dev/sdk/event-payloads/debugmeta/#pe-images-pdbs) or [the crate](https://docs.rs/debugid/latest/debugid/struct.DebugId.html).

text/0013-portable-pdb.md
```diff
@@ -0,0 +1,105 @@
...
+### Alternatives
+
+WASM stack frames can also refer to the concept of a function index. An alternative would be to use a property named `function_index` for both.
+
+## Debug Meta Interface Extensions
+
+We propose to add the following to the [Debug Images Interface](https://develop.sentry.dev/sdk/event-payloads/debugmeta#debug-images):
+
+- A new debug image type similar to the existing native image types.
+- It should have a `type: "portable-pe"` attribute.
```
jan-auer, line 76: PE already stands for "Portable Executable". Such files are in fact regular PEs, though they contain special sections. I would ask to add reasoning on why a dedicated image type is required as opposed to just calling this `pe`. 

As a suggestion for naming, you could call this `pe_net` or `pe_assembly`, though I would defer to others for better suggestions. 

Finally, please consider snake_case rather than kebab-case. Outside of CSP, our protocol almost exclusively uses snake_case.

text/0013-portable-pdb.md
Swatinem: [the spec](https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#pdb-checksum-debug-directory-entry-type-19) defines `SHA256`, `SHA384` and `SHA512` right now, but I can imagine this can change in the future.

text/0013-portable-pdb.md
loewenheim: How about this:
 - there is a field `instruction_addr`.
 - there is a field `addr_mode` which may be `abs` or `rel:n`.
 - there is an optional field `function_index` (name subject to change).

 `instruction_addr` is then interpreted as follows:

 - `addr_mode = abs`: `instruction_addr` is an absolute address
 - `addr_mode = rel:n` and `function_index` is missing: `instruction_addr` is relative to module `n` (the WASM case)
 - `addr_mode = rel:n` and `function_index = k`: `instruction_addr` is relative to function `k` in module `n` (in other words, an IL offset)

Swatinem:
The `.dll` has a text section and unwind info which we can dump via `dump_cfi`, but it does not seem to have a symbol table.
For the `.ni.pdb`, our `object_debug` reports a symbol table and debug info, but creating a symcache for it via `symcache_debug` leaves us empty.

text/0013-portable-pdb.md
mitsuhiko: I don't hate this but I wonder if we then don't want to have a `funcrel:n:m` where `n` is the module index and `m` is the function index?

Swatinem: We were also talking about that last week, and did not really like it because it was too "stringly typed". And parsing things from a specially formatted string is strictly worse than just having proper JSON properties for it.

However I do recognize that the whole problem boils down to having a properly discriminated union:
```rust
enum Addr {
  Absolute(u64),
  ModuleRelative { module_idx: u32, addr: u64 },
  FunctionRelative { module_idx: u32, function_idx: u32, addr: u64 },
}
```

Swatinem:
## Summary

* I settled for a new `"pe_dotnet"` type now. There is enough evidence that we do need to *somehow* flag these files differently from normal pe files so a new type it is.
* I went with `instruction_addr` + `addr_mode: "rel:X"` + `function_id: HexValue`. We discussed putting the function index into the `addr_mode`, but I believe having a separate field is better than putting more information into an opaque string that needs to be parsed.
* I verified that the chain from SDK through relay to sentry and symbolicator works fine and correctly symbolicates events, using the WIP PRs linked from the RFC.

## Next Steps

I believe we can merge the RFC and the PR to relay right away. The PR to sentry can also be merged, but will only take effect once the `symbolic` dependency is also updated to correctly recognize the new Portable PDB file type.
Support has already landed in symbolic master. We need a new release and can then merge the PRs to sentry-cli and symbolicator, and update the dependency in sentry as well.

That leaves the dotnet SDK as the last piece of the puzzle. My PR works, but does need some work. Also the current SDK architecture is not well suited to write out stack frames and debug images "atomically" at the same time.

jan-auer (approved):

PR #14, opened 2022-09-26
Propose a new way to continue a trace

### LINKED ISSUE #690: [DEV-DOCS] start_transaction(true)
### `start_transaction` doesn’t continue baggage by default 
- Nested transactions are problematic for baggage since we don’t automatically propagate them for custom transaction calls
- out of the box, we carry forward baggage and sentry-trace headers (propagate trace) for nested transactions
  - If there is a problem we can re-evaluate is needed
- Integrations can decide if they want to

### - Next steps: dev docs spec and try to move forward,

proposed solution: `start_transaction(create_new_trace=True)`

Places to document:
- [Static API Changes](https://develop.sentry.dev/sdk/performance/#static-api-changes)
- new argument

Open points:
- at the very least agreement on what the default behavior should be per platform
  - or more high level client side vs server side

```php
<?php

$sentryTraceHeader = $request->getHeaderLine('sentry-trace');
$baggageHeader = $request->getHeaderLine('baggage');

$transactionContext = \Sentry\Tracing\TransactionContext::continueFromHeaders($sentryTraceHeader, $baggageHeader);

// Start the transaction
$transaction = \Sentry\startTransaction($transactionContext);

...

$transactionContext = new \Sentry\Tracing\TransactionContext();

// Start a nested transaction
// In this case, we would not continue the trace and instead start a new one.
$transaction = \Sentry\startTransaction();

https://sentry.io/organizations/sentry-sdks/performance/sentry-php:c81f843eb6474d53aca7902229cbe32a

---

$transactionContext = \Sentry\Tracing\TransactionContext::fromParent($transaction);
// Start a nested transaction
$transaction = \Sentry\startTransaction($transactionContext);

public static function fromParent(Transaction $transaction)
{
    $context = new self();
    $context->traceId = $transaction->getTraceId();
    $context->parentSpanId = $transaction->getParentSpanId();
    $context->sampled = $transaction->getSampled();
    $context->getMetadata()->setBaggage($transaction->getBaggage());

    return $context;
}

https://sentry.io/organizations/sentry-sdks/performance/sentry-php:a2e3d29c847e49ebac13ce9448254af6

// - How would someone fetch the current transaction? SentrySdk::getCurrentHub()->getTransaction()?
// - Should we auto populate this inside TransactionContext::__construct()?
```

cleptric:
This RFC proposes a new way to continue a trace when creating nested transactions.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/continue-trace/text/0014-continue-traces.md)

Refs https://github.com/getsentry/develop/issues/690

### DIFF
--- a/text/0014-continue-traces.md
+++ b/text/0014-continue-traces.md
@@ -1,0 +1,63 @@
+* Start Date: 2022-09-26
+* RFC Type: feature
+* RFC PR: https://github.com/getsentry/rfcs/pull/14
+
+# Summary
+
+This RFC proposes a new way to continue a trace when creating nested transactions.
+
+# Motivation
+
+The current way we propagate `sentry-trace` and `baggage`, is to pass a correctly populated `TransactionContext` as the first argument to `startTransaction()`.
+
+```php
+use Sentry\Tracing\TransactionContext;
+use function Sentry\startTransaction;
+
+$transactionContext = TransactionContext::continueFromHeaders($sentryTraceHeader, $baggageHeader);
+$transaction = startTransaction($transactionContext);
+
+```
+
+In case someone starts another nested transaction without passing in any context, a new trace will be started and the Dynamic Sampling Context is lost as well.
+
+# Options Considered
+
+## Add TransactionContext::fromParent()
+
+```php
+use Sentry\Tracing\TransactionContext;
+use function Sentry\startTransaction;
+
+$transactionContext = TransactionContext::fromParent($transaction);
+$transaction = startTransaction($transactionContext);
+
+public static function fromParent(Transaction $transaction)
+{
+    $context = new self();
+    $context->traceId = $transaction->getTraceId();
+    $context->parentSpanId = $transaction->getParentSpanId();
+    $context->sampled = $transaction->getSampled();
+    $context->getMetadata()->setBaggage($transaction->getBaggage());
+
+    return $context;
+}
+```
+
+## Add a third argument to `startTransaction()`
+
+```php
+use Sentry\Tracing\TransactionContext;
+use function Sentry\startTransaction;
+
+$transactionContext = new TransactionContext();
+$transaction = startTransaction($transactionContext, [], bool $continueTrace = true);
+```
+
+This would require making the SDKs aware of the current request.
+In PHP, we could rely on `$_SERVER['HTTP_SENTRY_TRACE]` and `$_SERVER['HTTP_BAGGAGE]`, but this is not possible in all languages.
+
+# Unresolved questions
+
+* Can we rely on `SentrySdk::getCurrentHub()->getTransaction()` to fetch the current transaction to be passed into `TransactionContext::fromParent()` ?
+* How would we make `TransactionContext::__construct()` aware of the current request?
--- a/text/0014-continue-traces.md
+++ b/text/0014-continue-traces.md
@@ -56,6 +56,7 @@ $transaction = startTransaction($transactionContext, [], bool $continueTrace = t
 
 Inside `TransactionContext::__contruct`, we could check for an ongoing transaction on the Hub and continue the trace automatically.
 
-# Unresolved questions
+# Drawbacks/Impact
 
-* Can we rely on `SentrySdk::getCurrentHub()->getTransaction()` to fetch the current transaction to be passed into `TransactionContext::fromParent()` ?
+- This will increase the public API surface of our SDKs
+- Depending on the option, it's either more complex or more magical.
--- a/text/0014-continue-traces.md
+++ b/text/0014-continue-traces.md
@@ -54,10 +54,8 @@ $transactionContext = new TransactionContext();
 $transaction = startTransaction($transactionContext, [], bool $continueTrace = true);
 ```
 
-This would require making the SDKs aware of the current request.
-In PHP, we could rely on `$_SERVER['HTTP_SENTRY_TRACE]` and `$_SERVER['HTTP_BAGGAGE]`, but this is not possible in all languages.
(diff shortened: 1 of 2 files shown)


text/0014-continue-traces.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Motivation
+
+The current way we propagate `sentry-trace` and `baggage`, is to pass a correctly populated `TransactionContext` as the first argument to `startTransaction()`.
+
+```php
+use Sentry\Tracing\TransactionContext;
+use function Sentry\startTransaction;
+
+$transactionContext = TransactionContext::continueFromHeaders($sentryTraceHeader, $baggageHeader);
```
marandaneto, line 17: On Java, it's a bit different but similar https://docs.sentry.io/platforms/android/performance/instrumentation/custom-instrumentation/#distributed-tracing
We've not implemented the `continueFromHeaders` method, [reason is here](https://github.com/getsentry/sentry-java/issues/1346).

text/0014-continue-traces.md
```diff
@@ -0,0 +1,62 @@
...
+$transactionContext = new TransactionContext();
+$transaction = startTransaction($transactionContext, [], bool $continueTrace = true);
+```
+
+Inside `TransactionContext::__contruct`, we could check for an ongoing transaction on the Hub and continue the trace automatically.
+
+# Drawbacks/Impact
+
+- This will increase the public API surface of our SDKs
+- Depending on the option, it's either more complex or more magical.
```
lobsterkatie, line 62: I'd opt for exposing the option but giving it a reasonable default so that most people don't have to deal with it.

That said, I'd echo Manoel's question: In what circumstances is starting a new transaction in the same process better than just nesting spans?

text/0014-continue-traces.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Options Considered
+
+## Add TransactionContext::fromParent()
+
+```php
+use Sentry\Tracing\TransactionContext;
+use function Sentry\startTransaction;
+
+$transactionContext = TransactionContext::fromParent($transaction);
```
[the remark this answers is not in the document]
cleptric: The reason for this RFC was a use-case in [getsentry/sentry](https://github.com/getsentry/sentry) itself. There are cases where we start transactions inside transactions.

Link to our [internal Slack conversation].(https://sentry.slack.com/archives/C043MQLSLTS/p1663598491050169)

text/0014-continue-traces.md
cleptric: See https://github.com/getsentry/rfcs/pull/14#discussion_r984345016

Swatinem:
👍🏻 for the general idea. I implemented [`TransactionContext::continue_from_span`](https://docs.rs/sentry/latest/sentry/struct.TransactionContext.html#method.continue_from_span) in Rust since the very beginning since it felt like the right thing to do.

👎🏻 on a bool option that looks up the current `Hub`. Unless something changed, we modeled Transactions to be mostly independent from Hubs. There could be multiple concurrent transactions, and the decision which one is bound to the hub is up to the API user.

sl0thentr0py:
I would ask @beezz, @untitaker etc to also maybe clarify the use case and if this is a workaround for some other problem because it is not entirely clear to many of us why starting a new transaction is necessary instead of just adding spans.
 https://github.com/getsentry/sentry/blob/a10312cf1f10af0a79a07b9895565963429dce24/src/sentry/ingest/ingest_consumer.py#L156-L168

sl0thentr0py:
Regarding preference, I agree with @Swatinem that a new method with explicit semantics is better than another magic bool variable that breaks hub assumptions on the current one. 
The tradeoff here is that it won't magically make dynamic sampling work on those who use the current explicit `startTransaction`, they'll have to migrate their usages first.

text/0014-continue-traces.md
lobsterkatie: Okay, yes, thanks for that. But I'm still not sure if the issue is what this fixes or the fact that maybe we really shouldn't be nesting transactions...

marandaneto:
A similar issue that would be solved by a solution to continue the trace thru the same process.
https://github.com/getsentry/sentry-dart/issues/721
It works with the [current workaround](https://github.com/getsentry/sentry-dart/issues/721#issuecomment-1023125360) but this is semantically wrong because when you think about `header`, you think about external services (HTTP and so on).

cleptric:
Thanks again for all your input! 🙂 
I'm closing this for the moment and we can revisit it if the need for this will arise again in the future.

mitsuhiko:
Can we revisit this? We ran into this again with monitors where we need to continue a trace but not yet create a transaction which I believe part of this RFC covers.

Refs https://github.com/getsentry/sentry/issues/43647

cleptric:
@mitsuhiko After talking with @evanpurkhiser  and @davidenwang, it seems like this would be a different ask than the initial goal of this RFC.
What we had in mind here was to auto-magically continue a trace from `startTransaction`, the other use-case now being to propagate the trace without being in the context of a transaction.
Does this make sense?

mitsuhiko:
@cleptric we can probably separate these things down but there are two paths we might want to look at:

1. we continue a trace independent of the transaction (eg: is that what happens if we continue from a cronjob?)
2. we do always continue a transaction, but create a clearer user experience around making a transaction when there is already one ongoing. (is that what we want?)

We can open a new RFC with a clearer set goal but unless we have an initial hunch where we are leaning here we might want to understand this first.

```python
# Does this create a transaction or a trace. What are the consequences of either?
sentry_sdk.continue_trace_from_environment()
```

PR #15, opened 2022-09-26
Propose a new way to expose the handled property

cleptric:
This RFC proposes a new way to expose the handled property to a user.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/expose-handled/text/0015-expose-handeled-property.md)

### DIFF
--- a/text/0015-expose-handeled-property.md
+++ b/text/0015-expose-handeled-property.md
@@ -39,7 +39,7 @@ Allow a user to set the `handled` property through the `EventHint`.
 Instead of calling `captureException()`, people can access the `handeled` property by manually creating an event and passing in the desired `mechansim` to the `EventHint`, which then gets passed to `captureEvent()`.
 
 
-```
+```php
 $hint = EventHint::fromArray([
     'exception' => $exception,
     'mechanism' => new ExceptionMechanism(ExceptionMechanism::TYPE_GENERIC, false),


marandaneto:
Before we go ahead with this change, in case we wanna align that for the unified API, lets also discuss https://github.com/getsentry/rfcs/pull/10

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+): void {
+    captureException($exception);
+
+    parent::logException($exception, $request, $includeTrace);
+}
+``` 
+
+This creates the problem, that all exceptions/errors are now considered `handled: true`.
+
+# Options Considered
```
marandaneto, line 31: An option to consider is actually exposing the whole [Sentry protocol](https://develop.sentry.dev/sdk/event-payloads/) and actually making it possible that you can create a SentryEvent, with a mechanism. either handled or not.

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Add `captureUnhandledException()`
+
+Add a new global method to allow a user to capture unhandled exceptions explicitly.
+
+## Expose the handled property through the  `EventHint`
+
+Allow a user to set the `handled` property through the `EventHint`.
```
marandaneto, line 39: This is going to be a somehow hidden feature, people would need to learn about it anyway, so I'd rather not do it.

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Add `bool $handled = true` as an optional parameter to `captureException()`
+
+Add a fourth argument to `captureException()` to allow a user to set the `handled` property.
```
marandaneto, line 43: Not all languages have either optional params or overloads, in case that matters for the unified API.

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+}
+``` 
+
+This creates the problem, that all exceptions/errors are now considered `handled: true`.
+
+# Options Considered
+
+## Add `captureUnhandledException()`
+
+Add a new global method to allow a user to capture unhandled exceptions explicitly.
```
marandaneto, line 35: That sounds nice, I am wondering tho if users will actually call that manually, our global error handler should be able to create an event with a `handled=false` mechanism, if not, I'd rather fix this first.

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+use Psr\Http\Message\ServerRequestInterface;
+use Throwable;
+use function Sentry\captureException;
+
+public function logException(
+    Throwable $exception,
+    ?ServerRequestInterface $request = null,
+    bool $includeTrace = false
+): void {
+    captureException($exception);
```
marandaneto, line 23: What most SDKs do here is to create a new SentryEvent, with a mechanism and handled=false and instead of calling `captureException`, they call `captureEvent`, is that even possible?

text/0015-expose-handeled-property.md
vladanpaunovic: Semantically, this sounds like the most reasonable way to do this. PHP is already derailing a bit from other SDKs here (not that this is a good thing).

What I don't like here is that it will be a bit too much params for a simple function, yet from the user perspective it does sound like the least of all evils.

My vote goes to proceed with this approach.

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
...
+
+    parent::logException($exception, $request, $includeTrace);
+}
+``` 
+
+This creates the problem, that all exceptions/errors are now considered `handled: true`.
+
+# Options Considered
+
+## Add `captureUnhandledException()`
```
vladanpaunovic, line 33: I am not a fan of this because in that case we should rename `captureExceltion()` to `captureHandledException()` too in order to keep this change meaningful.

vladanpaunovic (commented):
After considering proposed options, my vote goes to
### Add `bool $handled = true` as an optional parameter to `captureException()`

text/0015-expose-handeled-property.md
antonpirker: In our framework integrations in the Python SDK we create a custom `_capture_exception()` methods that sets all the data that we want in the event and then call `capture_event`: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/flask.py#L204-L219

text/0015-expose-handeled-property.md
antonpirker: Example how Python does this: https://github.com/getsentry/sentry-python/blob/master/sentry_sdk/integrations/flask.py#L204-L219

text/0015-expose-handeled-property.md
```diff
@@ -0,0 +1,43 @@
+* Start Date: 2022-09-26
+* RFC Type: feature
+* RFC PR: https://github.com/getsentry/rfcs/pull/15
+
+# Summary
+
+Allow a user to decide if an exception/error is considered `handled: true` or `handled: false`.
```
antonpirker, line 7: I do not get the "why" for this request. If an exception is caught by the global exception handler of the SDK it is considered not handled. If the user catches the exception himself/herself and then sends it to Sentry with `captureException()` it is considered as handled, because it was handled by the user.

I think this is how it should work.

text/0015-expose-handeled-property.md
cleptric: I agree.

text/0015-expose-handeled-property.md
cleptric: A PHP application can only register one `error_handler` at a time.
The `error_handler` we ship with the Sentry SDK is therefore often overwritten by the framework `error_handler`, hence we have to hook into it. And then we can't access the `ExceptionMechanism` anymore.

text/0015-expose-handeled-property.md
cleptric: `Client::captureException()` calls `Client::captureEvent(Event::createEvent(), $hint, $scope);` in PHP, but attaches the `Exception` to the `hint`. Still, I can't access `handled` here.
Maye this whole RFC is just trying to fix an issue in the PHP SDK, idk.

marandaneto: Sounds like that, the `mechanism` is part of the event protocol and should be accessible, if that's possible to fix, I'd go this way instead of exposing new public APIs.

AbhiPrasad: I'm also +1 on exposing the mechanism to users - but then we should make a decision on what `handled` does in https://github.com/getsentry/rfcs/pull/10 first.

marandaneto: Makes sense, but I'd not block this RFC because of #10.
My recommendation would be: simply do `handled: false` if it's an unhandled issue and it should be shown as such in the UI + being able to filter and alert on them.

cleptric:
> What most SDKs do here is to create a new SentryEvent, with a mechanism and handled=false and instead of calling `captureException`, they call `captureEvent`, is that even possible?

So after picking around in the PHP SDK, I came to the conclusion that this would be the best option to move forward.

https://github.com/getsentry/sentry-php/pull/1416

Due to some BC issues, I "hacked" it in, but it works fine for now.
Shall we use this RFC as a conclusion that SDKs should take this path or shall we close it?

AbhiPrasad:
>  Shall we use this RFC as a conclusion that SDKs should take this path or shall we close it?

I'm in favour of this - exposing the mechanism in the hint is a pretty clean pattern + works with event processors.

cleptric:
This is ready to be merged. Pls add your ✅

vladanpaunovic (approved):
LGTM

AbhiPrasad (approved):

PR #16, opened 2022-09-26
decision: Automated code mappings

armenzg:
We are looking into creating code mappings automatically for organizations that have code source management installed.

We will initially target orgs with Github installations with Python and JS projects (when source maps are available) that do *not* have code mappings already in place.
Other source code management integrations and projects will be considered in the future.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/armenzg/code-mappings/text/0016-auto-code-mappings.md)


text/0016-auto-code-mappings.md
```diff
@@ -0,0 +1,86 @@
...
+  * Inaccurate at times (as per @asottile-sentry)
+
+### Considerations
+
+* Not all projects have a platform set (e.g. other), thus, we will need to look at file extensions
+* There are some orgs that have more than one Github org associated to their Sentry org
+  * I don't think this is a problem
+* There are some projects that have code from more than one repository
+* Some projects have more than one code mapping per repository for a given project
+  * For instance, Sentry has `sentry` and `sentry_plugins` code mappings
```
leeandher, line 61: These make it extremely difficult IMO for this feature to be used reliably. We can spend the R&D effort trying to guess the shape of their codebase, but it ultimately isn't teaching the user how/why Sentry needs the mappings in the first place. Plus this will be a difficult feature to scale to different mediums or compiled languages, which can introduce inconsistencies.

No one knows the codebase better than the developers, and it may be worth guiding the user along how to set up these codemappings for these unique scenarios with in-app tutorials/messaging rather than differing them to docs.

text/0016-auto-code-mappings.md
```diff
@@ -0,0 +1,86 @@
...
+* Code map derivation has not yet been tested against JS stack traces (**UNKNOWN**)
+  * Scheduled to be tackled in current sprint
+
+## Drawbacks
+
+None I can think of.
+
+## Unresolved questions
+
+* We are leaning toward not notifying users that code mappings were added to a project that did not have code mappings since they didn't even know there were missing out in any features. In the future, in cases where code mappings are already in place for a project, we will need to re-evaluate and discuss it since we may cause workflow regressions for ownership rules.
```
leeandher, line 86: I think the nitty gritty isn't that important to convey but not communicating that this change has been made might introduce some user confusion. My opinion is that changes to user-editable details (even minor) should probably be surfaced in some way.

Plus, if we're making those updates messaged, it'd also be a venue to add it this feature to the notification platform. If you restructure your app (e.g. `source/` to `src/`), features in Sentry will break without telling you they need attention. Adding these notifications for managers/owners might grant some visibility onto the features, and we can just leverage that for automatic updates like this doc specs out.

leeandher (commented):
I don't see any technical issues with initiative, but I do think that it may just be worth calling out a few smaller easy wins that will accomplish the goal here in a similar way:
- Surfacing an alert to add a codemapping elsewhere in the app (e.g. issue details - kinda like what we do if your sourcemaps aren't set for a JS project)
- Even better would be surfacing the interface directly, to change these codemappings, like in a modal or something directly whereever they should be using a feature, like stacktrace linking, or codeowners.


Also worth calling out:
- We should also focus on integrating code-mappings into the onboarding experience for GitHub, or else the problem might keep growing. What I can picture is you just get redirected from Github after adding your repos, and we list them all with (!) icons showing they require mappings for 'feature X, feature Y, feature Z' and how an interface to start setting that up.

text/0016-auto-code-mappings.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Details
+
+A code mapping allows going from a stackframe module path to a source file in the preferred source code management tool:
+
+<img src="0016-code-mapping-logic.png" width="500px" />
+
+The process is quite simple. Give a stacktrace frame, we look for the file name in all of the repos we have access for a customer's Github org. For instance, given the stacktrace path `sentry/integrations/gitlab/client.py` look for `integrations/gitlab/client.py` and if there is a unique match we have determined all the values for the code mapping. For the curious, a rudimentary POC (only for Python atm) can be viewed and tested [here](https://gist.github.com/armenzg/40ba48fff217815842c4fe16047d0835).
+
+Our initial thoughts are to run a **scheduled task** that will look for projects without code mappings. For each of those projects look at analyzing various stack traces and add code mappings for the modules that there are exact matches.
```
scefali, line 42: @armenzg I am curious how we find these exact matches. How many stack traces do we go through? How do we determine the best rule?

text/0016-auto-code-mappings.md
```diff
@@ -0,0 +1,86 @@
...
+    * Potential for a bug since it may be pointing to an old vendored version
+    * Issues come with info about packages. I wonder if in some cases we may be able to detect this.
+* Not all orgs will have granted us access to All their Github repositories upon installation (or modified it later)
+  * This will reduce the number of exact matches
+* As part of the automation we will be creating Repository objects
+  * If we end up adding a lot of repos for the org we may be indirectly causing UI issues when there’s no proper pagination including dropdowns
+* Available APIs
+  * Search API (proven to work in POC)
+    * The [Search API](https://docs.github.com/en/rest/search#rate-limit) has a rate limit of 30 requests per minute, thus, at best we could create 30 code mappings per minute
+      * **UNKNOWN** We may be able to create multiple tokens to increase the capacity
```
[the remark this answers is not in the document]
armenzg: We're definitely not going to use the Search API but needed to document that it was considered. 30 requests/minute does not make the cut. There are also inaccuracy concerns.

text/0016-auto-code-mappings.md
armenzg: This pass will not focus on editing values but adding code mappings where there are none set.

text/0016-auto-code-mappings.md
```diff
@@ -0,0 +1,86 @@
...
+
+About 20% of repos connected to Sentry have code mappings and very few have both the stack trace and source code root values set which can lead to improper stack trace linking and code assignments. For complete data follow [link for staff](https://www.notion.so/sentry/Deriving-Code-Mappings-9086faaf3fed4faca69e8b35f8f70e26#bae13a87214d4f52b936e0e1aa6829ec).
+
+## Details
+
+A code mapping allows going from a stackframe module path to a source file in the preferred source code management tool:
+
+<img src="0016-code-mapping-logic.png" width="500px" />
+
+The process is quite simple. Give a stacktrace frame, we look for the file name in all of the repos we have access for a customer's Github org. For instance, given the stacktrace path `sentry/integrations/gitlab/client.py` look for `integrations/gitlab/client.py` and if there is a unique match we have determined all the values for the code mapping. For the curious, a rudimentary POC (only for Python atm) can be viewed and tested [here](https://gist.github.com/armenzg/40ba48fff217815842c4fe16047d0835).
```
markstory, line 40: Should we be explicit about this *not* working for java as file paths are not always present?

armenzg:
Thank you all for the feedback! I will get back to you on Monday. My apologies on the delay.

armenzg:
> I don't see any technical issues with initiative, but I do think that it may just be worth calling out a few smaller easy wins that will accomplish the goal here in a similar way:
> - Surfacing an alert to add a codemapping elsewhere in the app (e.g. issue details - kinda like what we do if your sourcemaps aren't set for a JS project)

Now that anyone can add code mappings this will be powerful.

> - Even better would be surfacing the interface directly, to change these codemappings, like in a modal or something directly whereever they should be using a feature, like stacktrace linking, or codeowners.

That's right! This will be part of the upcoming UI changes.

> Also worth calling out:
> - We should also focus on integrating code-mappings into the onboarding experience for GitHub, or else the problem might keep growing. What I can picture is you just get redirected from Github after adding your repos, and we list them all with (!) icons showing they require mappings for 'feature X, feature Y, feature Z' and how an interface to start setting that up.

@robinrendle is already thinking about this and doing some design work for it.

text/0016-auto-code-mappings.md
scefali: @armenzg so what happens if you have code from multiple repos?

armenzg: @scefali I have updated the script to handle multiple repos:
https://gist.github.com/armenzg/40ba48fff217815842c4fe16047d0835#file-output-txt-L20-L29
and it produces the correct code mappings:
```
--CODE MAPPINGS--
{'sentry': {'repo': 'getsentry/sentry',
            'stacktrace_root': 'sentry',
            'src_path': 'src/sentry'},
 'sentry_plugins': {'repo': 'getsentry/sentry',
                    'stacktrace_root': 'sentry_plugins',
                    'src_path': 'src/sentry_plugins'},
 'getsentry': {'repo': 'getsentry/getsentry',
               'stacktrace_root': 'getsentry',
               'src_path': 'getsentry'}}
```

scefali (approved):
Seems solid to me

PR #19, opened 2022-09-27
Bring `beforeSend`-like capability to transactions

lobsterkatie:
Proposal to modify SDKs to allow transactions to run through `beforeSend` or equivalent method

_(Originally specifically a proposal for adding `beforeSendTransaction`, but revised after initial discussion)_

[Rendered RFC](https://github.com/getsentry/rfcs/blob/kmclb-beforeSendTransaction/text/0019-beforeSendTransaction.md)

### DIFF
--- a/text/0019-beforeSendTransaction.md
+++ b/text/0019-beforeSendTransaction.md
@@ -1,0 +1,61 @@
+* Start Date: 2022-09-26
+* RFC Type: feature
+* RFC PR: https://github.com/getsentry/rfcs/pull/19
+
+# Summary
+
+Proposal to add `beforeSendTransaction` to all SDKs.
+
+# Motivation
+
+For error events, `beforeSend` allows the user to make whatever changes they want to event data and to filter events however they want, which is useful for solving the cases where our built-in mechanisms for doing so don't correctly handle their particular use case. There is no such hook for transactions. Instead, users have to use `addGlobalEventProcessor` to add an event processor to the scope. This is less than ideal in a few ways:
+
+- Ease of use (or lack thereof)
+  - Nothing about `beforeSend` is obviously errors-only, so users assume it applies everywhere and try to use it for transactions.
+  - Users have to learn two different ways to do effectively the same thing.
+  - It forces users to write code in two separate places, which leads to potential redundancy if they want to act on any properties shared by error and transaction events.
+
+- Lack of ultimate control over transaction events
+  - Event processors happen in an unspecified order, and there's no guarantee a processor added by a user will run last. This means that an integration, for example, might change event data after the user's last chance to intervene.
+
+(That last point is not _strictly_ true. As a last resort, a user could create a custom transport or could proxy events through their own servers, but both of those are pretty ugly, cumbersome workarounds.)
+
+# Background
+
+When transactions were introduced, the decision was made not to run them through `beforeSend` because they follow a slightly different schema, and therefore had the potential to break any existing `beforeSend` which relies on its input being a certain shape. It's unclear whether a transaction-specific `beforeSend`-type hook was discussed at the time.
+
+# Supporting Data
+
+Issues where this has come up:
+- https://github.com/getsentry/sentry-docs/issues/5525
+- https://github.com/getsentry/sentry-javascript/issues/4723
+- https://github.com/getsentry/sentry-javascript/issues/5442
+- https://github.com/getsentry/sentry-python/issues/1226
+
+(and I'm sure many others)
+
+# Options Considered
+
+*Proposal*
+
+Add `beforeSendTransaction` to all SDKs, which (as the name implies) would work exactly the same way `beforeSend` does, but would act upon transactions.
+
+*Alternative*
+
+In the next major of each SDK, start sending both errors and transactions through `beforeSend`.
+
+*Comparison*
+
+The main advantage the proposed `beforeSendTransaction` option has over the everything-goes-through-`beforeSend` option is that it's not a breaking change, and therefore doesn't need to wait for a major release to be introduced. (Non-breaking changes are also always less friction for the user, at least in the short run.)
+
+The advantages the everything-goes-through-`beforeSend` option would have over the proposed `beforeSendTransaction` option are 1) all user code for filtering events and changing their data could live in one spot, and 2) we wouldn't be left with an option which is only for errors but doesn't say it's only for errors (similar to the current situation with `sampleRate` and `tracesSampleRate`). [EN: Reason number 2 actually makes me wish we _could_ go with the everything-goes-through-`beforeSend` option, but I recognize that avoiding user inconvenience has already pretty much won the day for `beforeSendTransaction`.]
+
+# Drawbacks
+
+Compared to doing nothing, the only drawbacks to either of the above options are the time and effort it will take to implement and document the changes.
+
+# Unresolved questions
+
+- Assuming we go with the `beforeSendTransaction` option, what all should go in the hint?
+- Once we do this, should we deprecate `addGlobalEventProcessor` as a public API method?
+- Are we happy with the `beforeSendTransaction` name, or are there better alternatives?
--- a/text/0019-beforeSendTransaction.md
+++ b/text/0019-beforeSendTransaction.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Proposal to add `beforeSendTransaction` to all SDKs.
+Proposal to make `beforeSend` (or a similar method) work for transactions (and possibly other payload types in the future) in all SDKs.
 
 # Motivation
 
@@ -36,26 +36,67 @@ Issues where this has come up:
 
 # Options Considered
 
-*Proposal*
+_(Revised after preliminary discussion in https://github.com/getsentry/rfcs/pull/19)_
+
+### **Option 1**
+
+_(Original proposal, demoted to one option of several after discussion)_
 
 Add `beforeSendTransaction` to all SDKs, which (as the name implies) would work exactly the same way `beforeSend` does, but would act upon transactions.
 
-*Alternative*
+Pros:
+- Not a breaking change, so less pain for customers in the short run.
+- Not a breaking change, so no need to wait for a major release to implement it.
+
+Cons:
+- Increased API surface.
+- Code that acts on shared event properties (tags, contexts, etc) would need to be duplicated between `beforeSend` and `beforeSendTransaction`.
+- Unclear/non-parallel naming when considered alongside `beforeSend`. (`beforeSendTransaction` is a clear name, but `beforeSend` is not - it really ought to be `beforeSendError`; similar problem to current confusion with `sampleRate` and `tracesSampleRate`.)
+- Future payload types (replays, profiles, etc) would need their own methods. Where does it end?
+
+### **Option 2**
 
 In the next major of each SDK, start sending both errors and transactions through `beforeSend`.
 
-*Comparison*
+Pros:
+- `beforeSend` has a seemingly-universally-applicable name, so the most intuitive thing for users is to have it fulfill its implied role and apply universally.
+- Future payload types could  be added without introducing new init options/versions of the method.
 
-The main advantage the proposed `beforeSendTransaction` option has over the everything-goes-through-`beforeSend` option is that it's not a breaking change, and therefore doesn't need to wait for a major release to be introduced. (Non-breaking changes are also always less friction for the user, at least in the short run.)
+Cons:
+- Potentially breaking change for everyone, even people not using tracing, because old assumptions about `beforeSend` input schema would no longer hold. (We could maybe work around this by just surrounding our `beforeSend` call with a `try-catch` and returning the event untouched if things go sideways.)
 
-The advantages the everything-goes-through-`beforeSend` option would have over the proposed `beforeSendTransaction` option are 1) all user code for filtering events and changing their data could live in one spot, and 2) we wouldn't be left with an option which is only for errors but doesn't say it's only for errors (similar to the current situation with `sampleRate` and `tracesSampleRate`). [EN: Reason number 2 actually makes me wish we _could_ go with the everything-goes-through-`beforeSend` option, but I recognize that avoiding user inconvenience has already pretty much won the day for `beforeSendTransaction`.]
+### **Option 3**
+
+Like option 2, except `beforeSend` doesn't change and instead we create a new universal method to take its place.
+
+Pros:
+- Not a breaking change, so less pain for customers in the short run.
+- Not a breaking change, so no need to wait for a major release to implement it.
+
+Cons:
+- What would we call it? The most logical name given other option names (`beforeBreadcrumb`, `beforeNavigate`, etc) is already taken.
+- What would happen to `beforeSend`? Either we'd eventually either: 1) deprecate and remove it, which just kicks the breaking-change can down the road, or 2) keep it around in perpetutity, which isn't great either, because it would probably continue to confuse people (both by providing a less-functional way to do what the new method does and by _still_ not being named `beforeSendError`).
+
+### **Option 4**
+
+_(Listed for the sake of completeness, and not a bad idea in and of itself, but by general consensus probably not the right option here)_
+
+Implement lifecycle hooks for transactions (and possibly replays, profiles, etc in the future)
+
+Pros:
+- Granular control
+- Depending on how we do it, might line us up with OTel.
+
+Cons:
+- Significantly bigger footprint than other options, not only in terms of work to build it but also documentation, support, and bundle size.
+- Would act primarily on `Transaction` objects whereas we're trying to match `beforeSend`'s ability to act on finalized events.
 
 # Drawbacks
 
-Compared to doing nothing, the only drawbacks to either of the above options are the time and effort it will take to implement and document the changes.
+Compared to doing nothing, the only drawbacks to any of the above options are the time and effort it will take to implement and document the changes. (In other words, the idea of giving people `beforeSend`-like powers over transactions doesn't have any notable drawbacks on its own.)
 
 # Unresolved questions
 
-- Assuming we go with the `beforeSendTransaction` option, what all should go in the hint?
-- Once we do this, should we deprecate `addGlobalEventProcessor` as a public API method?
-- Are we happy with the `beforeSendTransaction` name, or are there better alternatives?
+- If we go with the `beforeSendTransaction` option, what all should go in the hint? Are we happy with the `beforeSendTransaction` name, or are there better alternatives?
+- If we go with a new universal method, what should we call it? What should happen to `beforeSend` afterwards?
+- Once we do this, should we deprecate `addGlobalEventProcessor` as a public API method? (Not inherently a part of this change, but something that was spoken of in the past but then backed away from because event processors were the workaround to the problem this RFC is trying to solve.)


text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,61 @@
...
+The advantages the everything-goes-through-`beforeSend` option would have over the proposed `beforeSendTransaction` option are 1) all user code for filtering events and changing their data could live in one spot, and 2) we wouldn't be left with an option which is only for errors but doesn't say it's only for errors (similar to the current situation with `sampleRate` and `tracesSampleRate`). [EN: Reason number 2 actually makes me wish we _could_ go with the everything-goes-through-`beforeSend` option, but I recognize that avoiding user inconvenience has already pretty much won the day for `beforeSendTransaction`.]
+
+# Drawbacks
+
+Compared to doing nothing, the only drawbacks to either of the above options are the time and effort it will take to implement and document the changes.
+
+# Unresolved questions
+
+- Assuming we go with the `beforeSendTransaction` option, what all should go in the hint?
+- Once we do this, should we deprecate `addGlobalEventProcessor` as a public API method?
```
marandaneto, line 60: People still use event processors for different reasons on Mobile SDKs, not sure if removing them is the right thing to do, I find it useful in some use cases otherwise your `beforeSend` just gets bloated with so many things.

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,61 @@
...
+- Are we happy with the `beforeSendTransaction` name, or are there better alternatives?
```
marandaneto, line 61: Fine for me, but I still have the feeling that this leads to the same problem as `sampleRate` for errors and `tracesSampleRate` for transactions.
`beforeSend` for errors, `beforeSendTransaction` for transactions.
If that's not a concern, all good.

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,61 @@
...
+
+Add `beforeSendTransaction` to all SDKs, which (as the name implies) would work exactly the same way `beforeSend` does, but would act upon transactions.
+
+*Alternative*
+
+In the next major of each SDK, start sending both errors and transactions through `beforeSend`.
+
+*Comparison*
+
+The main advantage the proposed `beforeSendTransaction` option has over the everything-goes-through-`beforeSend` option is that it's not a breaking change, and therefore doesn't need to wait for a major release to be introduced. (Non-breaking changes are also always less friction for the user, at least in the short run.)
```
marandaneto, line 49: IIRC, there are more SDKs that run beforeSend for transactions.
Cocoa is one of them https://github.com/getsentry/sentry-cocoa/issues/1730

AbhiPrasad:
For this, I have a slightly alternative proposal, where in I would like us to adopt more granular span processing hooks, specifically `onSpanStart`, and `onSpanFinish`. This is more complicated in the short-term, but in the long term will help because we probably have to have similar APIs around Replays and Profiles.

To accomplish this, we can adopt a new top level option users can configure, a `Tracer`.  This also means users operate on transactions/spans directly instead of an event (which means they will have to figure out schema differences, like `transaction.name === event.transaction` etc.).

```ts
interface SentryOptions: {
  // ...

  tracer: Tracer

  // ... 
}

interface Tracer {
  onSpanStart?: (span: Span) => Span | null;

  onSpanFinish?: (span: Span) => Span | null;

  // did not include on transaction start since that's handled by sampling options

  onTransactionFinish?: (transaction: Transaction) => Transaction | null;
}
```

I'm not totally attached to the tracer, but I think it would be a nice way to encapsulate the logic. This is also similar to the open telemetry API. In languages where possible the tracer can just be a plain dictionary, otherwise it's a class you must construct.

marandaneto:
> ```ts
> onSpanFinish
> ```

We have something similar for a few integrations, similar to `onSpanFinish`.
Usually called `BeforeSpanCallback`.
The signature varies, but usually takes:
`fun execute(span: ISpan, request: Request, response: Response?): ISpan?`
So you can either drop or mutate the span data based on the request or response.

AbhiPrasad:
> I like the fine-grained lifecycle hooks as well, but they don't solve the problem of 'run some user code at the very end of all other sentry processing'.

Great point. Perhaps then we chose to somehow introduce a new `beforeX` method (idk what it's called), that is called for all events being sent.

```ts
// name pending - hardest problem in computer science...
beforeWhatever: (type: EventType, event: Event, hint?: Hint) => Event | null;
```

Then we have something that'll affect every event type, `replay`, `profile`, `transaction`, etc - making it fully forwards compatible.

sl0thentr0py:
Yeah, I also prefer a unified method, and that seems to be a fairly general sentiment. Just that the ideal name is already taken so we'll have to be creative. :)

text/0019-beforeSendTransaction.md
k-fish: I also somewhat have a feeling this will add confusion, having more functions on the sdk is just more surface area for people to miss.

k-fish:
> onSpanFinish
Will I agree we should move in this direction, until we have an exact use / need for span hooks I think that can be a separate discussion / rfc. 

> Then we have something that'll affect every event type, replay, profile, transaction, etc - making it fully forwards compatible.

Are you proposing just a single function which we can split based on a passed prop? I think that works better 🤔

lforst:
> > I like the fine-grained lifecycle hooks as well, but they don't solve the problem of 'run some user code at the very end of all other sentry processing'.
> 
> Great point. Perhaps then we chose to somehow introduce a new `beforeX` method (idk what it's called), that is called for all events being sent.
> 
> ```ts
> type EventType = 'error' | 'transaction' | 'profile' | 'replay' | 'whatever-comes-next';
> 
> // name pending - hardest problem in computer science...
> beforeWhatever: (type: EventType, event: Event, hint?: Hint) => Event | null;
> ```
> 
> Then we have something that'll affect every event type, `replay`, `profile`, `transaction`, etc - making it fully forwards compatible.

Just for TypeScript type inference this particular API is problematic because the `type` argument is technically completely independent from the `event` argument - which makes it impossible for users to narrow down the type of the event based on the `type` argument.

Ideally, we have a `type` field directly in the `Event` type. But this can be an implementation detail which each SDK decides on by itself.

AbhiPrasad:
> Ideally, we have a type field directly in the Event type. But this can be an implementation detail that each SDK decides on by itself.

We do have, https://develop.sentry.dev/sdk/event-payloads/types/#typedef-EventType, but this doesn't cover profiling and replay, which just use envelope item headers to determine the type of the payload (doesn't follow the event format).

profile - https://github.com/getsentry/relay/blob/4894c1d3bab6111585c29c9ea3b55e1511faca60/relay-profiling/src/lib.rs

replay - https://github.com/getsentry/relay/blob/a705412815a96f92eeb88df5ba553644c8f70ade/relay-replays/src/lib.rs#L14

Perhaps we need to standardize this better on ingestion?

AbhiPrasad:
Actually after writing out the above, maybe there is no world where we can standardize this, and we should just have the explicit methods. Overloaded methods is always a pain, better to be explicit!

stefanosiano:
Just throwing an idea:
make the Sentry events implement a common interface that allows to "register" a callback to.
In other words, have Errors and Transactions (and later maybe even Profiles and Replays) provide callbacks when they are created, making sampling decision, before sent, before processors, etc.
This way we may have methods on sdk init where the user implemets what they want, like `registerErrorCallback` and `registerTransactionCallback`, passing to each of them an object with different methods that can be implemented or ignored, like `onEventCreated`, `onEventSampling`, `onEventBeforeSent`, etc.
So we provide only 1 method to the user to use per event, instead of having to dig across several methods, but also enough granularity to do whatever they want.
I come from Android, so it looks fine for me (and that's what the Android sdk already does for its classes), but don't know how much sense it makes on dynamic languages

text/0019-beforeSendTransaction.md
lobsterkatie: I also think event processors can be useful. I included this question because long ago there was talk of deprecating the method (which is why it's undocumented). We backed off when it became necessary to use it for transaction processing, but our solution here would solve that. @mitsuhiko, I think you had thoughts on this at the time?

lobsterkatie:
> methods on sdk init where the user implemets what they want, like `registerErrorCallback` and `registerTransactionCallback`, passing to each of them an object with different methods that can be implemented or ignored, like `onEventCreated`, `onEventSampling`, `onEventBeforeSent`, etc.

This sounds like a different implementation of the hooks Abhi was talking about. Here I agree with Neel and Kevan - not a bad idea at all, but probably a separate conversation from this one, which is specifically "how do we give users the same swiss army knife for transactions which they have for errors?"

lobsterkatie:
So far it seems like people are favoring the single-method option. I actually prefer that option as well, for many of the reasons outlined, so maybe the question is actually a slightly different one: Assuming we go with a single method, should we have it be `beforeSend` and do it as a breaking change in our respective next major versions, or should we introduce a new method with a new name and eventually deprecate `beforeSend`?

I will update the RFC a bit to reflect the conversation so far.

vladanpaunovic:
> So far it seems like people are favoring the single-method option. I actually prefer that option as well, for many of the reasons outlined, so maybe the question is actually a slightly different one: Assuming we go with a single method, should we have it be `beforeSend` and do it as a breaking change in our respective next major versions, or should we introduce a new method with a new name and eventually deprecate `beforeSend`?
> 
> I will update the RFC a bit to reflect the conversation so far.

I would continue using `beforeSend()` method. Adding more data flow trough this method would not be considered as a breaking change. 
The `beforeSend()` method is semantically correct, it is widely used already and does not require any user actions.

To further expand my 2c on the topic of the RFC and echo what I already said in a meeting yesterday:
I believe that `beforeSend()` should act as inverted webhook, where instead of receiving various events declared by `event.type` we are sending events and allowing users to distinguish between them using `even.type` (or a similar signature). 

My vote goes to expand the usage of `beforeSend()` by passing transactions and later possibly more event types trough it.

text/0019-beforeSendTransaction.md
marandaneto: yeah, I recall event processors being for internal usage only, and never a public API, but they turned out to be very useful in some cases.

marandaneto:
> 

one problem with this solution is that some languages don't have Inheritance so you can't have easily 2 types going thru the same callback if strongly typed.

antonpirker:
So after reading this are we back on square one? The one `beforeSend` to handle all kind of "events" (errors, transactions, profiles, replay) does not work on strongly typed languages because there is no base class "Event" that all the other classes "Transaction", "Profile", "Replay" inherit from? 

Do we need to get our data structures in order first, before we can deal with all the different kind of events in a unified way?

antonpirker:
I guess also just piping all events/transactions/whatever into `beforeSend` from now on, can break a lot of users code because if they access a property of an error that does not exist in an transaction than we will break their code. 

So I guess we have to create a `beforeFinalized` (`beforeStored`, `beforeSaved`, `beforeAccepted`, `beforeDone`) that has the same signature as `beforeSend` and gets all the errors and additionally all the transactions/profiles/etc. 

After 12 months we remove `beforeSend` and after 12 more months we rename the new `beforeX` to `beforeSend` and have a nice API again.

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,102 @@
...
+
+For error events, `beforeSend` allows the user to make whatever changes they want to event data and to filter events however they want, which is useful for solving the cases where our built-in mechanisms for doing so don't correctly handle their particular use case. There is no such hook for transactions. Instead, users have to use `addGlobalEventProcessor` to add an event processor to the scope. This is less than ideal in a few ways:
+
+- Ease of use (or lack thereof)
+  - Nothing about `beforeSend` is obviously errors-only, so users assume it applies everywhere and try to use it for transactions.
+  - Users have to learn two different ways to do effectively the same thing.
+  - It forces users to write code in two separate places, which leads to potential redundancy if they want to act on any properties shared by error and transaction events.
+
+- Lack of ultimate control over transaction events
+  - Event processors happen in an unspecified order, and there's no guarantee a processor added by a user will run last. This means that an integration, for example, might change event data after the user's last chance to intervene.
```
antonpirker, line 19: This was also one thing we observed in Python. The user wanted to add an event processor to filter transactions based on something in the request. But our internal integration added this request to the event also with an event processor.
The  users event processor ran before the integrations event processor so there was no request in the event (or the hint? cant remember) so it was impossible to filter the transactions in an event processor.

Proposal:

We could add special `user_event_processors` (or whatever the name) and those event processors are run AFTER all the currently available event processors. To make sure all the SDK internal processing has been done and now the user can do stuff with the event. 

(One down side: "all the SDK internal processing" also means the event could have been dropped/sampled already)

stefanosiano:
> So I guess we have to create a beforeFinalized (beforeStored, beforeSaved, beforeAccepted, beforeDone) that has the same signature as beforeSend and gets all the errors and additionally all the transactions/profiles/etc.

> After 12 months we remove beforeSend and after 12 more months we rename the new beforeX to beforeSend and have a nice API again.

We could have a simple `beforeSendEvent`, deprecate the old `beforeSend` and create a common event type, so we don't have to name and rename the method.

The question to me is: do we want to keep this `beforeSend` or similar method in the future? 
If we are going to change how the events can be processed, then we may just go with a `beforeTransactionSend` so we don't break anything, since we would (breaking) change it anyway in the new major version. For example we could give a new method (one for all the events or one for each type of events) to give more granular control, or a new method to replace multiple "old" methods, like a callback that also gives the "status" of the event (created, processed, toBeSampled, etc.)
If we don't want to change how events are going to be processed, then the single method may be better, given it's possible to do in all of the languages.

> @vladanpaunovic one problem with this solution is that some languages don't have Inheritance so you can't have easily 2 types going thru the same callback if strongly typed.

@marandaneto Can you give an example of an sdk in a language without inheritance? We could check other ways, like composition or delegation, but if it's not possible then we'd have to separate the methods.

cleptric:
Using `beforeSend` for transactions also creates a lot of work, as we have to release new majors for all our SDKs and have our users update their code.
Going forward with `beforeSendTransactions` feels like it has way less friction.

lobsterkatie:
> I believe that `beforeSend()` should act as an _inverted webhook_, where instead of receiving various events declared by `event.type` we are sending events and allowing users to distinguish between them using `even.type` (or a similar signature).

I'm not sure what you mean, @vladanpaunovic, because those sound like the same thing to me. `beforeSend` receives events, those events have a `type` property, and the user then distinguishes between them based on that property. Am I misunderstanding?

vladanpaunovic:
> > I believe that `beforeSend()` should act as an _inverted webhook_, where instead of receiving various events declared by `event.type` we are sending events and allowing users to distinguish between them using `even.type` (or a similar signature).
> 
> I'm not sure what you mean, @vladanpaunovic, because those sound like the same thing to me. `beforeSend` receives events, those events have a `type` property, and the user then distinguishes between them based on that property. Am I misunderstanding?

No, you got it right. All good. My explanation wasn't great.

marandaneto:
@lobsterkatie option 2 sounds good, an alternative for statically typed languages:

```
public interface BeforeSendCallback {
    // already exists
    @Nullable
    SentryEvent execute(@NotNull SentryEvent event, @NotNull Hint hint);
    
    // adds new method for transaction.
    // sets method as `default` method for transaction, so theres no need to check the type of the event,
    // the method is called accordingly with the class type.
    // no need to use `SentryBaseEvent` nor `event instanceof SentryEvent`
    // friendlier for statically typed languages
    @Nullable 
    default SentryTransaction execute(@NotNull SentryTransaction transaction, @NotNull Hint hint) {
      return transaction;
    }
  }
```

Would that be ok?

@philipphofmann @brustolin would that work on iOS?

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,102 @@
...
+
+### **Option 2**
+
+In the next major of each SDK, start sending both errors and transactions through `beforeSend`.
+
+Pros:
+- `beforeSend` has a seemingly-universally-applicable name, so the most intuitive thing for users is to have it fulfill its implied role and apply universally.
+- Future payload types could  be added without introducing new init options/versions of the method.
+
+Cons:
```
adinauer, line 65: In Java we have `SentryBaseEvent` which is a super class of `SentryEvent` and `SentryTransaction`.

At the moment the signature of `beforeSend` in Java is:
```
SentryEvent execute(@NotNull SentryEvent event, @NotNull Hint hint);
```

Assuming we change the parameter to type `SentryBaseEvent` it would cause people to have to cast it to `SentryEvent` to access things like `setLevel` or `isCrashed`.

An alternative would be to have two parameters that are `@Nullable`.
I assume this also applies to other SDKs (.NET).

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,102 @@
...
+- Potentially breaking change for everyone, even people not using tracing, because old assumptions about `beforeSend` input schema would no longer hold. (We could maybe work around this by just surrounding our `beforeSend` call with a `try-catch` and returning the event untouched if things go sideways.)
```
philipphofmann, line 66: IMO another con is that users might end up checking the type of the payload and then casting it to different payload types.

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,102 @@
...
+
+Like option 2, except `beforeSend` doesn't change and instead we create a new universal method to take its place.
+
+Pros:
+- Not a breaking change, so less pain for customers in the short run.
+- Not a breaking change, so no need to wait for a major release to implement it.
+
+Cons:
+- What would we call it? The most logical name given other option names (`beforeBreadcrumb`, `beforeNavigate`, etc) is already taken.
+- What would happen to `beforeSend`? Either we'd eventually either: 1) deprecate and remove it, which just kicks the breaking-change can down the road, or 2) keep it around in perpetutity, which isn't great either, because it would probably continue to confuse people (both by providing a less-functional way to do what the new method does and by _still_ not being named `beforeSendError`).
```
philipphofmann, line 78: I think these are not cons; these are problems to solve. I think having one method called before sending anything makes sense. Alongside adding the universal method, we could also have fine-grained methods. Maybe that would be Option 5, then.

**Option 5**
Add a new universal `beforeSend` method, deprecate `beforeSend`, and add fine-grained `beforeSend` for events and transactions.
* `beforeSendAll` (name can be discussed)
* `beforeSendEvent`
* `beforeSendTransaction`

**Option 6**
Make `beforeSend` the new universal `beforeSend` method, deprecate, and add fine-grained `beforeSend` for events and transactions.
* `beforeSend`
* `beforeSendEvent`
* `beforeSendTransaction`

philipphofmann:
> @philipphofmann @brustolin would that work on iOS?

@marandaneto, on Cocoa we don't use an interface, we use a [typedef](https://github.com/getsentry/sentry-cocoa/blob/58f558dc58731af7d6522987fe5e95b0440025a0/Sources/Sentry/Public/SentryDefines.h#L46-L50). So what you [pointed out](https://github.com/getsentry/rfcs/pull/19#issuecomment-1282427510) wouldn't work.
```Objc
/**
 * Block can be used to mutate event before its send.
 * To avoid sending the event altogether, return nil instead.
 */
typedef SentryEvent *_Nullable (^SentryBeforeSendEventCallback)(SentryEvent *_Nonnull event);
```

text/0019-beforeSendTransaction.md
```diff
@@ -0,0 +1,102 @@
...
+  - It forces users to write code in two separate places, which leads to potential redundancy if they want to act on any properties shared by error and transaction events.
+
+- Lack of ultimate control over transaction events
+  - Event processors happen in an unspecified order, and there's no guarantee a processor added by a user will run last. This means that an integration, for example, might change event data after the user's last chance to intervene.
+
+(That last point is not _strictly_ true. As a last resort, a user could create a custom transport or could proxy events through their own servers, but both of those are pretty ugly, cumbersome workarounds.)
+
+# Background
+
+When transactions were introduced, the decision was made not to run them through `beforeSend` because they follow a slightly different schema, and therefore had the potential to break any existing `beforeSend` which relies on its input being a certain shape. It's unclear whether a transaction-specific `beforeSend`-type hook was discussed at the time.
```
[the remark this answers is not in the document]
lobsterkatie: > If I remember correctly, there we good reasons for not calling beforeSend for transactions

Yes - from what I recall, it was exactly the reason I listed there, that the schema mis-match had the potential to break people's `beforeSend` functions. Or do you mean there was a reason for not having a transaction-specific `beforeSend`-like option?

text/0019-beforeSendTransaction.md
lobsterkatie: If they did this, would it break downstream processing in the SDK?

text/0019-beforeSendTransaction.md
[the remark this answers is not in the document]
philipphofmann: > It's unclear whether a transaction-specific `beforeSend`-type hook was discussed at the time.

I think this sentence is inaccurate. We discussed it and most likely the outcome was what you just pointed out: `the schema mis-match had the potential to break people's beforeSend functions`

> Or do you mean there was a reason for not having a transaction-specific beforeSend-like option

Could also be, but I can't recall that.

text/0019-beforeSendTransaction.md
philipphofmann: I don't think it would, but IMO this is an unappealing API.

text/0019-beforeSendTransaction.md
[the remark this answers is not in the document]
lobsterkatie: Sorry, I think we're miscommunicating here, because I think we actually agree on the history. Let me try to rephrase:

I know that back in the day, when tracing was brand new, we discussed sending transactions through `beforeSend` itself, and decided against it because of the schema issue potentially creating a breaking change. What I don't know, and you're saying you can't recall either, is if a _separate_ `beforeSend`-like option (`beforeSendTransaction` or similar) was considered, and if it was, why it wasn't implemented. With regard to that question schemas/breaking changes are irrelevant, because no one would have had any existing code to break.

Does that make sense?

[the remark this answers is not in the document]
philipphofmann: >  if a _separate_ beforeSend-like option (beforeSendTransaction or similar) was considered, and if it was, why it wasn't implemented. With regard to that question schemas/breaking changes are irrelevant, because no one would have had any existing code to break.

Yep, that makes sense 👍

lobsterkatie:
@mydea, I believe what you're talking about is exactly what's covered in the first sentence of the `Background` section, no?

> When transactions were introduced, the decision was made not to run them through `beforeSend` because they follow a slightly different schema, and therefore had the potential to break any existing `beforeSend` which relies on its input being a certain shape.

lobsterkatie:
Update after IRL meeting:

Many of the points touched on above were discussed, but in the end there wasn't broad consensus on whether to go with a `beforeSendTransaction` or pipe everything through `beforeSend`. The idea of lifecycle hooks also came up again, and they are starting to seem more likely to happen, so as an in-between, let's-fix-things-for-now option, we decided that SDKs that want to can implement `beforeSendTransaction`, even if we do deprecate it later (at least in SDKs which switch to either a universal `beforeSend` or lifecycle hooks when a new major comes around).

I'm therefore going to close this for now. Once it's clear whether hooks are going to happen, and/or when this comes up again for other events types (profiles, replays), we can revisit the topic.

PR #20, opened 2022-09-28
Propose a Change to the Browser Trace Model

mitsuhiko:
This RFC proposes a change to the browser (frontend) trace model so that user sessions are reusing the traces or at least continuing the sampling context. This issue relates to #14 and https://github.com/getsentry/sentry/pull/39349 and the confusion it causes in the dynamic sampling configuration experience.

This RFC currently proposes to continuing the session but alternative options are possible. I will defer the decision for this RFC to @ale-cota since the main impact today is dynamic sampling.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/browser-traces/text/XXXX-browser-traces.md)

### DIFF
--- a/text/XXXX-browser-traces.md
+++ b/text/XXXX-browser-traces.md
@@ -1,0 +1,98 @@
+* Start Date: 2022-09-28
+* RFC Type: decision
+* RFC PR: <link>
+
+# Summary
+
+This RFC proposes to require a frontend SDK to retain a trace that it continues
+until the SDK naturally ends the user session.  This change is primarily to the
+browser trace model to better support dynamic sampling and to create a more
+coherent user experience.
+
+# Motivation
+
+Today there are two ways to start a trace: they can be started on the backend and then
+continued on the client, or they can be started on the client.  In either case however
+client SDKs are likely to create more traces that are disconnected to the backend trace
+on page navigation or view changes.  This creates the situation that today the only sensible
+way for dynamic sampling is to have traces started uniformly on the frontend project as
+otherwise the dynamic sampling rules from both projects need to be modified (head trace
+is where dynamic sampling rules apply).
+
+The secondary motivation is that creating new traces on navigation also wipes out the
+causal relationship to what happened before.  For instance it's more than possible that
+before a client-side navigation the state of the application corrupted, but we lose that
+trace relationship and a user has to manually piece it back together by for instance
+listing all transactions created by a specific user ID.
+
+# Background
+
+We became aware of this problem in two ways recently:
+
+1. Users want to sometimes create a transaction within another transaction.  Today there
+   is no way to link these together for the purpose of dynamic sampling.  A separate RFC
+   [0014](https://github.com/getsentry/rfcs/pull/14) is proposed to add an explicit way
+   to carry forward the sampling context for a new transaction started after an already
+   existing one.  It works by explicitly continuing the trace.  This solves part of this
+   issue, but it leaves out the case where the sampling context naturally moves to another
+   Sentry project.
+
+2. We wanted to change our own tracing integration to start tracing on the server
+   [Sentry PR #39349](https://github.com/getsentry/sentry/pull/39349) where this would
+   require mirroring the sampling settings to another Sentry project and would also affect
+   API requests detatched from user sessions.
+
+# Supporting Data
+
+The [honeycomb whitepaper on front-end observability](https://www.honeycomb.io/wp-content/uploads/2022/03/Front-end-Observability-Whitepaper-1.pdf)
+recommends continung traces from the server until the natural end of the user Session:
+
+> To accomplish this task, you will use the first event (page load) as the start of your trace and
+> connect that first event to additional spans to build a full trace of the user session. Each span
+> will represent a single thing that you want to track, such as a server request or a user click.
+
+# Options Considered
+
+There are multiple ways in which this problem can be addressed.
+
+## Encouraged Root Trace Project and Session Long Traces
+
+In the most trivial case the recommendation to customers would be to pick one project that
+starts traces for real user sessions.  This could be *either* frontend of backend but it
+should attempt to be consistent about it.  In either case the client SDK should *continue the
+trace* until the browser tab naturally closes.
+
+The consequences are that ``startTransaction()`` always anchors to the already open trace
+and dynamic sampling context.  There can be an extra flag to force the start of a new trace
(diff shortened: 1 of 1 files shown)


text/XXXX-browser-traces.md
```diff
@@ -0,0 +1,98 @@
...
+# Options Considered
+
+There are multiple ways in which this problem can be addressed.
+
+## Encouraged Root Trace Project and Session Long Traces
+
+In the most trivial case the recommendation to customers would be to pick one project that
+starts traces for real user sessions.  This could be *either* frontend of backend but it
+should attempt to be consistent about it.  In either case the client SDK should *continue the
+trace* until the browser tab naturally closes.
```
AbhiPrasad, line 63: The issue here is with traces that then can be multiple hours long - and so then the value of analyzing a trace is completely lost.

I'm strongly against this because I feel like it'll reduce the value of trace view, and make it harder for us to expand the tracing product.

text/XXXX-browser-traces.md
```diff
@@ -0,0 +1,98 @@
...
+trace is in fact the frontend for continued transactions after a page navigate, however the
+sampling context is reused from the original server side request.
+
+In this case the relationship of root project setting the dynamic sampling context would be
+broken up and instead a transaction can explicitly pick up the sampling context of another
+project but still issue disconnected traces.
+
+## Alternative B: Trace to Trace Relationships
+
+A potential alternative would be to continue the current project but allow a trace to annotate
```
AbhiPrasad, line 82: As an aside, in my eyes there are various correlation ID mechanisms that we chat about, and it would be nice to get an RFC going that establishes them all. Off the top of my head:

- release id
- user id
- session id
- trace id
- profile id
- replay id (basically a session id)

The most important relationships are as follows:
- 1 users x R releases
- 1 user x S sessions/replays
- 1 user x T traces
- 1 session/replay x T traces
- 1 profile x 1 trace (though this is changing to 1 profile x T trace iirc)

If we can establish these concepts consistently it'll be a nice foundation for us to keep building off of

text/XXXX-browser-traces.md
mitsuhiko: That sounds a bit like a UX issue? I don't think that it's inherently problematic if a trace is very long, we today already do not draw much of a value out of the trace view.

To me it feels at least like there is more value to the trace being connected than the trace being split into transactions on every navigation.

text/XXXX-browser-traces.md
```diff
@@ -0,0 +1,98 @@
...
+> will represent a single thing that you want to track, such as a server request or a user click.
+
+# Options Considered
+
+There are multiple ways in which this problem can be addressed.
+
+## Encouraged Root Trace Project and Session Long Traces
+
+In the most trivial case the recommendation to customers would be to pick one project that
+starts traces for real user sessions.  This could be *either* frontend of backend but it
```
dashed, line 61:
```suggestion
starts traces for real user sessions.  This could be *either* the frontend or backend, but it
```

text/XXXX-browser-traces.md
dashed: User inactivity in a browser tab could be another natural way of ending a session, and beginning a new one. 

There could be users that leave tabs open for a very long time, and having gaps between transactions might seem weird.

text/XXXX-browser-traces.md
smeubank: big +1 here, i still feel like we are moving torwards treating a trace more like a session, and it will get more and more confusing with our existing sessions/releases and session replay (which is closes to what is the standard definition of a session)

The problem we are solving for to relate transactions across a user session by exanding the inclusion of more traces in a single trace for sampling, I feel will just create more fuzzy area. Confusion for us and our users.

text/XXXX-browser-traces.md
```diff
@@ -0,0 +1,98 @@
+* Start Date: 2022-09-28
+* RFC Type: decision
+* RFC PR: <link>
+
+# Summary
+
+This RFC proposes to require a frontend SDK to retain a trace that it continues
+until the SDK naturally ends the user session.  This change is primarily to the
+browser trace model to better support dynamic sampling and to create a more
+coherent user experience.
```
marandaneto, line 10: I believe this affects mobile as well since the concept is similar to the browser trace model, right?

AbhiPrasad: Yes it will also affect mobile - by making ui.load and resulting interaction and navigation transactions linked.

smeubank:
hi folks we had some discussions in the office and for transaprency I want to share here.

Goal: Answer 3 questions:

1. Sample on traces? & Sessions?
   - Yes. Basically we do not yet, but there is a future where IF there is a session it could be considered in the sampling decision of traces.
2. Does a trace cross page loads? Should traces across page loads be linked?
   - A trace by definition is not across page loads. But we do agree that there is value in linking them
3. What is the scope of a session? Intended
   - We are really only discussing sessions in the situation of a user interacting with a Frontend website. Basically it is from first interaction till end, but there are of course cases where there isn't a clear end, so there could need to be a finite nature and timeout 
   
A graph for context. Here we see for comparison how a session covers the full scope of time of user's interaction with a website. With each row representing traces starting in a frontend and it's connected chain of services. Replay is there as a the new emerging product which has "sessions" but they are not synonymous with these user sessions. 

So our input is that Sentry does want a concept of a session as shown here, a trace should not be expanded in scope to be conflated with a session. Similar to other applications and products. We should continue to research how sessions are handled in other products. Define our definitions of each term, as stated in @AbhiPrasad 's comment above, and discuss how to facilitate such contextual data, considering but not limited to Dynamic Sampling.
   
image

AbhiPrasad:
For Sentry internal employees - the point above by @smeubank was elaborated on by @benvinegar in https://vanguard.getsentry.net/p/cl9eb7qko33680ls604otdki9

mitsuhiko:
@ale-cota with the changes to how dynamic tracing will work going forward on AM2 plans the the like, the direct control customers have on traces are greatly diminished. As a result I do not believe that the original issue here is particularly impactful.

That said, I believe expressing which project originally started the trace in terms of sampling decision might still be important for when starting a trace again later on. I would propose closing this RFC and instead opening one about the particular issues related to expression sampling consistency for sessions rather than traces. Thoughts?

ale-cota:
@mitsuhiko yes, I agree to closing this one and opening a new one regarding user sessions instead.

PR #21, opened 2022-09-29
RFC proposal to enable profiling into sentry-cli

viglia:

[Rendered RFC](https://github.com/getsentry/rfcs/blob/viglia/enable-profiling-in-sentry-cli/text/0021-enable-profiling-in-sentry-cli.md)

### DIFF
--- a/text/0021-enable-profiling-in-sentry-cli.md
+++ b/text/0021-enable-profiling-in-sentry-cli.md
@@ -1,6 +1,6 @@
 * Start Date: 2022-09-29
 * RFC Type: feature
-* RFC PR: <link>
+* RFC PR: [#21](https://github.com/getsentry/rfcs/pull/21)
 
 # Summary
 
--- a/text/xxxx-enable-profiling-in-sentry-cli.md
+++ b/text/xxxx-enable-profiling-in-sentry-cli.md
@@ -1,0 +1,53 @@
+* Start Date: 2022-09-29
+* RFC Type: feature
+* RFC PR: <link>
+
+# Summary
+
+This RFC details the proposal to add profiling instrumentation to the `sentry CLI`
+
+# Motivation
+
+We've recently added the support for profiling in the Rust SDK and we've been looking for a Sentry tool that we could use to dogfood it.
+
+Apart from the point mentioned above, we can use this as an opportunity to improve the performance of `sentry-cli`, should we find any issue from the profiles data.
+
+# Background
+
+The `sentry-cli` was temporarily instrumented for a short period between April and May 2018 with different opting strategy (first opt-out and then opt-in).
+
+As we're planning to re-introduce instrumentation by enabling profiling, we'll start collecting diagnostic data again, hence the need for this RFC.
+
+# Options Considered
+
+## Option 1: simple integration with no opting mechanism
+
+Since we're not collecting user PII we could simply enable profiling without providing a chance to opt-in or opt-out.
+
+**Pros**: we'd be sure to receive the data as there is no way to opt out of it. 
+
+**Cons**: developers using it might not like the idea of diagnostic data being collected without their own approval.
+
+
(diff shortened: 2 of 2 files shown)


kamilogorek:
From my perspective, I'd release the profiling behavior (and possibly also restore an error reporting? (TBD)), in the next minor release as an opt-in behavior. This will allow us to test it out on various machines, our own services, and as people working with other SDKs (eg. mobile or native) to help us test it out.
Once we confirm it's working as expected, we could flip the switch, and make it a default behavior, which people could opt-out of by changing configuration in global or per-project `.sentryclirc`, just like we had with error reporting few years back:

```ini
[sentry]
error_reporing = 0
profiling = 0
```

(option names are irrelevant for now)

mitsuhiko:
I think right now the issue we are running into with `sentry-cli` is that it's both very important to customers, yet we have no consistent commitment to it. I'm a bit worried that given that it sits in the CI infrastructure of our customers it needs to be stable as we know that breakages often break CI.

It should be pretty trivial to make an opt-in work, but we already know that nobody will opt into this, particularly if it needs flipping manually. We had error reporting in it historically and it was never sending anything ever.

viglia:
> It should be pretty trivial to make an opt-in work, but we already know that nobody will opt into this

@mitsuhiko that's also my fear

text/0021-enable-profiling-in-sentry-cli.md
```diff
@@ -0,0 +1,53 @@
...
+* RFC Type: feature
+* RFC PR: [#21](https://github.com/getsentry/rfcs/pull/21)
+
+# Summary
+
+This RFC details the proposal to add profiling instrumentation to the `sentry CLI`
+
+# Motivation
+
+We've recently added the support for profiling in the Rust SDK and we've been looking for a Sentry tool that we could use to dogfood it.
```
vaind, line 11: Maybe start by enabling this conditionally on SDKs that use sentry-cli in CI, i.e. check that `env['GITHUB_REPOSITORY_OWNER'] == 'getsentry'`?

viglia: @vaind good point! that would also be an option. 

Do we have a rough estimate of `getsentry` projects using `sentry-cli` in CI to figure out if and how much data we could expect to get?

One drawback I can think of is that probably most of the runner used in CI will be Linux machines so we won't get to test it on Intel Mac or M1 Mac.

vaind: `sentry-unity` integration tests run on the three platforms, still no arm mac, of course.

No idea about the other project usage though, IIRC it should be used by the android gradle plugin & fastlane plugin

viglia: @mitsuhiko @kamilogorek what do you think? it seems like this could be a viable option that would limit the risk for customers.

kamilogorek: How would that be any different from making it generally opt-in?

This way instead of `env['GITHUB_REPOSITORY_OWNER'] == 'getsentry'` you could do `env['SENRY_CLI_PROFILING'] == '1'` in chosen `getsentry` repositories.

You'd need to do that per repo, but this way you have also more granular control over which repos would run it, as well as allow for non-sentry repos to opt-in. We'd also not have any magic values in the cli itself, as `== 'getsentry'` would be stuck there forever in the binary.

vaind: > How would that be any different from making it generally opt-in?

Exactly by having to do the opt-in (PR, review) in every repo.

> We'd also not have any magic values in the cli itself, as == 'getsentry' would be stuck there forever in the binary.

My idea was to run this for a short while before doing the original proposal of this PR, i.e. enable it by default at which point the check of the repo owner can be removed.

PR #22, opened 2022-10-03
Response context RFC

marandaneto:
Add `Response` interface that contains information on a HTTP response related to the event.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/response-context/text/0022-response-context.md)


text/0022-response-context.md
```diff
@@ -0,0 +1,92 @@
...
+    "status_code": 500,
+    "status_message": "foo bar", // is that useful?
+    "is_redirect": false,
+    "response_body_size": 1000, // bytes
+  }
+}
+```
+
+The `data` field is semantically different than the `Request#data`. The `Request#data` is the data that was sent to the server, while the `Response#data` is just arbitrary data that is attached to the response. The `Response#data` is not used for anything in the SDKs, but it is useful for the user to attach arbitrary data to the response.
+We could rename to `other` or something else to avoid confusion.
```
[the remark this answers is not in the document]
marandaneto: It's not mandatory, we can remove it if it's confusing, I was just thinking that in case we wanna add something more to the `Response` protocol, it'd require adding to Relay, etc, with an arbitrary data field, I can just add something in there easily with back compatibility.

text/0022-response-context.md
```diff
@@ -0,0 +1,92 @@
...
+    },
+    "cookies": "PHPSESSID=298zf09hf012fh2; csrftoken=u32t4o3tb3gg43; _gat=1;",
+    "headers": {
+      "content-type": "text/html"
+    },
+    "env": {
+      "REMOTE_ADDR": "192.168.0.1"
+    },
+    "status_code": 500,
+    "status_message": "foo bar", // is that useful?
```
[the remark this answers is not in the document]
marandaneto: Not really, just found it handy as I've seen it here https://pub.dev/documentation/dio/latest/dio/Response/statusMessage.html
I agree, we can remove it, `status_code` is enough.

text/0022-response-context.md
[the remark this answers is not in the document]
marandaneto: I'd not add the `body` field because there are problems that come with it, and we agreed that the `body` would not make it into this feature yet.
Ideally, we'd add the `request` and `response` bodies via `attachments` to not exceed the max size of events. (1MB currently).
Also, we'd need to scrub (PII) very carefully.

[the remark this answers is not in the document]
jjbayer: > I was just thinking that in case we wanna add something more to the Response protocol, it'd require adding to Relay
I suspect that we can also make Relay accept and pass on any unknown fields in the response context (top level).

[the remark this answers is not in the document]
marandaneto: That would be awesome if possible.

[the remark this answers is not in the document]
marandaneto: TLDR: Either we have a data bag that accepts arbitrary key/values or the response protocol should retain unknown (high level) fields.

text/0022-response-context.md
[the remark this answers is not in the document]
marandaneto: TLDR: Let's remove `status_message`, `status_code` is enough for the user.

text/0022-response-context.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Drawbacks
+
+The `Response` interface is PII sensitive, so we need to be careful about how we scrub it.
+
+The [Dart Dio HTTP Client](https://docs.sentry.io/platforms/dart/configuration/integrations/dio/) integration already adds the `Response` field in the `Contexts` interface, so in case we go with Option 1, we'd need to migrate the data to the new `Response` interface.
+
+# Unresolved questions
+
+* Is `status_message` useful?
```
marandaneto, line 95: NO

text/0022-response-context.md
```diff
@@ -0,0 +1,98 @@
...
+* Should we rename `data` to `other` or something else to avoid confusion?
```
marandaneto, line 96: See https://github.com/getsentry/rfcs/pull/22/files#r986505011
Naming is not decided if we go with the databag.

text/0022-response-context.md
```diff
@@ -0,0 +1,98 @@
...
+* Are there any other fields that should be added to the `Response` interface?
```
marandaneto, line 97: After addressing https://github.com/getsentry/rfcs/pull/22/files#r986505011 this becomes less relevant.

text/0022-response-context.md
```diff
@@ -0,0 +1,98 @@
...
+* Some fields from the `Response` should be the very same as the `Request` interface (such as `method`, `url`, ...), should we just omit them?
```
marandaneto, line 98: The downside is that one needs to fill the `request` and `payload` always, which might be a problem if you don't have control/access over the requests but only the responses.
I suggest we keep it as it is, the duplicated fields are very small and just a few.

text/0022-response-context.md
marandaneto: If we go with arbitrary and retained fields, we don't need this databag at all.

text/0022-response-context.md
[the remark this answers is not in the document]
untitaker: If the purpose of the databag is to blindly forward all possible values into the database then we might as well accept arbitrary additional fields. However I would prefer if we did neither of those and type out fields we want to send properly, for documentation purposes.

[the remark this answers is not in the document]
marandaneto: it's one or the other, not both.

> arbitrary and retained fields (either this or data)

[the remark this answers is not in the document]
marandaneto: I'm fine not having both, but its a problem for back compatibility when adding new fields (older self-hosted versions will drop fields), and its also not that flexible since the user always know more than the SDK and can add more context that helps identify the issue.
I'm fine either way, either the databag or arbitrary fields are retained, any preference?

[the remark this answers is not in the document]
jjbayer: @untitaker we do have `retain = "true"` on all other contexts as far as I can see, so I would just do the same here.

https://github.com/getsentry/relay/blob/eace81d7f4ad8ecf3a94e76738e9e9bce9639116/relay-general/src/protocol/contexts.rs#L580-L582

@marandaneto I definitely prefer arbitrary fields over a dedicated databag. But the goal should be to add any new field to the schema eventually.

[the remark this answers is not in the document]
marandaneto: Sounds good, I will remove the databag from the RFC.

text/0022-response-context.md
```diff
@@ -0,0 +1,145 @@
...
+      "status_code": 500,
+      "is_redirect": false,
+      "body_size": 1000,
+      "arbitrary_field": "arbitrary" // arbitrary and retained fields for backwards compatibility when adding new fields
+    }
+  }
+}
+```
+
+The `Response` interface should be a mimic of the [Request](https://develop.sentry.dev/sdk/event-payloads/types/#typedef-Request) spec with additional fields.
```
jjbayer, line 95:
```suggestion
```

jjbayer (approved):

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+}
+```
+
+* `type`: `response` as `String`.
+* `cookies`: Can be given unparsed as `String`, as `Dictionary`, or as a `List of Tuples`.
+* `headers`: A `Dictionary` of submitted headers, this requires a special treatment in the data scrubbing rules.
+* `status_code`: The HTTP status code, `Integer`.
+* `is_redirect`: A `Boolean` indicating if the response was a redirect.
```
jan-auer, line 99: To clarify: this is `false` if the request was _redirected_?

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+      "body_size": 1000,
+      "arbitrary_field": "arbitrary" // arbitrary and retained fields for backwards compatibility when adding new fields
+    }
+  }
+}
+```
+
+* `type`: `response` as `String`.
+* `cookies`: Can be given unparsed as `String`, as `Dictionary`, or as a `List of Tuples`.
+* `headers`: A `Dictionary` of submitted headers, this requires a special treatment in the data scrubbing rules.
```
jan-auer, line 97: Ideally we mention that these share an identical schema with the `request` interface. For instance, headers can be a list of tuples since it's valid for headers to repeat.

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+* `status_code`: The HTTP status code, `Integer`.
+* `is_redirect`: A `Boolean` indicating if the response was a redirect.
+* `body_size`: A `Number` (absolute/positive) indicating the size of the response body in bytes.
+* `inferred_content_type`: A `String` indicating the inferred content type of the response.
```
jan-auer, line 101: The `inferred_content_type` is set during ingestion (Relay) by inferring it from the `content-type` header or the request/response payload, and is not supposed to be ingested. Hence, the name "inferred" and this is also why it is [not documented](https://develop.sentry.dev/sdk/event-payloads/request/).

Since we usually have headers, there should not be a need to ingest the content type explicitly.

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+
+The `data` field won't be added to the `Response` interface, a phase 2 of this RFC will propose add Request and Response bodies are sent as attachments.
+
+## Must have
+
+A tag should be created for `Request#url` and `Response#status_code` fields, people should be able to search for events with a specific `url` and/or `status_code`, also to alert on them.
+
+# Drawbacks
+
+The `Response` interface is PII sensitive, so we need to be careful about how we scrub it.
```
jan-auer, line 113: Would it make sense to define the scrubbing behavior in the field list above?

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+
+# Appendix
+
+## Removed Proposals
```
jan-auer, line 117: These options are now twice in this document: once under "Options considered" and once here.

text/0022-response-context.md
marandaneto: It was like that, @jjbayer prefered to remove it, https://github.com/getsentry/rfcs/pull/22/commits/bd0728d9eac30e365723fa9ecb78d20899bcdd9b

text/0022-response-context.md
marandaneto: Makes sense, removed it.

text/0022-response-context.md
```diff
@@ -0,0 +1,144 @@
...
+* `body_size`: A `Number` (absolute/positive) indicating the size of the response body in bytes.
+* `inferred_content_type`: A `String` indicating the inferred content type of the response.
+
+The `url`, `method`, `query_string`, `fragment`, `env` fields are not part of the `Response` interface and they should be set under the `Request` field, even if inferred from the HTTP response in case you don't have control over the HTTP Request object.
+
+The `data` field won't be added to the `Response` interface, a phase 2 of this RFC will propose add Request and Response bodies are sent as attachments.
+
+## Must have
+
+A tag should be created for `Request#url` and `Response#status_code` fields, people should be able to search for events with a specific `url` and/or `status_code`, also to alert on them.
```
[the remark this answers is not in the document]
marandaneto: I thought that creating tags is the way to index them, if there's another way, happy to go for it.
I changed the sentence a bit.

text/0022-response-context.md
marandaneto: The fields that have PII are similar to the request one, so I guess it's fine.
I added one sentence about this tho, mainly `cookies` and `headers`.

text/0022-response-context.md
marandaneto: Makes sense, removed the section `Options Considered`.

text/0022-response-context.md
marandaneto: Its already there in my last push.

text/0022-response-context.md
jan-auer: If this can be inferred directly from the status code, then we could get rid of this attribute. Otherwise, we should also have more flags like `is_success`, `is_server_error`, `is_client_error` -- or even an interpreted status field. I believe this is a bit redundant, however.

jan-auer (approved):
Good to go once the comment on on `is_redirect` is resolved. Thank you!

text/0022-response-context.md
marandaneto: makes sense, removing it.

PR #24, opened 2022-10-04
Create 0023-improve-fingerprinting-for-performance-issues

smeubank:
Request for ideas around improving fingerprinting to enable scaling out to more performance issue types

[Rendered RFC](https://github.com/getsentry/rfcs/blob/0023-improve-fingerprinting-for-performance-issues/text/0023-improve-fingerprinting-for-performance-issues)

### DIFF
--- a/text/0023-improve-fingerprinting-for-performance-issues
+++ b/text/0023-improve-fingerprinting-for-performance-issues
@@ -1,0 +1,48 @@
+* Start Date: 2022-10-04
+* RFC Type: feature 
+* RFC PR: <link>
+
+# Summary
+
+Performance issues has started with detecting N+1 issues and adding them to the issues page. This was a first proof of concept, now we are looking to add different types of performance issues to and group those events on the issues page. However there is an drawback of performance spans and transactions versus error events. Error events have titles and stack traces which can provided unique meta-data for [fingerprinting](https://docs.sentry.io/product/data-management-settings/event-grouping/fingerprint-rules/) and [grouping](https://docs.sentry.io/product/sentry-basics/grouping-and-fingerprints/). Today spans do not necessarily provide enough distinct metadata to be able to efficaciously group performance events.
+
+# Motivation
+
+We want to enable scaling out the number of different types of performance issues which can be detected and grouped.
+
+# Background
+
+We need communicate the issue around fingerprinting for performance issues, and suggest different ways to add more contextual data to improve fingerprinting. Alternatively we are open to solutions which work around fingerprinting.
+
+# Supporting Data
+
+Example performance issue type proposals (only ideation from this list and for context not commitments for delivery):
+* Slow DB spans
+* Consecutive similar spans
+* Identical spans
+*  N+1 DB Queries
+* N+1 API Calls
+* UI Freeze
+* Slow Assets
+* Component Re-renders
+* Main thread I/O + ANR/Slow Frame
+* Shader Compilation
+* Txn Deviations
+* Cancelled HTTP Requests / Flaky Network
+* Degraded performance for a demographic
+* Excessive DOM size
+
+# Options Considered
+
+1. Client side/SDK Includes application file name in span/transaction
+2. Can we fetch more info on Sentry server side from profiling if available 
+3. Could SDKs detect something and create a unique identifier artificially to empower fingerprinting?
+
+# Drawbacks
+
+Depending on where we put the logic for adding more contextual data, would we be adding too much overhead in SDKs/host application.
+
+# Unresolved questions
+
+* Can we specify why fingerprinting is required?
+* what the threshold would be for uniqueness to enable fingerprinting for performance issues?


text/0023-improve-fingerprinting-for-performance-issues.md
```diff
@@ -0,0 +1,50 @@
...
+* Txn Deviations
+* Cancelled HTTP Requests / Flaky Network
+* Degraded performance for a demographic
+* Excessive DOM size
+
+https://www.notion.so/sentry/dff3aad2a74940779c3e89d8d8af2537?v=03d1efc28741433c9a203b92fe4cef6e
+
+# Options Considered
+
+1. Client side/SDK Includes application file name in span/transaction
```
AbhiPrasad, line 39: I assume this is inspired from [opentelemetry source code attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#source-code-attributes), but we have to be careful about the runtime costs here.

text/0023-improve-fingerprinting-for-performance-issues.md
```diff
@@ -0,0 +1,50 @@
...
+2. Can we fetch more info on Sentry server side from profiling if available 
+3. Could SDKs detect something and create a unique identifier artificially to empower fingerprinting?
```
AbhiPrasad, line 41: One thing we can do is inject more thing during build time. For browser SDKs we have bundler plugins, in android we have the gradle plug-in, and in web SDKs we run in CI to create a release.

we can dynamically inject code based on parsed modules to attach metadata to events. If it’s expensive to generate this per event, we can send it as part of the release, and then resolve it server-side.

text/0023-improve-fingerprinting-for-performance-issues.md
```diff
@@ -0,0 +1,50 @@
...
+* Cancelled HTTP Requests / Flaky Network
+* Degraded performance for a demographic
+* Excessive DOM size
+
+https://www.notion.so/sentry/dff3aad2a74940779c3e89d8d8af2537?v=03d1efc28741433c9a203b92fe4cef6e
+
+# Options Considered
+
+1. Client side/SDK Includes application file name in span/transaction
+2. Can we fetch more info on Sentry server side from profiling if available
```
jernejstrasner, line 40: We talked about this for a bit and the main thing right now is that we don't have a simple way / any infrastructure built that makes it possible to leverage profiling data from other products. A big consideration going forward but for now it makes it unfeasible to rely on profiling for any of the work for next quarter.

text/0023-improve-fingerprinting-for-performance-issues.md
smeubank: I am not sure where the original inspiration came from, pull this from a notion doc with performance team, will try to get them involved here

gggritso: This is also inspired by our PHP SDK which surprisingly send code line numbers with the spans. We're worried about runtime costs, too, but it's worth looking into since it would _dramatically_ improve fingerprinting quality

text/0023-improve-fingerprinting-for-performance-issues.md
gggritso: Does Django (or other frameworks) have any useful runtime state that we could inject into the spans? e.g., when we were talking about detector evidence, there was an idea of parsing out the table names from the span description, and looking up the names of the corresponding Django models. Any relationship (even tenuous) between the span description and the code is helpful!

text/0023-improve-fingerprinting-for-performance-issues.md
```diff
@@ -0,0 +1,50 @@
...
+3. Could SDKs detect something and create a unique identifier artificially to empower fingerprinting?
+
+# Drawbacks
+
+Depending on where we put the logic for adding more contextual data, would we be adding too much overhead in SDKs/host application.
+
+# Unresolved questions
+
+* Can we specify why fingerprinting is required?
+* what the threshold would be for uniqueness to enable fingerprinting for performance issues?
```
gggritso, line 50: Fingerprinting is required for grouping problematic transactions into issues. The goal of grouping is to create exactly one issue for exactly one performance problem in the code. Fingerprints are the mechanism that accomplishes this for both error events and transaction events. Each event is "hashed" into a string fingerprint. Events with the same fingerprint are grouped and become an Issue!

Fingerprinting transaction events is hard because we don't have stack traces. We can't determine the code location precisely, so have to infer it using spans. When fingerprinting is too strict, we end up creating lots of issues even though the transactions come from one piece of code. This causes notification spam. When fingerprinting is too loose we end up creating just one group for multiple unrelated problems. This causes issues to become unresolved when a user fixes just one of the underlying problems.

That's the overall threshold/heuristic we're striving towards: to create exactly one issue for one piece of code that causes the performance problems 🙏🏻

text/0023-improve-fingerprinting-for-performance-issues.md
AbhiPrasad: Ok so crazy idea. We could potentially walk the AST of any codebase and upload that information during CI, since Sentry often runs in CI to do things like upload release artifacts (debug symbols, sourcemaps) and create a release.

This means we can actually do this evidence mapping in Relay itself, since it can just use the release information to make better decisions about spans. For example, in Django we could upload all the models, views, controllers - and then map those automatically to certain spans/transactions based on their name. We could even upload line/col numbers during this process, and then resolve it like how we do stacktraces in the product.

@mitsuhiko this comes back to our discussions about uploading transaction name info during build time.

Also @benvinegar @HazAT comes back to our general discussions about build time insights, uploading information at build that helps enhance our production data.

k-fish: During CI / buildtime is a good idea 👍 we just have to make sure the venn diagram of overlapping performance / sdk / and people willing to setup CI tools from us makes this worth it as a happy path.

> This means we can actually do this evidence mapping in Relay itself

I think we'd have to be careful with performance costs since we'll potentially want to run this on processed events and pulling data into Relay would be expensive. Potentially we could do something like the projectoption cache sync, but we'd want fairly responsive cache updates upon pushing so we catch performance issues properly upon deploy instead of having a lag and making the `start time` of the issue not line up. Additionally, we would need to figure out if a non-AST evidence issue and an AST evidence issue could be merged so as not to have 2 different issues in the time between a cache update.

text/0023-improve-fingerprinting-for-performance-issues.md
k-fish: Yeah, it will be nice to leverage profiling in the future but I agree it's probably premature to plan for it in fingerprinting if we don't have the infrastructure setup yet.

text/0023-improve-fingerprinting-for-performance-issues.md
philipphofmann: On Cocoa, serializing the stacktrace for a single thread would be too much overhead for multiple spans during a transaction. We could do that for selected spans, and maybe we don't need the full stacktrace, but only the top n frames. If we only want to use this for fingerprinting and don't want to display the stacktrace, we could focus on only sending the bare minimum for symbolication to reduce the payload size.

text/0023-improve-fingerprinting-for-performance-issues.md
philipphofmann: We must ensure that those artificial fingerprints work across releases and multiple SDK installations. This could be a challenge.

PR #27, opened 2022-10-13
add rfc for manually disabling flaky tests

asottile-sentry:
this rfc details a process for manually disabling flaky tests

[Rendered RFC](https://github.com/getsentry/rfcs/blob/asottile-disable-flaky-tests/text/0027-manual-disabling-of-flaky-tests.md)

### DIFF
--- a/text/0027-manual-disabling-of-flaky-tests.md
+++ b/text/0027-manual-disabling-of-flaky-tests.md
@@ -1,0 +1,75 @@
+* Start Date: 2022-10-13
+* RFC Type: informational
+* RFC PR: https://github.com/getsentry/rfcs/pull/27
+
+# Summary
+
+flaky tests in `sentry` / `getsentry` significantly impact developer
+productivity.  ideally an automated system would improve this situation but
+in the interim this rfc aims to document a manual process for disabling
+flaky tests.
+
+# Motivation
+
+currently any test flake will require the entire job to be rerun, which
+usually wastes about ~20 minutes of developer time per test flake.
+
+to minimize disruptiveness, this rfc proposes manually disabling with a
+procedure.
+
+# Supporting Data
+
+https://sentry.io/organizations/sentry/dashboard/7997/?project=5350637&project=4857230&project=2423079&statsPeriod=14d
+
+# Options Considered
+
+## manual disabling process
+
+**a test should be disabled if it has flaked on the primary branch**
+
+- identify the owner of the test in question
+    - often `.github/CODEOWNERS` can help
+    - otherwise utilize `git log -- path/to/the/test/file`
+- find the corresponding sentry issue:
+    - for example: [django tests](https://sentry.io/organizations/sentry/issues/?limit=5&project=2423079&query=&sort=freq&statsPeriod=14d)
+- create a JIRA ticket (or if the team uses github issues, do that instead)
+    - subject: `disabled flaky test: <testname>`
+    - body: (something like this)
+
+      ```
+          the following test is flaky so it is being disabled:
+
+          here is the sentry issue: <<< link here >>>
+
+          ```
+          <<< test failure output here >>>
+          ```
+
+          please fix this test and re-enable it
+      ```
+
+- make a pull request to disable the test (make sure to reference the ticket)
+    - for python tests, the test can be decorated with:
+
+      ```diff
+      +@pytest.mark.skip(reason='flaky: TICKET-1234')
+       def test_which_is_flaky():
+      ```
+
+    - for js tests, the test can be disabled with:
+
+      ```diff
+      -it('does the thing', () => {
+      +it.skip('flaky: TICKET-1234: does the thing', () => {
+      ```
+
+## automated disabling process
+
+this would require engineering effort to solve so for now we're choosing a
+lightweight manual process
+
+one could imagine a system where either the test runner or a test collector
+knows historical data about tests and can automatically make decisions about
+ignoring known failures and/or ticketing them
+
+# Unresolved questions


text/0027-manual-disabling-of-flaky-tests.md
```diff
@@ -0,0 +1,73 @@
...
+# Options Considered
+
+## manual disabling process
+
+- identify the owner of the test in question
+    - often `.github/CODEOWNERS` can help
+    - otherwise utilize `git log -- path/to/the/test/file`
+- find the corresponding sentry issue:
+    - for example: [django tests](https://sentry.io/organizations/sentry/issues/?limit=5&project=2423079&query=&sort=freq&statsPeriod=14d)
+- create a JIRA ticket (or if the team uses github issues, do that instead)
```
wmak, line 33: Should we be consistent here to reduce the overhead in this process to make it easier for the party disabling the test?
ie. Create a jira project for all flakey tests, or always use github issues with a specific label?

Also this way the skip reasons are easy to connect back to their corresponding issue (instead of having to hunt both Jira & Github)

asottile-sentry: without assigning them directly to teams there won't be accountability and they won't be in team's sprint boards -- from experience if disabled tests aren't owned they won't get fixed

JonasBa:
@asottile-sentry I fully support the RFC, re-running flaky tests is a loss of time that requires unnecessary context switching and requires you to also remember to check on your test later if it flaked again, all of which disrupts productivity. 

One thing that comes to mind (and this is probably thinking too far ahead) is that we should probably have a guideline on when is it ok to disable a test. Specifically, what I am afraid is that since this is a behavioral change, some folks may opt to just restart a flaky test and if it passes just go on with their day like they have been doing so far instead of disabling the test. Experiences and opnions may also vary depending on the person dealing with the flaky test, probably more true for newly onboarded team members who may be a bit more reluctant to make an action like this. Maybe trivial, but having something like "when is it ok to disable a test" could probably help with empowering this behavior.

asottile-sentry:
> @asottile-sentry I fully support the RFC, re-running flaky tests is a loss of time that requires unnecessary context switching and requires you to also remember to check on your test later if it flaked again, all of which disrupts productivity.
> 
> One thing that comes to mind (and this is probably thinking too far ahead) is that we should probably have a guideline on when is it ok to disable a test. Specifically, what I am afraid is that since this is a behavioral change, some folks may opt to just restart a flaky test and if it passes just go on with their day like they have been doing so far instead of disabling the test. Experiences and opnions may also vary depending on the person dealing with the flaky test, probably more true for newly onboarded team members who may be a bit more reluctant to make an action like this. Maybe trivial, but having something like "when is it ok to disable a test" could probably help with empowering this behavior.

do you think "a sentry issue exists" is a good enough threshold?  imo if it has failed on master at least once it's flaky and should get fixed

text/0027-manual-disabling-of-flaky-tests.md
wmak: Fair enough 👍, should we push teams to publish where to put the issue at least?

asottile-sentry: yeah I think "ownership" is a bit of an unsolved problem here at sentry -- ideally there would be a consistent and reliable way to create an issue for a team but I think that's a bit orthogonal to this (though it does make this step the most difficult of this manual process)

JonasBa:
> > @asottile-sentry I fully support the RFC, re-running flaky tests is a loss of time that requires unnecessary context switching and requires you to also remember to check on your test later if it flaked again, all of which disrupts productivity.
> > One thing that comes to mind (and this is probably thinking too far ahead) is that we should probably have a guideline on when is it ok to disable a test. Specifically, what I am afraid is that since this is a behavioral change, some folks may opt to just restart a flaky test and if it passes just go on with their day like they have been doing so far instead of disabling the test. Experiences and opnions may also vary depending on the person dealing with the flaky test, probably more true for newly onboarded team members who may be a bit more reluctant to make an action like this. Maybe trivial, but having something like "when is it ok to disable a test" could probably help with empowering this behavior.
> 
> do you think "a sentry issue exists" is a good enough threshold? imo if it has failed on master at least once it's flaky and should get fixed

Yes, imo that should be enough to warrant disabling the test.

wedamija (commented):
One thing that would be nice is some procedure for following up to understand how many flakey tests are being fixed vs being left disabled forever. I suspect it will mostly end up being the latter, and coverage will reduce over time.

asottile-sentry:
> One thing that would be nice is some procedure for following up to understand how many flakey tests are being fixed vs being left disabled forever. I suspect it will mostly end up being the latter, and coverage will reduce over time.

it's ~fairly easy to find these with `git grep` alone:

```console
$ git grep -E '(pytest\.mark\.skip\(|it\.skip)'
static/app/plugins/components/pluginIcon.spec.jsx:  it.skip('renders with default icon with invalid plugin id', function () {
static/app/views/dashboardsV2/create.spec.tsx:    it.skip('can create with new widget', async function () {
tests/acceptance/test_account_settings.py:    @pytest.mark.skip(
tests/acceptance/test_organization_dashboards.py:    @pytest.mark.skip(reason="flaky behaviour due to loading spinner")
tests/acceptance/test_organization_events_v2.py:    @pytest.mark.skip(reason="causing timeouts in github actions and travis")
tests/acceptance/test_organization_events_v2.py:    @pytest.mark.skip(reason="not done")
...
```

text/0027-manual-disabling-of-flaky-tests.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Supporting Data
+
+https://sentry.io/organizations/sentry/dashboard/7997/?project=5350637&project=4857230&project=2423079&statsPeriod=14d
+
+# Options Considered
+
+## manual disabling process
+
+**a test should be disabled if it has flaked on the primary branch**
```
armcknight, line 28: Why only the primary branch? This is something currently plaguing our PRs in sentry-cocoa, and we recently came to a decision that's close to what this RFC outlines. My current process is that if a test fails and then passes after I retry it in a PR, I immediately open an issue for it and a separate PR to disable the test.

asottile-sentry: yeah there's a bit of nuance to what I've said here -- usually you'll _find_ the flake on a PR and then it's useful to cross reference against the primary branch to make sure it has flaked there at least once

if it's only on the PR then it's probably the PR's fault

admittedly this rfc mostly has sentry / getsentry in mind and there's good instrumentation of the testsuite as sentry issues which "enforces" that this isn't a "my branch only" problem

armcknight: I get what you're saying about stronger verification that a PR didn't cause a failure. I can't see any way to assign cause to either a PR's changes or the test itself other than on a case-by-case basis, that'll be up to authors, reviewers and maintainers.

The perniciousness of a flaky test is the false negative (flaky test passing): you see it fail in a pr, then pass on a retry (indicating it's probably flaky) and then it passes again on the main branch after merging.

Once I see an indication that something is up with a test flipping back and forth, that's cause for further investigation. If a code change causes a failure, that test should _consistently_ fail, otherwise the test is unreliable. 

At that point you can't make any inference on whether it should fail on the main branch since flakiness is by nature indeterminate; it could go on passing for a long time just due to chance.

wedamija:
> > One thing that would be nice is some procedure for following up to understand how many flakey tests are being fixed vs being left disabled forever. I suspect it will mostly end up being the latter, and coverage will reduce over time.
> 
> it's ~fairly easy to find these with `git grep` alone:
> 
> ```
> $ git grep -E '(pytest\.mark\.skip\(|it\.skip)'
> static/app/plugins/components/pluginIcon.spec.jsx:  it.skip('renders with default icon with invalid plugin id', function () {
> static/app/views/dashboardsV2/create.spec.tsx:    it.skip('can create with new widget', async function () {
> tests/acceptance/test_account_settings.py:    @pytest.mark.skip(
> tests/acceptance/test_organization_dashboards.py:    @pytest.mark.skip(reason="flaky behaviour due to loading spinner")
> tests/acceptance/test_organization_events_v2.py:    @pytest.mark.skip(reason="causing timeouts in github actions and travis")
> tests/acceptance/test_organization_events_v2.py:    @pytest.mark.skip(reason="not done")
> ...
> ```

Right, but on top of this I'd like to understand have some sort of stats to understand how many tests we disable, how many actually get fixed, how many end up deleted, etc. Otherwise it's hard to assess what this change is going to have on the stability of our codebase.

armenzg (approved):
lgtm

joshuarli (approved):

JonasBa (approved):

mattgauntseo-sentry (approved):

PR #29, opened 2022-10-26
Added supergroups

mitsuhiko:
This RFC describes supergroups. This is the new location of an earlier document. Supergroups are a second order container that summarizes and groups issues together by common root cause. Where the first order group are events with the same fingerprint, a supergroup is a group of multiple issues that share the same root cause even if they might not contain exactly the same stack trace.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/supergroups/text/XXXX-supergroups.md)

### DIFF
--- a/text/XXXX-supergroups.md
+++ b/text/XXXX-supergroups.md
@@ -1,0 +1,275 @@
+* Start Date: YYYY-MM-DD
+* RFC Type: feature / decision / informational
+* RFC PR: <link>
+
+# Summary
+
+This RFC describes the motivation for supergroups.
+
+## Motivation
+
+When an error or exception is reported, it's desireable to fingerprint them so that the
+frequency of their ocurrence can be determined.  However what makes an error "the same"
+as another error?  Sentry has always run an algorithm to calculate the unique fingerprint
+of the error.  Any further error with the same fingerprint is fed into the same group and
+tallied up.  However no error is quite the same so the fingerprinting algorithm intentionally
+removed quite a bit of information or cleaned it up, to increase the chance of getting the
+same fingerprint.
+
+In doing so, there is a direct relationship to the creation of a Sentry issue to the
+calculation of the fingerprint.  This proposal wants to break this process up into two
+independent steps.
+
+It's evident from empiric evidence at Sentry that creating a overly precise fingerprint is
+problematic if each individual fingerprint creates a new issue in Sentry.  Because of this
+coupling the grouping algorithms have learned a few tricks over the years to disregard
+certain patterns that are known to be unstable between errors.  However there are a few
+fundamental issues that the grouping algorithm cannot address today.
+
+### Non Stack Fingerprints
+
+The primary way of creating fingerprints for errors are components of the stack trace.  There
+are however situations where a stack trace is unavailable or the stack trace does not have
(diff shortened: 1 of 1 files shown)


text/XXXX-supergroups.md
```diff
@@ -0,0 +1,275 @@
...
+This RFC describes the motivation for supergroups.
+
+## Motivation
+
+When an error or exception is reported, it's desireable to fingerprint them so that the
+frequency of their ocurrence can be determined.  However what makes an error "the same"
+as another error?  Sentry has always run an algorithm to calculate the unique fingerprint
+of the error.  Any further error with the same fingerprint is fed into the same group and
+tallied up.  However no error is quite the same so the fingerprinting algorithm intentionally
+removed quite a bit of information or cleaned it up, to increase the chance of getting the
```
markstory, line 16:
```suggestion
tallied up.  However no error is exactly the same so the fingerprinting algorithm intentionally
removed quite a bit of information or cleaned it up, to increase the chance of getting the
```

text/XXXX-supergroups.md
```diff
@@ -0,0 +1,275 @@
...
+  File "demo.py" in parse_uuid
+InvalidUuid
+```
+
+Here the error is in fact not in `parse_uuid`.  The function operates as expected, but the
+calling functions are failing to catch down the invalid input error.  We at this point do not
+even know yet if the `parse_request` or `load_images` function are supposed to catch down the
+error or if someone higher up in the stack is supposed to, but we do know that it somewhere made
+it to a point where a Sentry error was created.  In this case we would not want this to group
+together under `parse_uuid`.
```
markstory, line 103:
```suggestion
calling functions are failing to catch down the invalid input error. At this point we do not
yet know if the `parse_request` or `load_images` function are supposed to catch the
error or if someone higher up in the stack is supposed to, but we do know that it somewhere made
it to a point where a Sentry error was created.  In this case we would not want this to group
together under `parse_uuid`.
```

text/XXXX-supergroups.md
```diff
@@ -0,0 +1,275 @@
...
+with it and there are quite significant technical hurdles to this.  The proposal is to not touch this
+association but instead augment it on a higher level.  The idea is to actually create these Sentry
+issues still based on the unique fingerprint but to lower the consequences of such a group being
+created.  Related fingerprints could then at regular intervals be sweeped up in to larger super
+groups.
+
+This means that from a user's point of view the individual groups still exist, but the main way to
+interact with them would be larger groups that group around the already existing smaller groups.
+Since the events stay associated with their fingerprint created group the event itself is immutable
+and stays that way, but the merge just becomes an association of that group with the supergroup.
```
markstory, line 270: How do you see supergroups being represented in the datamodel? Are they an issue made of other 'similiar' issues?

mitsuhiko: @markstory i think supergroups might be just a special case of a re-built merge where merges are multiple hashes being associated to a group by some mutable mechanism. The difference between a merge and a supergroup in that sense is, that a supergroup carries additional metadata that gives the merge some meaning.

There is some related thinking happening here: https://github.com/getsentry/rfcs/pull/30

PR #31, opened 2022-10-27
feat(rfc): Add RFC on Scaling Code Owners

NisanthanNanthakumar:
This RFC details my proposals on how to address Code Owners scaling challenges

[Rendered RFC](https://github.com/getsentry/rfcs/blob/0028-scaling-codeowners/text/0028-scaling-codeowners.md)

### DIFF
--- a/text/0028-scaling-codeowners.md
+++ b/text/0028-scaling-codeowners.md
@@ -1,6 +1,6 @@
 * Start Date: 2022-10-27
 * RFC Type: feature
-* RFC PR: <link>
+* RFC PR: https://github.com/getsentry/rfcs/pull/31
 
 # Summary
 


text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+
+We have Code Mappings that map the transformation of the source code to the production build. Code Mappings help us reverse-engineer the file paths from an error’s stacktrace to source code file paths. 
+
+A customer can apply a Code Mapping to a CODEOWNERS file to get the file paths that should match against the stacktrace file paths. The transformed output is stored in `sentry_projectcodeowners` and is what Sentry will use to determine assignments.
+
+When Sentry ingests a new event, we go through each line in our transformed CODEOWNERS and see if each line has a match across any of the frames in the stacktrace. We only use the last matching line for the assignment.
+
+# Current System Constraints
+
+- We store the original and transformed CODEOWNERS file in a `sentry_projectcodeowners` row. However to prevent overloading Postgres from queries on every event, we cache the row in Memcache. Memcache can only support 1mb. Rows larger than 1mb will hit the database directly. Since we store the original and transformed CODEOWNERS in the same row, files larger than 500k chars will exceed this limitation.
```
armenzg, line 38: What's the value of storing the untranformed code owners? Is this referring the file that is available in the repository?

FYI I have noticed that we store comments from the original codeowners.

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+> The calculation is currently O((m*l + o)*n) time complexity
+
+To handle the quadratic time complexity, we can use Aho-Corasick Algorithm to bring the time complexity down to O(n). [https://pyahocorasick.readthedocs.io/en/latest/](https://pyahocorasick.readthedocs.io/en/latest/) Open to better algorithm suggestions.
+
+A simple optimization is to only try to find matches on the first `in_app` frame of a stacktrace, similar to Commit Context. This would bring the time complexity down to O(n).
+
+
+> We recalculate the assignment for each event
+
+We can store the assignments and skip the calculations for new events. We already started storing the Code Owners assignments in the `sentry_groupowner` table as a part of the Commit Context project. Since we currently make the assumption that our grouping algorithm is accurate, we can assume that the stacktrace for each event in a group will be the same. Currently, we store the Suspect Committer assignment for 7 days before we re-calculate on new events. And we can follow the precedent set by Commit Context.
```
armenzg, line 63: Why was 7 days chosen rather than 1 day or 30 days?
Will we take CODEOWNERS changes into account to recalculate? (invalidate the cache)

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+- On each event, the assignment calculation is currently O((m*l + o)*n) time complexity, where *m* is the # of lines in the CODEOWNERS file, *l* is the number of code mappings mapped to the project, *o* is the number of lines in the Ownership Rules and n is the number of lines in the stacktrace. Essentially quadratic.
+- We recalculate the assignment for each event, even if the Ownership Rules and CODEOWNERS file has not been updated.
+
+# Proposal
+
+First we need baseline metrics. We need to know how long it currently takes a worker to calculate the assignments with a 100k file.
+
+> Memcache can only support 1mb.
+
+To handle Memcache’s 1mb limit, we can zip to compress the rows and store that. However, we need to consider the memory limits of our workers during the unzip process. We need to consider the impact of (3mb CODEOWNERS raw file  + 3mb CODEOWNERS transformed file)* x number of code mappings applied to that file for a project is potentially unbounded. Two options:
```
armenzg, line 48: Here are some changes I suggest:
* Do not store the untransformed codeowners
* Do not store a row per code mapping but a row per repo
  * All code mappings for a specific repo should be applied at the same time (e.g. sentry/ and sentry_plugins/ for the Sentry repo)

This means we would have:
```diff
- 3mb CODEOWNERS raw file  + 3mb CODEOWNERS transformed file)* x number of code mappings
+                            3mb CODEOWNERS transformed file)* x number of repos with a code mapping
```

Also some other representation changes:
* Do not store comments
* Create a data structure that groups paths into buckets of teams
  * e.g. team A matches list of paths
  * Perhaps this should lead to a new table to store a more optimized structured to replace the current one
  * I would consider moving away completely from storing things raw but create a model(s) that could represent the information you care about
* Instead of storing team as strings, we should store them a ID to real teams. If we don't have team matches we don't have any specific team to match but everyone in the project

When reading the data from either the DB or memcache you can construct a more optimized structure based on what is stored.

For instance, you may want to store a tree where each node is part of the path while the leaves are either an asterisk or file name.

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+1. Users can upload a Code Owners File Size that's up to 3Mb in size
+2. The processing time to calculate assignments should not increase significantly. 
+
+# How it Works (simplified)
+
+We have Code Mappings that map the transformation of the source code to the production build. Code Mappings help us reverse-engineer the file paths from an error’s stacktrace to source code file paths. 
+
+A customer can apply a Code Mapping to a CODEOWNERS file to get the file paths that should match against the stacktrace file paths. The transformed output is stored in `sentry_projectcodeowners` and is what Sentry will use to determine assignments.
+
+When Sentry ingests a new event, we go through each line in our transformed CODEOWNERS and see if each line has a match across any of the frames in the stacktrace. We only use the last matching line for the assignment.
```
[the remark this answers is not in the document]
wedamija: We used to do first match, but we changed it to match github I believe

text/0028-scaling-codeowners.md
wedamija: I think it's necessary to be able to display the codeowners file in the ui. This is a holdover from the issue owners file, but it's probably still relevant here.

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Current System Constraints
+
+- We store the original and transformed CODEOWNERS file in a `sentry_projectcodeowners` row. However to prevent overloading Postgres from queries on every event, we cache the row in Memcache. Memcache can only support 1mb. Rows larger than 1mb will hit the database directly. Since we store the original and transformed CODEOWNERS in the same row, files larger than 500k chars will exceed this limitation.
+- On each event, the assignment calculation is currently O((m*l + o)*n) time complexity, where *m* is the # of lines in the CODEOWNERS file, *l* is the number of code mappings mapped to the project, *o* is the number of lines in the Ownership Rules and n is the number of lines in the stacktrace. Essentially quadratic.
+- We recalculate the assignment for each event, even if the Ownership Rules and CODEOWNERS file has not been updated.
```
wedamija, line 40: It's hard to avoid this to some extent, because we could have different events in a single issue.

text/0028-scaling-codeowners.md
wedamija: I suspect that loading a 3mb file from either the database or memcache will be pretty memory intensive as well. I think we should probably do some kind of profiling on how long it takes to fetch even a compressed file from memcache

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+
+To handle Memcache’s 1mb limit, we can zip to compress the rows and store that. However, we need to consider the memory limits of our workers during the unzip process. We need to consider the impact of (3mb CODEOWNERS raw file  + 3mb CODEOWNERS transformed file)* x number of code mappings applied to that file for a project is potentially unbounded. Two options:
+
+- We can implement a limit on the number of codemappings per (repository, project)
+- We can run the calculation in batches
+ 
+
+> The calculation is currently O((m*l + o)*n) time complexity
+
+To handle the quadratic time complexity, we can use Aho-Corasick Algorithm to bring the time complexity down to O(n). [https://pyahocorasick.readthedocs.io/en/latest/](https://pyahocorasick.readthedocs.io/en/latest/) Open to better algorithm suggestions.
```
wedamija, line 56: This seems like the right idea. I don't like that it is using pickle to serialize to disk though, I would prefer that we figure out how to serialize to json if we want to cache this instead of calculating it every time.

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
...
+
+A simple optimization is to only try to find matches on the first `in_app` frame of a stacktrace, similar to Commit Context. This would bring the time complexity down to O(n).
+
```
wedamija, line 59: I think this will break existing functionality, but really a product decision.

text/0028-scaling-codeowners.md
wedamija, line 63: We can't assume the stacktrace will be the same for all events. Merging breaks this assumption, there might be other things that also break it.

wedamija: That said, some kind of sampling here is probably fine

wedamija (commented):
Should these huge files be stored in the database at all? Or should we store them as a `File`?

mitsuhiko:
While memcache (particularly in the open source version of Sentry) will only support 1MB, our production memcached cluster runs with elevated slab sizes of 5MB. https://github.com/getsentry/getsentry/blob/cdf711591e0859c52ad06c29028c212ac6647ca8/getsentry/conf/settings/prod.py#L425-L432

text/0028-scaling-codeowners.md
mitsuhiko: Aho-corasick seems sensible. One thing I want to point out though is that this problem is similar to the user agent parsing problem we have in Relay where we ultimately decided against building large regex sets / aho-corasick states after benchmarking this. In particular actually building this in python might result in you spending more time building the aho-corasick internal states every time you match them than actually doing basic matches.

mitsuhiko: I would probably actually try to do the parsing and see if it is indeed faster. The best speedups we got on mass matching btw was using the Rust regex library which runs circles around the python one. We are currently using this already for the glob matching in Python which calls into a Rust implementation via the relay library.

text/0028-scaling-codeowners.md
```diff
@@ -0,0 +1,63 @@
+* Start Date: 2022-10-27
+* RFC Type: feature
+* RFC PR: https://github.com/getsentry/rfcs/pull/31
+
+# Summary
+
+Sentry supports GitHub and Gitlab Code Owners files for auto-assigning Issues. However, Sentry only supports 100K characters as the file size limit for Code Owners. GitHub supports a 3Mb file size for Code Owners.
```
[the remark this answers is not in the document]
markstory: 100_000 characters is between 100KB and 400KB depending on how much unicode is in the codeowners file.

text/0028-scaling-codeowners.md
markstory: Custom grouping fingerprints can also combine dissimilar stacktraces.

text/0028-scaling-codeowners.md
wedamija: One thing we should consider as well is that for codeowner files, most rules are exact matches or doing suffix matching, and look something like
`path/to/my/file.py`
`path/to/my/*`
and relatively few looks like
`*/to/my/file.py`
`path/*/my/file.py`

It might be simpler to build a relatively simple trie here that handles the majority of rules, and then just run the regexes for the rules that are the exception here. 
```
path
       \
       to
        \
         my
          /\
         *  file.py
```
If we hit a `*` in the trie then we just instantly pass, and then same thing if we hit a leaf node.

fpacifici: How big would the Trie for all projects be in memory? You could consider running this part of code owners detection in a dedicated celery task and load everything in memory at startup (not per event) and avoid the fetch of up to 1Mb from memcache per event. 
We can run Kubernetes pods with 10s of Gb of memory

Angelodaniel:
Hey Team, would the plan to find a match for the first stack frame, if none is found then Sentry will try to find a match for the second stack frame also apply when working **with only ownership rules**. I'm receiving that feedback from customers as well

text/0028-scaling-codeowners.md
JoshFerge: @NisanthanNanthakumar couple notes from our in person convo:
1. likely need to cache the regex values, as rust regex complation is very slow https://github.com/rust-lang/regex/blob/master/PERFORMANCE.md#thou-shalt-not-compile-regular-expressions-in-a-loop
2. definitely test rust code directly, see if python binding potentially could make it slow

PR #32, opened 2022-10-27
Improved API Tokens

mdtro:
Improved API tokens to provide customers an easier means of detecting accidentally leaked secrets by integrating with Github's Secret Scanning Service and other static analysis tools.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/mdtro/api-tokens/text/0032-improved-api-tokens.md)

### DIFF
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -152,6 +152,7 @@ To avoid the two different token versions, we could automatically prepend the pr
 
 # Unresolved questions
 
+- How do we deploy the necessary changes for the backend and frontend separately and in a backwards compatible way?
 - Implementation path for `APIApplication` secrets.
 - Is there a downside to allowing `null` in the additional columns?
 - What is the median time between token use?
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -133,9 +133,9 @@ Second, we will need to secure the tokens. This involves four primary goals.
 - Newly created user API application tokens have the `sntrya_` prefix
 - We encourage users in the UI via a notification/banner to recreate their tokens in order to get new ones with a prefix
 
-1. Nullable `hashed_token` and `hashed_refresh_token` fields are added to the `ApiToken` model
-2. The `save()` method on `ApiToken` is updated to calculate and store the token's SHA-256 hash in `hashed_token`.
-3. Update the `UserAuthTokenAuthentication` middleware to:
+1. Nullable `hashed_token` and `hashed_refresh_token` fields are added to the `ApiToken` model [#65300](https://github.com/getsentry/sentry/pull/65300) [#66679](https://github.com/getsentry/sentry/pull/66679)
+2. The `save()` method on `ApiToken` is updated to calculate and store the token's SHA-256 hash in `hashed_token`. [#67969](https://github.com/getsentry/sentry/pull/67969)
+3. Update the `UserAuthTokenAuthentication` middleware to: [#67969](https://github.com/getsentry/sentry/pull/67969)
 
    1. Caculate the SHA-256 hash and use the hash value for the table lookup on the `hashed_token` or `hashed_refresh_token` column.
    2. If the hash is not found, use the plaintext token for the table lookup on the `token` or `refresh_token` column.
@@ -147,16 +147,14 @@ Second, we will need to secure the tokens. This involves four primary goals.
    > _It's important to note that this does not update the token to the new prefixed format._
 
 4. A nullable `token_type` field is added to the `ApiToken` model. It should accept a limited set of choices to indicate whether the token is `sntryu_`, `sntrya_`, etc. A _null_
-   value would indicate a legacy token that is not prefixed regardless of whether it is a user or application token.
-5. A new _class method_ named `create_token(..)` is created on the `ApiToken` model. This method will return the plaintext token, plaintext refresh token, and `ApiToken`
-   instance to the caller. The plaintext token will be needed to display to the user temporarily in the UI.
-6. Calls to `ApiToken.objects.create(..)` should be replaced with the new `ApiToken.create_token(..)` method.
-7. API endpoints that retrieve the full plaintext token value should be updated to no longer do so. This should only be available on creation.
-8. A notification/banner in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
+   value would indicate a legacy token that is not prefixed regardless of whether it is a user or application token. [#65684](https://github.com/getsentry/sentry/pull/65684)
+5. Adjust `create(...)` method on the `ApiToken` model to hash the plaintext token values and temporary access to the plaintext values. [#68148](https://github.com/getsentry/sentry/pull/68148)
+6. API endpoints that retrieve the full plaintext token value should be updated to no longer do so. This should only be available on creation. [#68148](https://github.com/getsentry/sentry/pull/68148)
+7. A notification/banner in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
 
 Next, any remaining legacy tokens that do not have hashed values will need to be handled:
 
-1. As a Django migration, a bulk operation is executed to update all remaining legacy tokens in the database.
+1. As a Django migration, a bulk operation is executed to update all remaining legacy tokens in the database. [#71728](https://github.com/getsentry/sentry/pull/71728)
    - This operation will hash the legacy `token` and `refresh_token` values and store them in the database.
    - It does **not** update the tokens to the new format.
 
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -149,7 +149,8 @@ Second, we will need to secure the tokens. This involves four primary goals.
 5. A new _class method_ named `create_token(..)` is created on the `ApiToken` model. This method will return the plaintext token, plaintext refresh token, and `ApiToken`
    instance to the caller. The plaintext token will be needed to display to the user temporarily in the UI.
 6. Calls to `ApiToken.objects.create(..)` should be replaced with the new `ApiToken.create_token(..)` method.
-7. A notification/banner in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
+7. API endpoints that retrieve the full plaintext token value should be updated to no longer do so. This should only be available on creation.
+8. A notification/banner in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
 
 Next, any remaining legacy tokens that do not have hashed values will need to be handled:
 
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -124,7 +124,7 @@ First, we will need to support naming and showing a mostly obfuscated token in t
 4. Change the `ApiToken` serializer to send the `token_last_characters` in the response for use in the frontend. [#63473](https://github.com/getsentry/sentry/pull/63473)
 5. Change the frontend to use the new `token_last_characters` value and show an obfuscated token in the UI. [#63485](https://github.com/getsentry/sentry/pull/63485)
 6. Update the backend serializer for `ApiToken` to accept and return an optional `name` field. [#64493](https://github.com/getsentry/sentry/pull/64493)
-7. Update the frontend to support creation of a token with a `name` and displaying of the `name` when listing tokens.
+7. Update the frontend to support creation of a token with a `name` and displaying of the `name` when listing tokens. [#64509](https://github.com/getsentry/sentry/pull/64509)
 
 Second, we will need to secure the tokens. This involves four primary goals.
 
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -117,7 +117,7 @@ This also allows for a smooth transition of self-hosted instances keeping pace w
 First, we will need to support naming and showing a mostly obfuscated token in the UI to help users identify them.
 
 1. The frontend is updated to no longer display the token value for existing tokens. [#61941](https://github.com/getsentry/sentry/pull/61941)
-2. Nullable `name` and `token_last_characters` fields are added to the `ApiToken` model.
+2. Nullable `name` and `token_last_characters` fields are added to the `ApiToken` model. [#58945](https://github.com/getsentry/sentry/pull/58945)
    - New `ApiToken`s created should automatically have the `token_last_characters` populated based on an option.
      The option is required in order to properly test the upcoming backfill migration. [#62972](https://github.com/getsentry/sentry/pull/62972)
 3. A backfill migration is created and ran to fill in the `token_last_characters` for all `ApiToken` entries. [#63342](https://github.com/getsentry/sentry/pull/63342)
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -17,10 +17,10 @@
     - [Secret Verification Service](#secret-verification-service)
 - [Background](#background)
 - [Options Considered](#options-considered)
-  - [Option #1](#option--1)
-  - [Option #2](#option--2)
+  - [Option #1](#option-1)
+  - [Option #2](#option-2)
     - [Drawbacks](#drawbacks)
-  - [Option #3](#option--3)
+  - [Option #3](#option-3)
     - [Drawbacks](#drawbacks-1)
 - [Unresolved questions](#unresolved-questions)
 
--- a/text/0032-improved-api-tokens.md
+++ b/text/0032-improved-api-tokens.md
@@ -33,7 +33,7 @@ Improve on Sentry's API token implementation to a more secure pattern. We will h
 1. Using hashed values in the database
 2. Only display the token once to the end user upon creation
 3. Allow users to _name_ the tokens ([#9600](https://github.com/getsentry/sentry/issues/9600))
-4. Use a predictable prefix to integrate with various secret scanning services (ex. Github's Secret Scanning)
+4. Use a predictable prefix and suffix to integrate with various secret scanning services (ex. Github's Secret Scanning)
 
 # Motivation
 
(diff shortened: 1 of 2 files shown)


text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,163 @@
...
+- Add a Django migration including the `hashed_token`, `name`, and `token_version` fields.
+  - columns should allow `null` values for now.
+- Remove displaying the plaintext token in the frontend, and only display newly created tokens once.
+- Implement new logic to generate the new token pattern.
+  - The `token_version` should be set as `2`.
+- Implement a backwards compatible means of verifying legacy tokens. This should:
+  - Verify the legacy token is valid/invalid.
+  - If valid, hash the legacy token and store it in the table.
+  - Remove the plaintext representation of the token in the table.
+  - Mark the `token_version` as `1`.
```
markstory, line 123: This seems like a task we could do incrementally as tokens are used. During authentication we know if a token is v1, and could either replace the token synchronously or dispatch a task to update the token.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,163 @@
...
+- A notification/warning in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
+
+The `n+1` Sentry version should run in production for sufficient time to update the majority of the rows in the database as the legacy tokens are _used_ by users.
+
+### `n+2`
+
+- Add a Django migration to generate the `hashed_token` value for any tokens in `sentry_apitoken` that do not already have a hashed value.
+  - Rows matching this are legacy tokens that were not used during the transition period.
+  - The `token_version` should be set to `1` for these rows.
```
markstory, line 132: I don't think we should try to straddle multiple token versions for a long period of time. If we update all the currently active tokens incrementally could we follow that up with a bulk operation to replace all the remaining tokens?

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,163 @@
...
+- This is less than ideal because of the potential performance impact while the migration mutates each row in the table. Potential impacts to self-hosted installs would be unknown as well.
+
+## Option #3
+
+- Users would not be able to benefit from the Github Secrets Scanning since they would still be using their 64 character alphanumeric string without the prefix.
+- Authentication views/logic would become more complex with branching code to handle the with and without prefix cases.
+
+# Unresolved questions
+
+- How do we deploy the necessary changes for the backend and frontend separately and in a backwards compatible way?
```
markstory, line 155: Generally we deploy backend changes first, and then frontend changes.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,163 @@
...
+- Implementation path for `APIApplication` secrets.
+- Is there a downside to allowing `null` in the additional columns?
+- What is the median time between token use?
+  - _This value could be used to inform how long we wait between versions for the migration that will edit pending rows in the database._
+- Do we want to support [API Keys](https://docs.sentry.io/api/auth/#api-keys) as well, even though they are considered a legacy means of authentication?
```
markstory, line 160: I would love for us to not have API keys, but that is a separate discussion.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,163 @@
...
+- Are there additional secret types we would like to apply this to?
+- Should the secret verification service live within the monolith or run as a separate service entirely?
```
markstory, line 162: If it is going to change the monolith database, we should keep the code in the monolith. Otherwise we end up with an integration database :cry:

Large batches are fine we can process the batch in slices asynchronously with celery.

misha-sentry:
a few questions:
- "Verify the legacy token is valid/invalid" - are you just checking if its expired or not?
- For legacy tokens, you are proposing to store it in a hashed format, and will you also have it prepended with the token type identifier?
- What backwards compatibility issues need to be considered for making changes to the frontend (ie. not displaying the token after the one time after it is created)?

text/0032-improved-api-tokens.md
mdtro: Yes, exactly! 😄 I'll try to re-word this to make it more clear.

text/0032-improved-api-tokens.md
mdtro: +1

This is the goal. I'll update the proposal to clarify, but in general we would:

1. Release version 22.11.0 with the incremental token updates as they are used.
2. Release version 22.12.0 with the bulk operation to update any tokens that were not updated between the release dates.

mdtro:
@misha-sentry 

> "Verify the legacy token is valid/invalid" - are you just checking if its expired or not?

This would be our standard verification. For example, does the token exist and is it not expired?

> For legacy tokens, you are proposing to store it in a hashed format, and will you also have it prepended with the token type identifier?

Only the hashed format. The plaintext token value that would contain the prefix would not exist in the database.

> What backwards compatibility issues need to be considered for making changes to the frontend (ie. not displaying the token after the one time after it is created)?

We deploy frontend and backend separately. Deploying both at the same time is high risk since rollouts happen gradually. 
Instead, we'll have to figure out a path where they can each be deployed separately without breaking each other. I'll need to partner with the engineering teams to figure that one out.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+---
+
+# Summary
+
+Improve on Sentry's API token implementation to a more secure pattern. We will have three major goals:
+
+1. Using hashed values in the database
+2. Only display the token once to the end user upon creation
+3. Allow users to _name_ the tokens ([#9600](https://github.com/getsentry/sentry/issues/9600))
+4. Use a predictable prefix to integrate with various secret scanning services (ex. Github's Secret Scanning)
```
bukzor-sentryio, line 36: Can we get a suffix as well?
It will make scanning/matching much more precise.
One or (preferably) two bytes would be sufficient.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+
+Existing API tokens, should be hashed and their original value discarded.
+
+A notice in the Sentry UI should be presented to suggest the user rotate and generate a new token to the new and improved version.
+
+### Token Format
+
+We need a predictable token format in order to integrate properly with secret scanning services. Our current format is a 64 character alphanumeric string. This is insufficient and would likely produce a high amount of false positives in tooling like [TruffleHog](https://github.com/trufflesecurity/trufflehog), [detect-secrets](https://github.com/Yelp/detect-secrets), [Github's Secret Scanning](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning), etc.
+
+The suggested pattern is `snty[a-zA-Z]_[a-zA-Z0-9]{64}`. The character _after_ `snty` will be used to identify the token type.
```
bukzor-sentryio, line 63: At the risk of bikeshedding, my suggestion:  `\bsn[a-zA-Z]_[a-zA-Z0-9]{64}ty\b`

bukzor-sentryio, line 63: Ergonomics would benefit greatly from some internal hyphenation.
Since your bitfield is twice the size of uuid, a doubled uuid scheme would make sense, to me:

`snX-fe611dff-98e0-423a-aa9a-d1f629b0f869-05b934cf-61d8-4c1e-90ae-1d8ff290cca7ty'

The static suffix provides less (but nonzero) benefit in this case.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+For example:
+
+- `sntyu_a1b2c3...` - identifies a **user** API token
+- `sntya_a1b2c3...` - identifies an **API Application** token
+
+### Backend Token Hygiene
+
+API tokens should be stored in the database with a cryptographically secure hash (minimum SHA-256). Salting is not necessary since these values are sufficient in length.
+
+An additional column identifying the API token version will be added in order to correctly identify legacy API tokens.
```
bukzor-sentryio, line 74: Arguably these are separate datatypes that deserve separate columns.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+
+We can help customers further protect their accounts and data by providing a means of auto-detecting leaked API tokens.
+
+# Options Considered
+
+## Option #1
+
+This option strives to limit impact to the database by avoiding large bulk operations. To do this we'll distribute the complete implementation over two features. This also allows for a smooth transition of self-hosted instances keeping pace with releases.
+
+**In the first release:**
```
bukzor-sentryio, line 107: I would have expected another, prior, backend-only release that ensures the APIs are forward-backward compatible with these changes.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+3. A notification/warning in the UI should be displayed recommending users recreate their tokens, resulting in the new token version.
+4. New tokens are only displayed once at creation time to the user.
+5. New tokens are created following the new format.
+   - A Django migration will be needed to add fields: `hashed_token`, `version`, and `name`.
+   - legacy tokens will be `version = 1` and new tokens `version = 2`
+6. As old tokens are used, they are hashed and stored in the database.
+   - This does not update the token to the new format.
+   - The plaintext token value should be removed in this action.
+
+**In the second release:**
```
bukzor-sentryio, line 120: I would have expected two more releases afterward to 1) remove the special case frontend code 2) remove the API compatibility code.

Otherwise these become a permanent technical burden. It may be small but many such small burdens add up over time.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+
+1. As a Django migration, a bulk operation is executed to update all remaining legacy tokens in the database.
```
bukzor-sentryio, line 122: Would adding a new hashed-legacy-key column, populating it, then removing the unhashed column be cheaper?
The new-column population could be done with arbitrary slowness. to avoid adding appreciable load to systems.
Then the bulk update would only need to drop the old column (after APIs stop using it).

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+### Drawbacks
+
+- Users would not be able to benefit from the Github Secrets Scanning since they would still be using their 64 character alphanumeric string without the prefix.
+- Authentication views/logic would become more complex with branching code to handle the with and without prefix cases.
+
+# Unresolved questions
+
+- How do we deploy the necessary changes for the backend and frontend separately and in a backwards compatible way?
+- Implementation path for `APIApplication` secrets.
+- Is there a downside to allowing `null` in the additional columns?
```
bukzor-sentryio, line 147: Every use of the column will need to decide, agree on the semantics of a null value.
A guarantee of meaningful value (i.e. no nulls) will appreciably reduce ambiguity.

ghost (commented):
Please feel free to ack and resolve any of my ideas that have more cost than benefit.
Similarly if they're founded on my own ignorance or misapprehension :)

ghost (commented):
I believe this RFC actually contains two ideas that could be implemented separately:

* keys at rest should be hashed -- a backend-only change
* a new format for displaying and receiving secrets -- with care this can also be backend-only: API responses will only use the new form, while requests can optionally use the old during a (relatively lengthy) transition period.

I would likely want to separate these projects so that they can proceed each at their own pace.

text/0032-improved-api-tokens.md
mdtro: Are you referring to the User API Token vs. the API Application Token?
If so, those are stored in two different tables. :)

text/0032-improved-api-tokens.md
mdtro: I re-wrote this whole section to clarify. Let me know your thoughts. :)

text/0032-improved-api-tokens.md
mdtro: Apologies, I had a typo here. It would be distributed "over two releases". For the sake of this PR (and maybe I should clarify this in the document somewhere), I just looking at the major releases, for example, 22.11.0 and 22.12.0. This doesn't mean a single PR for each release, but enough incremental changes and PRs to make sure we do this in a backwards compatible way.

mitsuhiko:
@markstory this is largely unrelated to this RFC but I wonder if there would be some benefit to encoding the silo into the token. That way we could route based on the information in the token from the central silo. Since we are now talking about re-issuing tokens anyways we might take that opportunity.

markstory:
> I wonder if there would be some benefit to encoding the silo into the token. That way we could route based on the information in the token from the central silo. 

The `ApiToken` models we currently have don't include any organization links which makes encoding a region into them hard as the tokens would work for any organization the user has access to.

text/0032-improved-api-tokens.md
mdtro: I'll keep API keys outside of the scope of this RFC.

text/0032-improved-api-tokens.md
mdtro: I like the idea of the suffix, but I'm going to make some minor adjustments. 😄 

I'm still a fan of keeping the prefix as `snty`. This gives the opportunity for the regex to bail out early, and I'm making an assumption there's higher potential for more `sn`'s out there than `snty`. So we don't have two `ty`'s lol, I've opted to use `io` as our suffix.

If there are stronger feelings/reasonings about this, I'm glad to consider changing the format again.

mdtro: I wouldn't expect most would be typing this key in manually, but rather a copy/paste?

I don't really hold a strong opinion here on the hyphens. Arguably, if most are copying and pasting the value the impact won't be felt, so it might just be a win-win here?

text/0032-improved-api-tokens.md
mdtro: An additional column will be used to represent the API token version.

text/0032-improved-api-tokens.md
mdtro: Opting to keep with just the prefix.

text/0032-improved-api-tokens.md
mdtro: Opting for just the prefix.

text/0032-improved-api-tokens.md
mdtro: See https://github.com/getsentry/rfcs/pull/32#discussion_r1010581248. I'll update the document to make this more clear.

text/0032-improved-api-tokens.md
bukzor-sentryio: I meant that the old and new tokens are separate ideas.
If you use a new, separate column for new tokens, it becomes possible for users to have both during a transition period. This can be very handy in certain transition scenarios.
I'm not trying to hardsell the idea; I just want to communicate it clearly :)

ghost (approved):
Approved: I'm not qualified to approve the RFC itself, but this PR is for a draft-status rfc.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+### Token Format
+
+We need a predictable token format in order to integrate properly with secret scanning services. Our current format is a 64 character alphanumeric string. This is insufficient and would likely produce a high amount of false positives in tooling like [TruffleHog](https://github.com/trufflesecurity/trufflehog), [detect-secrets](https://github.com/Yelp/detect-secrets), [Github's Secret Scanning](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning), etc.
+
+The suggested pattern is `snty[a-zA-Z]_[a-zA-Z0-9]{64}io`. The character _after_ `snty` will be used to identify the token type.
+
+For example:
+
+- `sntyu_a1b2c3...` - identifies a **user** API token
+- `sntya_a1b2c3...` - identifies an **API Application** token
```
oioki, line 68: We already have `sntrys_` prefix for [org auth tokens](https://github.com/getsentry/rfcs/pull/105). I think we can use the same prefix and derive the type of token from the payload (aka "facts").

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+
+# Options Considered
+
+## Option #1
+
+This option strives to limit impact to the database by avoiding large bulk operations. To do this we'll distribute the complete implementation over two releases. This also allows for a smooth transition of self-hosted instances keeping pace with releases.
+
+**In the first release:**
+
+1. The frontend is updated to no longer display the token value for existing tokens.
```
oioki, line 109: We will also need to deprecate `/api/0/api-tokens/` API endpoint with some notice period.

text/0032-improved-api-tokens.md
mdtro: The `sentrys_` prefix is for **s**tructured tokens. We probably should have going with `o` or something to indicate an org auth token.

I think it'll be a bit before we we can derive the token type from the facts within the payload. For example, there isn't a field in the current org auth token that indicates the token type.

I think it's important to keep these separate prefixes especially when it comes to building out our service that integrates with GitHub's secret scanning program. With them, we do not need to _decode_ a token to determine it's type, avoiding the extra processing. We can look at the first characters, determine the token type, perform a quick hashing operation, and then check for matches.

text/0032-improved-api-tokens.md
mdtro: A `null` is necessary here to keep the migrations safe as far as I can tell.

text/0032-improved-api-tokens.md
```diff
@@ -0,0 +1,153 @@
...
+
+Existing API tokens, should be hashed and their original value discarded.
+
+A notice in the Sentry UI should be presented to suggest the user rotate and generate a new token to the new and improved version.
+
+### Token Format
+
+We need a predictable token format in order to integrate properly with secret scanning services. Our current format is a 64 character alphanumeric string. This is insufficient and would likely produce a high amount of false positives in tooling like [TruffleHog](https://github.com/trufflesecurity/trufflehog), [detect-secrets](https://github.com/Yelp/detect-secrets), [Github's Secret Scanning](https://docs.github.com/en/code-security/secret-scanning/about-secret-scanning), etc.
+
+The suggested pattern is `snty[a-zA-Z]_[a-zA-Z0-9]{64}io`. The character _after_ `snty` will be used to identify the token type.
```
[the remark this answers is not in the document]
mdtro: I'll bring this up in TSC, but I don't have a strong preference here. The additional underscores and hyphens can cause issues when copying and pasting though, so unless there's a strong benefit to them -- I'd rather omit them.

text/0032-improved-api-tokens.md
mdtro: Good point, or at least update it to no longer return the full token value when just listing them (post token creation). I'll update the RFC.

BARRYPMARSHALL:
Hi! I noticed you are looking for security scanning help.

I run **FREE33** — a free-to-try microservices platform with 11 services including:
* **Code Review** — 11 languages (Python, JS, TS, Go, Rust, Java, Ruby, PHP, C/C++), ruff + eslint
* **Security Scanning** — secrets, API keys, SQLi, XSS, credential leakage
* **WASM Compiler**, **JSON Validator**, **Diff Engine**, **API Mocker**, and more

All services offer free trials, no signup required. Micro-credit pricing starts at 0.0003 cr per call.

**Live endpoint:** https://ref-used-out-alliance.trycloudflare.com
**GitHub repo:** https://github.com/BARRYPMARSHALL/free33

Let me know if FREE33 can help — happy to walk you through integration!

PR #33, opened 2022-11-02
View Hierarchy

brustolin:
# Summary

Allow SDKs to send View hierarchy of the application UI.

# Motivation

Being able to see what the user was looking at is a good way to found potential problems in a mobile application. 
This RFC proposes a way to send information from the UI state of an application, known as "view hierarchy".

[Rendered RFC](https://github.com/getsentry/rfcs/blob/feat/view-hierarch/text/0033-view-hierarchy.md)


marandaneto:
Add RFC ID to the README, See example https://github.com/getsentry/rfcs/pull/22/files

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+* RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
+
+# Summary
+
+Allow SDKs to send View hierarch of the application UI.
+
+# Motivation
+
+Being able to see what the user was looking at is a good way to found out problems in the application. 
+This RFC proposes a way to send information from the UI state of an application, known as "view hierarch".
```
marandaneto, line 12: I'd elaborate a bit more here, it's pretty straightforward but assumes everyone knows what a view hierarchy is and how this is a good thing, linking to the other PoCs that have screenshots, etc.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
+* Start Date: 2022-11-02
+* RFC Type: feature
+* RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
```
marandaneto, line 3: Status and driver are missing

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+    "width": 100, 
+    "height": 100,
+    "depth": 100, //if 3D
+    "x": 0,
+    "y": 1,
+    "z": 2, //if 3D.
+    "visible": true|false,
+    "alpha": 1,
+    "{extra_properties}": "{property value}" //adicional information by platform
+}
```
marandaneto, line 32: We agreed that this would be a List since there are devices with multiple screens/views.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+# Summary
+
+Allow SDKs to send View hierarch of the application UI.
+
+# Motivation
+
+Being able to see what the user was looking at is a good way to found out problems in the application. 
+This RFC proposes a way to send information from the UI state of an application, known as "view hierarch".
+
+# Proposal
```
marandaneto, line 14: The proposal should include an example of the envelope as well, with a `type` suggestion and links to the https://develop.sentry.dev/sdk/envelopes/ spec

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+    "z": 2, //if 3D.
+    "visible": true|false,
+    "alpha": 1,
+    "{extra_properties}": "{property value}" //adicional information by platform
+}
+```
+
+# Options Considered
+
+We could send the same information as attachment, but this impact on attachment quota.
```
marandaneto, line 37: Not only quota, also because we'd not know if its a view hierarchy or not, link to the screenshots spec that rely on the name of the file-attachment.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+    "x": 0,
+    "y": 1,
+    "z": 2, //if 3D.
+    "visible": true|false,
+    "alpha": 1,
+    "{extra_properties}": "{property value}" //adicional information by platform
+}
+```
+
+# Options Considered
```
marandaneto, line 35: Sending as part of the event should be here too, but because of the size limit, isn't recommended.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+
+We could send the same information as attachment, but this impact on attachment quota.
+
+# Drawbacks
```
marandaneto, line 39: I'd write that we've decided to not add the content of textfields or editables because of PII

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,41 @@
...
+Allow SDKs to send View hierarch of the application UI.
+
+# Motivation
+
+Being able to see what the user was looking at is a good way to found out problems in the application. 
+This RFC proposes a way to send information from the UI state of an application, known as "view hierarch".
+
+# Proposal 
+
+The proposal is to create a new envelope item in the form of a json representation of the UI elements, like so:
```
marandaneto, line 16: I'd write down the average of such view hierarchy size for a simple screen just for having an idea, Ingest team would really appreciate this info.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,46 @@
...
+    "z": 2, //if 3D.
+    "visible": true|false,
+    "alpha": 1,
+    "{extra_properties}": "{property value}" //adicional information by platform
+}
+```
+
+# Options Considered
+
+- We could send the same information as attachment, but this impact on attachment quota. And also we need to rely on file name to determined whether an attachment is a view hierarch or not, this is a problem we already have with [screenshots](https://develop.sentry.dev/sdk/features/#screenshots).
```
AbhiPrasad, line 39: IMO it’s completely fair to send these as attachements. It costs us money to store and use this information.

We can make this a new attachment type so we don’t have the same problem with filename (we can probably do the same with screenshots as well).

text/0033-view-hierarch.md
```diff
@@ -1,12 +1,12 @@
...
 * RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
-* RFC Status: draft
-* RFC Driver: -
+* RFC Status: -
+* RFC Driver: [Dhiogo Ramos Brustolin](https://github.com/brustolin)
 
 # Summary
 
-Allow SDKs to send View hierarch of the application UI.
+Allow SDKs to send a textual representation of the application's UI view hierarch.
```
marandaneto, line 9:
```suggestion
Allow SDKs to send a textual representation of the application's UI view hierarchy. 
```

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,56 @@
...
+Being able to see what the user was looking at is a good way to found out problems in the application. 
+This RFC proposes a way to send information from the UI state of an application, known as "view hierarch".
+
+# Proposal 
+
+The proposal is to create a new [envelope item](https://develop.sentry.dev/sdk/envelopes/) in the form of a json representation of the UI elements. 
+This is how the envelope item header should look like:
+
+```json
+{"type":"view_hierarch","length":...,"content_type":"application/json"}\n
```
marandaneto, line 22:
```suggestion
{"type":"view_ hierarchy","length":...,"content_type":"application/json"}\n
```

brustolin: Are you sure about this space?

marandaneto: nope, GH suggestion messed up somehow, its about the `y` at the end, more places are missing.

text/0033-view-hierarch.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarch for each available screen/window. 
+This is how it should look like: 
+
+```json
+[
+{
```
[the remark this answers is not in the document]
brustolin: Depending on the environment we may have the window position in coordinates, but left/right I would say its not relevant.

text/0033-view-hierarch.md
brustolin: I dont have data to back this up. View hierarchy can vary a lot.

README.md
```diff
@@ -19,4 +19,5 @@ This repository contains RFCs and DACIs.  Lost?
 * [0016-auto-code-mappings](text/0016-auto-code-mappings.md): Automatic code mappings
 * [0022-response-context](text/0022-response-context.md): Response context
 * [0027-manual-disabling-of-flaky-tests](text/0027-manual-disabling-of-flaky-tests.md): processes for manually disabling flaky tests in `sentry` and `getsentry`
-* [0042-gocd-succeeds-freight-as-our-cd-solution](text/0042-gocd-succeeds-freight-as-our-cd-solution.md): plan to replace freight with GoCD
+* [0033-view-hierarch](text/0033-view-hierarch.md): View Hierarch
```
AbhiPrasad, line 22:
```suggestion
* [0033-view-hierarchy](text/0033-view-hierarchy.md): View Hierarchy
```

ueman:
What about code (de)obfuscation?

text/0033-view-hierarch.md
marandaneto: That's why a sample with a simple screen already helps, right now I don't know if it's 1KB/1MB/10MB/100MB.

text/0033-view-hierarch.md
marandaneto: Indeed, good idea @AbhiPrasad 
https://develop.sentry.dev/sdk/envelopes/#attachment
Either a new envelope item type or a new attachment type -> `attachment_type` in the spec.
I'd go this way if we wanna charge for this feature.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,56 @@
...
+        "children": [...], //An array of ui elements.
+        "width": 100, 
+        "height": 100,
+        "depth": 100, //if 3D
+        "x": 0,
+        "y": 1,
+        "z": 2, //if 3D.
+        "visible": true|false,
+        "alpha": 1,
+        "{extra_properties}": "{property value}" //adicional information by platform
```
marandaneto, line 42: What if we make a databag here, `Map<String, Any>`?

text/0033-view-hierarch.md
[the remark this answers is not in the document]
marandaneto: That's just an example, if the x, y, z solves the issue, all good.
We might need a state field tho, for example, on Android, its possible to be 
`half_open`.

marandaneto:
> What about code (de)obfuscation?

Good point, on Android at least, native widgets such as EditText, etc are not obfuscated, Activities also not, but custom widgets and Fragments are, those will be obfuscated, @brustolin is worth mentioning this on the RFC.
Its a good point to start thinking about obfuscation in all the features, breadcrumbs etc are also obfuscated.

text/0033-view-hierarch.md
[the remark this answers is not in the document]
brustolin: Im proposing the minimum required properties. Each platform may add more if needed.

text/0033-view-hierarchy.md
brustolin: It's possible. I'm not a fan. 
Property names can't duplicate, I don't see a reason we need an extra layer of properties.

text/0033-view-hierarch.md
AbhiPrasad: Thought about this more, and I'd vote for an attachment with a different `attachment_type` for the following reasons:
- We'd can immediately start sending and using this data, no need for large Relay updates
- Less new code in the SDK vs. adding a new envelope item, so less hassle to maintain/test
- We don't have to touch outcomes/rate limiting/processing code at all
- We always can eventually migrate to a new envelope item type if we feel the need to, that option is still left open.

marandaneto: The only pending point is what older versions of Relay do with unknown `attachment_type`, is going to be a back compatibility issue and we are fine with it? I guess the same would happen with a new envelope item anyway.

text/0033-view-hierarchy.md
marandaneto: No strong opinions, at least on Relay, we often have the pattern of this `data` databad that holds extra values.

narsaynorath: I'd prefer being able to separate out dynamically added properties (i.e. between platforms) from common ones. I'm still trying to understand how this data will get surfaced in the product after ingestion, but knowing that certain keys are always there vs not will be helpful when working with the data type in the front end

AbhiPrasad: I think we should always all for forwards compatibility for event schemas, so we allow for arbitrary keys, but we don't do anything with them.

text/0033-view-hierarch.md
AbhiPrasad: [As per this logic](https://github.com/getsentry/relay/blob/f0d3edc54b63f4b75f6b8c50a488369d8b2275b0/relay-server/src/envelope.rs#L608-L618), if a relay encounters an unknown attachment type it sets it to the default `event.attachment`: https://github.com/getsentry/relay/blob/f0d3edc54b63f4b75f6b8c50a488369d8b2275b0/relay-server/src/envelope.rs#L372-L376

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,57 @@
...
+This is how the envelope item header should look like:
+
+```json
+{"type":"view_hierarchy","length":...,"content_type":"application/json"}\n
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarchy for each available screen/window. 
+This is how it should look like: 
+
+```json
```
ueman, line 28: Maybe use shorter keys to save some precious file size. Some UI hierarchies can become quite huge.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,57 @@
...
+# Proposal 
+
+The proposal is to create a new [envelope item](https://develop.sentry.dev/sdk/envelopes/) in the form of a json representation of the UI elements. 
+This is how the envelope item header should look like:
+
+```json
+{"type":"view_hierarchy","length":...,"content_type":"application/json"}\n
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarchy for each available screen/window.
```
ueman, line 25: How does it differentiate between multiple windows and multiple root views?

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarchy for each available screen/window. 
+This is how it should look like: 
+
+```json
+[
+    {
+        "type": "element_class_name", 
+        "identifier": "element_identifier",
```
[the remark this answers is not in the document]
brustolin: Yes

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,57 @@
...
+        "children": [...], //An array of ui elements.
+        "width": 100, 
+        "height": 100,
+        "depth": 100, //if 3D
+        "x": 0,
+        "y": 1,
+        "z": 2, //if 3D.
+        "visible": true|false,
```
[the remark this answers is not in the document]
brustolin: This `visible` property is for `isHidden` in iOS, or `visibility` GONE or INVISIBLE for Android. 
Out of bound views may be determined by the position of the view relative to the size of the parent.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,57 @@
...
+        "alpha": 1,
+        "{extra_properties}": "{property value}" //adicional information by platform
```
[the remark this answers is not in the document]
brustolin: Yes. Each platform can determine the required properties. The ones in this RFC are a small sample set.

text/0033-view-hierarchy.md
brustolin: Each Window is a UI element in the array, and since properties are dynamic, the platform may choose different properties for Windows.

Mixed UI will have different properties, but still UI Elements in a hierarchy.

text/0033-view-hierarch.md
brustolin: I really like the idea, but im wondering if its possible to de-obsfucate information in the attachment.

philipphofmann (commented):
Our Replay product feature uses the DOM. I think it's worth checking which representation they use.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,59 @@
...
+
+```json
+{"type":"view_hierarchy","length":...,"content_type":"application/json"}\n
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarchy for each available screen/window. 
+This is how it should look like: 
+
+```json
+[
```
jjbayer, line 29: Since this is a list, should each list element have information about what screen/window it refers to? Or is that covered by the `identifier`?

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Options Considered
+
+- We could send the same information as attachment, but this impact on attachment quota. And also we need to rely on file name to determined whether an attachment is a view hierarchy or not, this is a problem we already have with [screenshots](https://develop.sentry.dev/sdk/features/#screenshots).
+- The view hierarchy could be part of the Event, but because of the size limit, isn't recommended. 
+
+# Drawbacks
+
+- This can significantly increase the envelope size. 
+- Obfuscation is a problem for custom UI controls, we'll probably need ingestion to de-obfuscate in the server side.
```
jjbayer, line 57: Could you elaborate on this? what kind of obfuscation are we talking about?

text/0033-view-hierarchy.md
brustolin: This is an array of screen/windows. Every other ui element of each screen will be listed inside the `children` property.

text/0033-view-hierarchy.md
brustolin: Java for example. A custom view may have its class name minified by pro-guard.

jjbayer: Right. In that case we might need to do a similar thing as we do for profiles: https://github.com/getsentry/sentry/blob/cf71af372677487d7d0a7fd8ac9dd092f9596cf4/src/sentry/profiles/task.py#L350-L360.

I guess the entries of the view hierarchy would then have to contain the fully qualified class name of the view (either as `identifier` or as additional field, see https://github.com/getsentry/rust-proguard/blob/4bc2cddd79cc8f7c9d14dda2f9f4c431e10c9424/src/mapper.rs#L169-L172.

brustolin: They will have, in the `type` property.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,59 @@
...
+{"type":"view_hierarchy","length":...,"content_type":"application/json"}\n
+```
+
+And the envelope item payload will be a JSON array that contains a view hierarchy for each available screen/window. 
+This is how it should look like: 
+
+```json
+[
+    {
+        "type": "element_class_name",
```
marandaneto, line 31:
```suggestion
        "type": "element_full_qualified_class_name", 
```

text/0033-view-hierarch.md
AbhiPrasad: We can apply processing on an attachment - this is what we do with minidumps for example.

narsaynorath: I've also been trying to figure out the work to use this new envelope type, so I met yesterday with @JoshFerge for insight on how they handled a new envelope type for replay, and he echoes @AbhiPrasad's comments r.e a substantial amount of work required in Relay to support this (adding the new type to relay and adding a consumer being the biggest portions)

If we have these sending as attachments, our team could likely do something quick to render the attachment preview (otherwise the user just gets their JSON which isn't too helpful) before we get into the main project development

JoshFerge:
super neat stuff! just wanted to call out @indragiek's and I's hackweek project code may be interesting to ya'll, as there is some iOS view hierarchy listening code in there. (Goal was to render the view in HTML form in order to eventually have mobile replays) https://github.com/getsentry/hackweek-ios-replays/tree/main/ios

text/0033-view-hierarchy.md
```diff
@@ -10,12 +10,69 @@ Allow SDKs to send a textual representation of the application's UI view hierarc
...
-The proposal is to create a new [envelope item](https://develop.sentry.dev/sdk/envelopes/) in the form of a json representation of the UI elements. 
+# Proposal
+
+Capture the view hierarchy in our mobile SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.viewhierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
+
+The attachment payload is a JSON structure containing the view hierarchy, here's an example:
+```json
+{
+    "rendering_system": "compose", // or android_view_system, apple_uikit, apple_swiftui, unity, flutter, ...
+    "screens": [
```
brustolin, line 29: @markushi I think windows is more correct, screens is more related to displays imo.

```suggestion
    "windows": [
```

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,116 @@
...
+
+Example View Hierarchies from PoC implementations:
+* [Android](https://github.com/getsentry/sentry-java/pull/1998#issuecomment-1104806065)
+* [iOS](https://github.com/getsentry/sentry-cocoa/pull/2044#issue-1329719543)
+* [Flutter](https://github.com/ueman/sentry-dart-tools/blob/b850db587f0e099bed253d13055c88c03d536875/sentry_flutter_plus/lib/src/integrations/tree_walker_integration.dart)
+* [Unity](https://github.com/getsentry/team-mobile/issues/64#issuecomment-1290868653)
+
+# Proposal
+
+Capture the view hierarchy in our mobile SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.viewhierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
```
jjbayer, line 23: We need to add this type to the corresponding Relay type, otherwise the envelope will fail to parse afaict: https://github.com/getsentry/relay/blob/0586fa62eb3ee75697120ea6d89efe51c740f6ad/relay-server/src/envelope.rs#L362-L369.

brustolin, line 23:
```suggestion
Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.viewhierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
```

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Options Considered
+
+- We could send the same information as attachment, but this impact on attachment quota. And also we need to rely on file name to determined whether an attachment is a view hierarchy or not, this is a problem we already have with [screenshots](https://develop.sentry.dev/sdk/features/#screenshots).
+- The view hierarchy could be part of the Event, but because of the size limit, isn't recommended. 
+
+# Drawbacks
+
+- This can significantly increase the envelope size. 
+- Obfuscation is a problem for custom UI controls, we'll probably need ingestion to de-obfuscate in the server side.
```
marandaneto, line 114: Link this to https://docs.sentry.io/product/data-management-settings/scrubbing/attachment-scrubbing/

JoshFerge:
> Our Replay product feature uses the DOM. I think it's worth checking which representation they use.
> 
> Do we know what we want to do with the view hierarchy data in Sentry? The data format depends on the answer to the previous question. Maybe we can use an existing solution, which requires a specific data format.

We use rrweb's json format, and their player library. for our hackweek we attempted to translate the iOS view hierarchy into rrweb JSON. you could try something similar for this -- I'm sure instead of a player you could just have it render HTML as well.

text/0033-view-hierarchy.md
brustolin: We just need to include the new type in the enum? Or there is more changes do to?

text/0033-view-hierarchy.md
brustolin: Maybe `identifier` to `id`. but this properties will appear as information to the user. So we need to be clear.

brustolin:
@JoshFerge @philipphofmann The plan now is to have a textual tree representation of the View hierarchy. I assume that the RRWeb json format is more complex than what we are proposing and therefore would delay the release of this feature.

brustolin:
@jjbayer do you believe we have enough information regarding what we need from ingestion? 
Is it possible to approve this proposal?

text/0033-view-hierarchy.md
jjbayer: Yes that should be enough.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+# Drawbacks
+- By utilizing attachments, [attachment scrubbing](https://docs.sentry.io/product/data-management-settings/scrubbing/attachment-scrubbing/) is relevant for any PII attributes in the view hierarchy
+- Increases event payload size
+- We've decided to not add the content of textfields, editables or labels because of PII.
+    - We may evaluate an options to allow PII.
+
+
+## Initial Proposal
```
jjbayer, line 75: Will the initial proposal remain a part of the final doc? It's a bit confusing to read.

text/0033-view-hierarchy.md
jjbayer: @jan-auer since we're not forward compatible in regards to attachment types, meaning that outdated external Relays will [fail to parse the envelope](https://github.com/getsentry/relay/blob/a9297f9cecb00b2caa09033a3c12438cc8c23b76/relay-server/src/envelope.rs#L1041) when there's a new attachment type, should we just use `AttachmentType::Attachment` here instead?

I noticed that for screenshots, there is [no special attachment type](https://github.com/getsentry/sentry/blob/1c9c4829a3d9f8957eb050b78539432792ae861f/src/sentry/models/eventattachment.py#L23-L30) either.

brustolin:
>I noticed that for screenshots, there is [no special attachment type](https://github.com/getsentry/sentry/blob/1c9c4829a3d9f8957eb050b78539432792ae861f/src/sentry/models/eventattachment.py#L23-L30) either.

And this is a problem. Front-end needs to check for attachment file name which IMO is not good.

text/0033-view-hierarchy.md
brustolin: I think we can remove it.

text/0033-view-hierarchy.md
brustolin: Instead of attachment type we could maybe use a different mime type.

jjbayer:
> > I noticed that for screenshots, there is [no special attachment type](https://github.com/getsentry/sentry/blob/1c9c4829a3d9f8957eb050b78539432792ae861f/src/sentry/models/eventattachment.py#L23-L30) either.
> 
> And this is a problem. Front-end needs to check for attachment file name which IMO is not good.

Discussed this with @jan-auer and agreed that we should introduce a new attachment type. We will add forward compatibility to the attachment type in Relay. Because external Relays could still be outdated when we launch this feature, we should warn users who enable view hierarchies that they _have_ to update their Relays first (if they have any).

brustolin:
>Discussed this with @jan-auer and agreed that we should introduce a new attachment type. We will add forward compatibility to the attachment type in Relay. Because external Relays could still be outdated when we launch this feature, we should warn users who enable view hierarchies that they have to update their Relays first (if they have any).

Make sense. Thanks!!

jjbayer (approved):

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,78 @@
...
+
+Example View Hierarchies from PoC implementations:
+* [Android](https://github.com/getsentry/sentry-java/pull/1998#issuecomment-1104806065)
+* [iOS](https://github.com/getsentry/sentry-cocoa/pull/2044#issue-1329719543)
+* [Flutter](https://github.com/ueman/sentry-dart-tools/blob/b850db587f0e099bed253d13055c88c03d536875/sentry_flutter_plus/lib/src/integrations/tree_walker_integration.dart)
+* [Unity](https://github.com/getsentry/team-mobile/issues/64#issuecomment-1290868653)
+
+# Proposal
+
+Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.viewhierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
```
AbhiPrasad, line 23: Do we want to use `event.view_hierarchy` instead of `event.viewhierarchy`? I'll let you make the final call on that.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,78 @@
...
+            "identifier": "R.id.button_logout",
+            "children": [], // nested elements
+            "width": 100.0,
+            "height": 100.0,
+            "depth": 100.0, // if applies
+            "x": 0.0,
+            "y": 0.0,
+            "z": 2.0, // if applies.
+            "visible": true|false,
+            "alpha": 1, // a float number from 0 to 1, where 0 means totally transparent, and 1 totally opaque.
```
AbhiPrasad, line 41:
```suggestion
            "alpha": 1, // a float number from 0 to 1, where 0 means transparent, and 1 is opaque.
```

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,78 @@
...
+    ]
+}
+```
+
+Some remarks on the example:
+ * `windows`: contains all visible windows, on mobile it's typically just one or two (e.g. if a dialog is open)
+ * `type`: The fully qualified widget class name, this name may be obfuscated on certain platforms (e.g. Android release builds with proguard enabled)
+ * `children` nests all child UI widgets, which then builds up the whole UI tree
+
+A typical Android/iOS view hierarchy for a single window consists of around 100-200 objects. Taking the attributes from the example above this generates a raw JSON file with a size of around 50KB. More complex view hierarchies for e.g. a Unity game may hold 1000-2000 objects, producing a JSON file of around 500KB.
```
AbhiPrasad, line 53:
```suggestion
A typical Android/iOS view hierarchy for a single window consists of around 100-200 objects. Taking the attributes from the example above this generates a raw JSON file with a size of around 50KB. More complex view hierarchies, for example a Unity game, may hold 1000-2000 objects, producing a JSON file of around 500KB.
```

brustolin:
Thanks @AbhiPrasad!

text/0033-view-hierarchy.md
brustolin:
```suggestion
Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.view_hierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
```

philipphofmann (approved):
Great job 👏 . Let's go ahead and iterate over it if it doesn't work.

text/0033-view-hierarchy.md
```diff
@@ -0,0 +1,79 @@
...
+    - We can immediately start sending and using this data, no need for large Relay updates
+    - Less new code in the SDKs when compared to adding a new envelope item, so less hassle to maintain/test
+    - We don't have to touch outcomes/rate limiting/processing code at all
+    - We always can eventually migrate to a new envelope item type if we feel the need to, that option is still left open.
+- The view hierarchy could be part of the Event, but because of the size limit, it is not recommended
+
+
+# Drawbacks
+- By utilizing attachments, [attachment scrubbing](https://docs.sentry.io/product/data-management-settings/scrubbing/attachment-scrubbing/) is relevant for any PII attributes in the view hierarchy
+- Increases event payload size
```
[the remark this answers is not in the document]
brustolin: Thanks. Proposal changed, didn't change the drawbacks.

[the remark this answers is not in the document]
brustolin:
```suggestion
```

text/0033-view-hierarchy.md
jjbayer: I agree that `event.view_hierarchy` is more readable, but we have a precedent with `event.applecrashreport`, so I would stick to that convention: https://github.com/getsentry/relay/blob/36ca3675eae2202478cc7e001c2d8c9f189581c8/relay-server/src/envelope.rs#L334-L336

brustolin: I see. Thanks for the reference

marandaneto (approved):

narsaynorath (approved):
Looks great, thanks for the detailed example for the JSON structure and the expected sizes of view hierarchies. That information became handy in a product meeting with MDX :pray:

PR #34, opened 2022-11-03
rfc(feature): SDK Lifecycle Hooks

AbhiPrasad:
[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc-sdk-lifecycle-hooks/text/0034-sdk-lifecycle-hooks.md)

This PR proposes the introduction of lifecycle hooks in the SDK, improving extensibility and usability of SDKs at Sentry.

To test out this RFC, we implemented it in the [JavaScript SDK](https://github.com/getsentry/sentry-javascript/blob/7aa20d04a3d61f30600ed6367ca7151d183a8fc9/packages/types/src/client.ts#L153) with great success, so now we are looking to propose and implement it in the other SDKs.

Lifecyle hooks can be registered on a Sentry client, and allow for integrations/sdk users to have finely grained control over the event lifecycle in the SDK.

```ts
interface Client {
  // ...

  on(hookName: string, callback: (...args: unknown) => void) => void;

  emit(hookName: string, ...args: unknown[]) => void;

  // ...
}
```

JS SDK work: https://github.com/getsentry/sentry-javascript/issues/7262


text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+
+This PR proposes the introduction of lifecycle hooks in the SDK, improving extensibility and usability of SDKs at Sentry.
+
+Lifecyle hooks can be registered as a top level method, and allow for integrations/sdk users to have finely grained control over the event lifecycle in the SDK.
+
+```ts
+Sentry.on("hook-name", (...args) => {
+  someLogic();
+});
+```
```
antonpirker, line 17: Maybe also something like this to have everything in the `init()` function:

```python
sentry_sdk.init(
    dsn="...",
    hooks=[
        ("hook-name", some_function),
        ("other-hook", other_function),
    ]
)
```

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+
+```
+| Session | ---> | Envelope | ---> | Transport | ---> | Sentry |
+```
+
+The SDKs provide a few ways to extend this pipeline:
+
+1. Event Processors (what Integrations use)
+2. `beforeSend` callback
+3. `beforeBreadcrumb` callback
```
antonpirker, line 41:
```suggestion
3. `beforeBreadcrumb` callback
4. `beforeSendTransaction` callback (coming soon)
```

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+
+But these are all top level options in someway, and are part of the unified API as a result. This means that in certain scenarios, they are not granular enough as extension points.
+
+# Proposal
+
+SDK hooks live on the client, and are **stateless**. They are called in the order they are registered. SDKs can opt-in to whatever hooks they use, and there can be hooks unique to an SDK.
+
+```ts
+class Client {
+  hooks: {
```
antonpirker, line 51: If order is important I would use an array/list instead of a dict (in Python the order of dicts is not defined)

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+
+  on(hookName: string, callback: HookCallback) {
+    this.hooks[hookName].push(callback);
+  }
+}
+```
+
+## Hooks
+
+Hooks can return `null` to short-circuit the pipeline.
```
antonpirker, line 63: A more detailed description of what "short-circuit" means should be provided, so that all SDKs handle hooks the same way.

yordis: Instead of doing this `null` trick, I think a more functional approach is to give the next callback in the pipeline.

Like Redux middleware, or for example: https://github.com/straw-hat-team/nodejs-monorepo/blob/master/packages/%40straw-hat/fetcher/docs/middlewares.md

I think it will be valuable to see if that approach will simplify the architecture of the internal code and also simplify the usage by making explicitly the chaining calls.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+Sentry.on("hook-name", (...args) => {
+  someLogic();
+});
+```
+
+# Motivation
+
+There are three main ways users can extend functionality in the SDKs right now.
+
+At it's current form, the SDK is an event processing pipeline. It takes in some data (an error/message, a span, a profile), turns it into the event, attaches useful context to that event based on the current scope, and then sends that event to Sentry.
```
yordis, line 23: I remember talking about this, and you created a lovely table and categorization of the extensions, I can't find it, but I will use to share it here as well.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,106 @@
...
+
+- `captureEvent`
+
+```ts
+type onCaptureEvent = (
+  event: Event,
+  hint?: EventHint,
+  scope?: Scope
+) => string | null;
+```
```
yordis, line 94:
```suggestion
```ts
type onCaptureException<T = any> = (args: {
  exception: T,
  hint?: EventHint,
  scope?: Scope
}) => T | null;
```

- `captureMessage`

```ts
type onCaptureMessage = (args: {
  message: string,
  level?: Severity,
  hint?: EventHint,
  scope?: Scope
}) => string | null;
```

- `captureEvent`

```ts
type onCaptureEvent = (args: { 
  event: Event,
  hint?: EventHint,
  scope?: Scope
}) => string | null;
```


If you want to avoid breaking changes, I suggest turning the position parameters into a single argument message.

You can rename, keep adding more to an object and avoid confusing positioning, especially with multiple optional parameters.

## Caveats

Make sure to come up with a good name for that arg, and don't assume that people will destructure as a habit, so better off giving them a good answer for a variable name.

## Drawbacks

??

text/0034-sdk-lifecycle-hooks.md
yordis: Is it related to JS only to the RFC, or overall for all clients?

The reason I say that is because it could be a `Map` that keeps the order of the keys, so maybe the RFC should be apparent in terms of "must be in order" but stay away from the technicality of data structures, `if` this is overall for all clients.

Otherwise, use a `Map` then.

sl0thentr0py (commented):
shippppit
should we think of common hooks across SDKs or is this completely left to SDK authors?

AbhiPrasad:
>  should we think of common hooks across SDKs or is this completely left to SDK authors?

We will define a set of initial ones in the develop docs, but the rest can be left to SDK authors to solve their platform specific problems.

Since for now hooks are meant to be used internally I don't think we worry about SDK user concerns yet, but maybe we can re-evaluate this in 2-3 months after more SDKs adopt it.

cleptric (approved):
This is a great example of finding a very concise and simple solution to a problem that might be super platform-specific, so being able to extend this on an SDK by SDK basis is 👌 

:shipit:

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+Hooks are meant to be mostly internal APIs for integration authors, but we can also expose them to SDK users if there is a use case for it.
+
+As hook callbacks are not processed by the client, they can be async functions.
+
+```ts
+// Example implementation in JavaScript
+
+type HookCallback = (...args: unknown[]): void;
+
+class Client {
```
yordis, line 74: I am a bit confused, who implements this `Client`? What exactly the end-user interacts with?

cleptric: Thanks a lot for showing an interset in our RFC process! 🚀 
This is meant to be the client implementation we have in each SDK, see https://develop.sentry.dev/sdk/unified-api/#client

yordis: I need to think a layer higher and across multiple languages and clients. Sorry for the confusion, and thank you so much for your response!

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+```ts
+on('finishSpan', callback: (span: Span) => void) => void;
+```
+
+`beforeEnvelope`:
+
+```ts
+on('beforeEnvelope', callback: (envelope: Envelope) => void) => void;
+```
```
marandaneto, line 125: Why do we need a hook for `beforeEnvelope`? Can we make it internal at least? AFAIK this is not exposed for clients and the whole envelope impl. is marked as internal in Mobile SDKs at least.

text/0034-sdk-lifecycle-hooks.md
marandaneto: Was gonna ask about `beforeSendTransaction` as well, some SDKs already added it.
The suggestion is resolved but not committed, mind explaining?

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+  emit(hookName: string, ...args: unknown[]): void {
+    this.hooks[hookName].forEach(callback => callback(...args));
+  }
+}
+```
+
+SDKs are expected to have a common set of hooks, but can also have SDK specific hooks.
+
+## Hooks
```
marandaneto, line 91: What's about hooks for `startSession` and `endSession`?

Lms24 (approved):
Great RFC and big +1 on the  "implementing it as needed per SDK" approach!

mitsuhiko:
I would like to explicitly document the requirements for callbacks with regards to forward compatibility. In particular in Python for instance we should expect them to always catch extra keyword arguments so that additional data can be passed in the future.

text/0034-sdk-lifecycle-hooks.md
philipphofmann: @cleptric, I also need some clarification. Does this mean I need to implement a client when I want to add lifecycle hooks? Can you share the code a user must write when implementing a lifecycle hook?

AbhiPrasad: The `on` and `emit` hooks will be defined on the client interface (much like `flush` and `capture_event` are at the moment).

Usually users (if writing custom clients) are inheriting from a base client implementation anyway, so they should get the hooks functionality out of the box.

text/0034-sdk-lifecycle-hooks.md
AbhiPrasad: Yes we can make some hooks internal. I will clarify this in the RFC.

text/0034-sdk-lifecycle-hooks.md
AbhiPrasad: This was not an exhaustive list, but I will add those.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+- Integrations want to mutate an error on `captureException`
+- Integrations want to override propagation behaviour (extracing/injecting outgoing headers)
+
+# Proposal
+
+SDK hooks live on the client, and are **stateless**. They are called in the order they are registered. SDKs can opt-in to whatever hooks they use, and there can be hooks unique to an SDK.
+
+Hooks are meant to be mostly internal APIs for integration authors, but we can also expose them to SDK users if there is a use case for it.
+
+As hook callbacks are not processed by the client, they can be async functions.
```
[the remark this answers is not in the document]
AbhiPrasad: Good point, I will add notes around this.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+SDKs are expected to have a common set of hooks, but can also have SDK specific hooks.
+
+## Hooks
+
+These following are a set of example hooks that would unblock some use cases listed above. The names/schema of the hooks are not final, and are meant to be used as a starting point for discussion.
+
+To document and approve new hooks, we will create a new page in the develop docs that lists all the hooks, and what they are used for.
+
+`startTransaction`:
```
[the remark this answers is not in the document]
AbhiPrasad: Yeah that makes sense to me - can add some notes

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+class Client {
+  hooks: {
+    [hookName: string]: HookCallback[];
+  };
+
+  on(hookName: string, callback: HookCallback): void {
+    this.hooks[hookName].push(callback);
+  }
+
+  emit(hookName: string, ...args: unknown[]): void {
```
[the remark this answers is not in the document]
AbhiPrasad: `on` is called to register a hook - the hook names are dynamic.

`emit` is called to emit values to all registered hooks with that name.

If I first do:

```
client.on('startTransaction', callback);
```

then that callback function pointer will be called every time I call `client.emit('startTransaction')`, along side any other functions that were registered under the `startTransaction` name.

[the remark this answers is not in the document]
philipphofmann: Ah, okay, now I get it as you have to implement your custom client, it makes sense.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Hooks
+
+These following are a set of example hooks that would unblock some use cases listed above. The names/schema of the hooks are not final, and are meant to be used as a starting point for discussion.
+
+To document and approve new hooks, we will create a new page in the develop docs that lists all the hooks, and what they are used for.
+
+`startTransaction`:
+
+```ts
```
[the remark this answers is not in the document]
yordis: Why would that be? They are two different functions as far as I can tell 🤔

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+- RFC Status: approved
+- RFC Driver: [Abhijeet Prasad](https://github.com/AbhiPrasad)
+
+# Summary
+
+This PR proposes the introduction of lifecycle hooks in the SDK, improving extensibility and usability of SDKs at Sentry.
+
+To test out this RFC, we implemented it in the [JavaScript SDK](https://github.com/getsentry/sentry-javascript/blob/7aa20d04a3d61f30600ed6367ca7151d183a8fc9/packages/types/src/client.ts#L153) with great success, so now we are looking to propose and implement it in the other SDKs.
+
+Lifecyle hooks can be registered on a Sentry client, and allow for integrations/sdk users to have finely grained control over the event lifecycle in the SDK.
```
philipphofmann, line 13: > integrations/sdk users

That confused me a bit. I thought this API also targets our users, but as you @AbhiPrasad, explained in the TSC, this API is mostly internal. This information is essential to me as it completely changes how I evaluate this RFC.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+- Integrations want to add information to spans when they start or finish. This is what [RFC #75 is running into](https://github.com/getsentry/rfcs/pull/75), where they want to add thread information to each span.
+- Integrations want to add information to envelopes before they are sent to Sentry.
+- Integrations want to run code on transaction/span finish (to add additional spans to the transaction, for example).
+- Integrations want to mutate an error on `captureException`
+- Integrations want to override propagation behaviour (extracing/injecting outgoing headers)
+
+# Proposal
+
+SDK hooks live on the client, and are **stateless**. They are called in the order they are registered. SDKs can opt-in to whatever hooks they use, and there can be hooks unique to an SDK.
```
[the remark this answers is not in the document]
philipphofmann: It would be great to put this info in the RFC or the develop docs, as this was unclear to me when reading the RFC.

I didn't write JavaScript for 6 years or so, but event emitter still rings a bell now. For statically typed languages, we might have to adapt them to the language. For Swift and Java, it would maybe be something like `addObserver` and `notifyObserver`, or maybe we just stick to the JavaScript syntax. Thanks for explaining.

text/0034-sdk-lifecycle-hooks.md
```diff
@@ -0,0 +1,125 @@
...
+
+But these are all top level options in someway, and are part of the unified API as a result. This means that in certain scenarios, they are not granular enough as extension points.
+
+The following are some examples of how sdk hooks can unblock new features and integrations, but not a definitive list:
+
+- Integrations want to add information to spans when they start or finish. This is what [RFC #75 is running into](https://github.com/getsentry/rfcs/pull/75), where they want to add thread information to each span.
+- Integrations want to add information to envelopes before they are sent to Sentry.
+- Integrations want to run code on transaction/span finish (to add additional spans to the transaction, for example).
+- Integrations want to mutate an error on `captureException`
+- Integrations want to override propagation behaviour (extracing/injecting outgoing headers)
```
philipphofmann, line 59: On Mobile, we want to take a screenshot or attach the view hierarchy when capturing an exception. That's a good use case. Could we replace our custom logic of [attachment processors](https://github.com/getsentry/sentry-cocoa/blob/9d5623218051d36e447a50515fad0f0252e48d65/Sources/Sentry/SentryClient.m#L389-L396) with that, @AbhiPrasad? We call some code that returns attachments and then bundle it up into an envelope when capturing exceptions.

philipphofmann (commented):
If the answer to my question https://github.com/getsentry/rfcs/pull/34#discussion_r1146777066 is yes, then LGTM. Great work @AbhiPrasad 👏

text/0034-sdk-lifecycle-hooks.md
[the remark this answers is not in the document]
philipphofmann: Sorry, I got it wrong, @yordis. I'm going to resolve this comment.

marandaneto (approved):

text/0034-sdk-lifecycle-hooks.md
AbhiPrasad: This was a remnant from before, forgot to change this!

text/0034-sdk-lifecycle-hooks.md
AbhiPrasad: Yes! We could for sure replace that with this, will add your example to the list!

PR #36, opened 2022-11-09
Auto-Instrumentation UI thread

romtsn:
Add new `blocked_main_thread` attribute to the `data` key-value map of the [`Span`](https://develop.sentry.dev/sdk/event-payloads/span/) interface, indicating whether an auto-instrumented span has run its entire duration on the UI/Main thread.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/feat/file-io-main-thread/text/0036-auto-instrumentation-ui-thread.md)

### DIFF
--- a/text/0036-auto-instrumentation-ui-thread.md
+++ b/text/0036-auto-instrumentation-ui-thread.md
@@ -84,11 +84,13 @@ An example of a File I/O span payload with the newly added attributes:
 }
 ```
 
-## (De)obfuscation
+## Symbolication
 
-As the `call_stack` may contain obfuscated frames we need to process them server-side, similar to what we already do for 
+As the `call_stack` may contain obfuscated frames we need to symbolicate them server-side, similar to what we already do for 
 [profiles](https://github.com/getsentry/sentry/blob/cf71af372677487d7d0a7fd8ac9dd092f9596cf4/src/sentry/profiles/task.py#L350-L360). 
-In addition, after de-obfuscation, the frames should be filtered to only `in-app` frames, using the `app.app_identifier` context.
+
+To be able to symbolicate the `call_stack` which is part of the `Span` interface server-side, we have to start sending the `debug_meta`
+information for transactions as well.
 
 # Options Considered
 
@@ -120,7 +122,3 @@ should not be a problem. Other potential detectors we could create for detecting
 * Network I/O on main thread
 * Database operations on main thread
 * Resources (images/audio/video) manipulation on main thread (e.g. decoding, resizing, etc.)
-
-# Unresolved questions
-
-* Should we ignore native framework I/O operations, or should we add a new flag in the data bag? iOS screen parser always runs in the UI Thread and the developer has no control over this. Raising issues for this scenario doesn't seems right.
--- a/text/0036-auto-instrumentation-ui-thread.md
+++ b/text/0036-auto-instrumentation-ui-thread.md
@@ -1,6 +1,6 @@
 * Start Date: 2022-11-08
 * RFC Type: decision
-* RFC PR: <link>
+* RFC PR: [#36](https://github.com/getsentry/rfcs/pull/36)
 * RFC Status: -
 * RFC Driver: [Roman Zavarnitsyn](https://github.com/romtsn)
 
--- a/text/0036-auto-instrumentation-ui-thread.md
+++ b/text/0036-auto-instrumentation-ui-thread.md
@@ -74,4 +74,5 @@ should not be a problem. Other potential detectors we could create for detecting
 * Should the `ui_thread` attribute be part of the protocol or just a new key in the data bag?
 * How to properly fingerprint the File I/O performance issue? Do we need to provide a `caller_function`, which has 
 performed the File I/O operation?
+* Should we ignore native framework I/O operations, or should we add a new flag in the data bag? iOS screen parser always runs in the UI Thread and the developer has no control over this. Raising issues for this scenario doesn't seems right.
 * (Out of scope) De-obfuscation/symbolication of the `caller_function`.
--- a/text/0036-auto-instrumentation-ui-thread.md
+++ b/text/0036-auto-instrumentation-ui-thread.md
@@ -6,8 +6,7 @@
 
 # Summary
 
-Add new `blocked_ui_thread` attribute to the `data` key-value map of the [`Span`](https://develop.sentry.dev/sdk/event-payloads/span/) interface, indicating whether
-an auto-instrumented span has run its entire duration on the UI/Main thread.
+Add new `blocked_main_thread` attribute to the `data` key-value map of the [`Span`](https://develop.sentry.dev/sdk/event-payloads/span/) interface, indicating whether an auto-instrumented span has run its entire duration on the UI/Main thread.
 
 # Motivation
 
@@ -42,7 +41,7 @@ to the `Span.data` map.
 
 Add two new attributes to the `data` key-value map of the [`Span`](https://develop.sentry.dev/sdk/event-payloads/span/) interface:
 
-  1. `blocked_ui_thread` - indicates whether the Span has spent its entire duration on the Main/UI thread.
+  1. `blocked_main_thread` - indicates whether the Span has spent its entire duration on the Main/UI thread.
   2. `call_stack` - contains the most relevant stack frames, that lead to the File I/O span. The stack frame should adhere to the [`StackFrame`]
   (https://develop.sentry.dev/sdk/event-payloads/stacktrace/#frame-attributes) interface. When possible, should only include `in-app` frames. 
   Used for proper fingerprinting and grouping of the File I/O performance issues.
@@ -60,7 +59,7 @@ An example of a File I/O span payload with the newly added attributes:
   "trace_id": "b2a33f3f79fe4a7c8de3426725a045cb",
   "status": "ok",
   "data": {
-    "blocked_ui_thread": true,
+    "blocked_main_thread": true,
     "call_stack": [
       {
         "function": "onClick",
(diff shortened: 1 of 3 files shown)


text/0036-auto-instrumentation-ui-thread.md
```diff
@@ -0,0 +1,77 @@
...
+cannot be expanded to custom spans. However, given, that our auto-instrumentation offering is growing, this 
+should not be a problem. Other potential detectors we could create for detecting performance issues:
+
+* Network I/O on main thread
+* Database operations on main thread
+* Resources (images/audio/video) manipulation on main thread (e.g. decoding, resizing, etc.)
+
+# Unresolved questions
+
+* Should the `ui_thread` attribute be part of the protocol or just a new key in the data bag?
```
romtsn, line 74: I guess this is the main question here to you folks @AbhiPrasad and @jjbayer (unless we say this option is total garbage).

AbhiPrasad: I really like this direction, and I think we can make it a forwards compatible data bag. This means we treat certain keys as special cased in Relay, but we don't need to restrict the data bag to just those values.

These keys have special _semantic meaning_ - this is how OpenTelemetry operates as well. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/README.md

We can maintain a list of these keys in our develop docs/relay - so we can use them across SDKs.

AbhiPrasad: It’s because this can potentially balloon as we want to do similar analysis on other span attributes.

I think of the data bag on the span similar to contexts on events - we can special case certain values in there to treat them differently during processing/in the product.

marandaneto: The problem is that the `data` bag is open for users as well.
One can easily coincidentally use the very same key and break the feature.
Contexts on the event have the same problem.
A few alternatives.
* Top level - optional field
* A Sentry prefix for special cased keys in the `data` bag (avoiding collision)
* A different `data` bag only for special cased keys

AbhiPrasad: This same thing applies for contexts as well, as well as span op and description (which is also used by detectors).

the likelyhood of this happening in my opinion is very low since

a) we are setting on data not tags (set tag is probably more used for spans since it matches the set tag for events)
b) spans are not indexed, so most people are probably setting attributes on the transaction, not the span.

text/0036-auto-instrumentation-ui-thread.md
```diff
@@ -0,0 +1,77 @@
...
+* How to properly fingerprint the File I/O performance issue? Do we need to provide a `caller_function`, which has
```
brustolin, line 75:
```suggestion
* Should we ignore native framework I/O operations, or should we add a new flag in the data bag? iOS screen parser is always made in the UI Thread and the developer has no control over this. Raising issues for in this scenario doesn't seems right,
* How to properly fingerprint the File I/O performance issue? Do we need to provide a `caller_function`, which has 
```

text/0036-auto-instrumentation-ui-thread.md
```diff
@@ -0,0 +1,77 @@
...
+* Network I/O on main thread
+* Database operations on main thread
+* Resources (images/audio/video) manipulation on main thread (e.g. decoding, resizing, etc.)
+
+# Unresolved questions
+
+* Should the `ui_thread` attribute be part of the protocol or just a new key in the data bag?
+* How to properly fingerprint the File I/O performance issue? Do we need to provide a `caller_function`, which has 
+performed the File I/O operation?
+* (Out of scope) De-obfuscation/symbolication of the `caller_function`.
```
brustolin, line 77:
```suggestion
* Should we ignore native framework I/O operations, or should we add a new flag in the data bag? iOS screen parser always runs the UI Thread and the developer has no control over this. Raising issues for this scenario doesn't seems right.
* (Out of scope) De-obfuscation/symbolication of the `caller_function`.
```

text/0036-auto-instrumentation-ui-thread.md
marandaneto: Less likely than contexts, yes, but this problem exists for contexts, so maybe a `prefix` won't hurt.

AbhiPrasad: > maybe a prefix won't hurt

My hesitation against a prefix here is that we'll end up adding more and more of these, and I'd rather we just start a set of semantic conventions, and then educate our users about this - it's a more scalable approach.

What is the chance that someone is going to grab the active span and set this exact data field?

marandaneto: I can't tell you the chance but this is a known issue for Contexts, customers already came up with "bugs" because they were using "our reserved" words, people don't look up the docs for that, so, it will happen.

romtsn: @marandaneto @AbhiPrasad I'm thinking about changing the attribute name  to something more explicit - for example `blocked_ui_thread`. As @mattjohnsonpint mentioned during the TSC, `ui_thread` might be too misleading to be applied for _any_ auto-instrumentation, because e.g. for activity/viewcontroller transactions (that are also auto-instrumented) we cannot say for sure how much time they spent on the UI vs background thread.

`blocked_ui_thread` would be more explicit in telling that the span was actually blocking the UI thread its entire duration.

What do y'all think?

marandaneto: A TR/Span can technically start in the UI thread and finish in the background, and vice-versa, are we considering this use case?

romtsn: not with this RFC, this is solely focused on auto-instrumented spans, that cannot span across multiple threads, like File I/O

marandaneto: I see, the problem is that the naming is very generic, but it has a specific scope (such as auto instrumentation for File IO), so I'd either put a more specific naming or the RFC should also take this under consideration, otherwise this will be misused or be open to misinterpretation.

romtsn: I agree, that's why I think `blocked_ui_thread` is more specific in telling that the span has blocked the UI thread for its entire duration. Or do you think it's still too generic?

We could also change it to something like `io_ui_thread` which would also work for Network I/O or Database I/O

brustolin: IMO `blocked_ui_thread` is a good option

marandaneto: @AbhiPrasad we are already using the `data` field without prefixes so I am fine with it, even though they are "reserved" by sentry such as [http.status_code](https://github.com/getsentry/develop/issues/445), so, I agree.

@romtsn fine with the `blocked_x_thread` as well, but maybe we use `main` instead of `ui`? the same would apply for a CLI app that has its main thread blocked but its not an UI app.

romtsn: I'm fine with `main`, but `ui` sounds like it needs more attention (in the mobile/desktop world).

Also, if the main thread is blocked in a CLI app it's.. probably fine(?), since there's no user interface involved and maybe it's intentional

text/0036-auto-instrumentation-ui-thread.md
```diff
@@ -0,0 +1,127 @@
+* Start Date: 2022-11-08
+* RFC Type: decision
+* RFC PR: [#36](https://github.com/getsentry/rfcs/pull/36)
+* RFC Status: -
+* RFC Driver: [Roman Zavarnitsyn](https://github.com/romtsn)
+
+# Summary
+
+Add new `blocked_ui_thread` attribute to the `data` key-value map of the [`Span`](https://develop.sentry.dev/sdk/event-payloads/span/) interface, indicating whether
+an auto-instrumented span has run its entire duration on the UI/Main thread.
```
marandaneto, line 10: I read somewhere that ui thread blocked thresholds are:

> performance interaction latency thresholds
> good 0-199 ms
> needs improvements 200 - 499 ms
> poor >= 500  ms

So in case the total amount of all the blocked UI thread spans, we could actually raise performance issues in case it's `poor`, maybe?

romtsn: Probably here https://web.dev/inp/#whats-a-good-inp-value

idk, on the one hand android's strictmode warns you about any kind of file i/o performed on main thread, on the other hand it's probably too strict. For now we've started with 16ms (that's when a frame drop occurs on most devices), but I guess we'll tune this based on customer's data/feedback.

text/0036-auto-instrumentation-ui-thread.md
marandaneto: The blocked `ui` thread can also be intentional, I'm just trying to make it more generic so it's applicable for other platforms, your call here.
https://developer.android.com/guide/components/processes-and-threads mentions many times the name `main`, also if you get the current `Looper` on Android, its called `main`, so its semantically correct.

romtsn: I'm good with `main`, but would need @wmak to change it on the detector's site then

wmak: Feel free to change it to main :+1: I can update the detectors. Just let me know what the final decision is

romtsn: alright, I changed it to `blocked_main_thread` on Android and will update the RFC as well. cc @wmak @brustolin

text/0036-auto-instrumentation-ui-thread.md
```diff
@@ -0,0 +1,126 @@
...
+    "file.size": 4010
+  },
+  "hash": "8add714f71a52ef2"
+}
+```
+
+## (De)obfuscation
+
+As the `call_stack` may contain obfuscated frames we need to process them server-side, similar to what we already do for [profiles]
+(https://github.com/getsentry/sentry/blob/cf71af372677487d7d0a7fd8ac9dd092f9596cf4/src/sentry/profiles/task.py#L350-L360). In addition, after de-obfuscation, the frames
```
romtsn, line 90:
```suggestion
As the `call_stack` may contain obfuscated frames we need to process them server-side, similar to what we already do for [profiles](https://github.com/getsentry/sentry/blob/cf71af372677487d7d0a7fd8ac9dd092f9596cf4/src/sentry/profiles/task.py#L350-L360). In addition, after de-obfuscation, the frames
```

brustolin (approved):
This looks very good!

PR #37, opened 2022-11-15
rfc: Calculate ANR rates with sessions data

shruthilayaj:
Calculate ANR (App Not Responding) rates with session 
backed data to match Google Play apps Android vitals 
as closely as possible.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/shruthi/rfc/accurate-anr-rates/text/0037-anr-rates.md)


text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+    AND in((project_id AS _snuba_project_id), [...])
+    AND greaterOrEquals((timestamp AS _snuba_timestamp), toDateTime('...', 'Universal'))
+    AND less(_snuba_timestamp, toDateTime('...', 'Universal'))
+    AND in(_snuba_metric_id, [user])
+  LIMIT 2
+```
+
+## Drawbacks
+
+1. We are introducing a new terminal session state but the android system displays a dialogue asking the user if they would like to wait or force quit the app. So an ANR doesn't necessarily mean a terminating session state. Downstream session status changes will no longer be captured.
```
untitaker, line 102: 1. there is also already a status for the session update we send when an error happens within the application: `errored`. Are we saying that ANR errors do not mark a session as errored anymore? If ANR does not become its own terminal state, do we still consider the session crashed if it contained an ANR at some point?

2. if ANR is not terminal, how do you deal with multiple ANR in the same session? (that is a question for both proposals)

   i assume it doesn't matter for user count at a high level, but we need to figure out if we want to send multiple session updates

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+Issues outlined in 1 & 2 will result in us showing *wrong* ANR rates and 3 will limit the capabilities of ANR rate - can’t allow sort and search, can’t add it to dashboards or alerts and is not future-proof in case we want to use ANR rates in issue detection.
+
+# Supporting Data
+
+
+# Options Considered
+
+## Introduce a new tag in the sessions dataset (Option 1)
+
+Introduce a new optional tag `freeze_mechanism` in the sessions dataset and track ANRs in sessions in addition to sending the ANR error events.
```
untitaker, line 52: this seems more straightforward to me. we already send `errored` session updates afaik when the application errored, so we could just add a `mechanism` tag to it (and restrict its cardinality to the values we actually care about)

(not sure if we actually do this in all sdks already but we do have affordances for it in the protocol)

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+```bash
+(unique users who experienced error events with tag mechanism: ANR)
+___________________________________________________________________ x 100 
+                    (total unique users)
+```
+
+There are a couple problems with this:
+
+1. Total ANRs is affected by client side sampling and dropped events if org/project is close to error quota
+2. The most accurate count of total unique users will probably come from the sessions dataset, or maybe even the metrics dataset (but again this one is affected by client-side sampling)
```
untitaker, line 40: i don't think we have any features for client-side sampling for sessions

untitaker (commented):
if at some point dynamic sampling for errors becomes a thing, we could theoretically extract an ANR count on the serverside instead of updating all sdks. i.e. derive metrics from error events. i don't think this is an option right now but IMO it would be the ideal one

text/0037-anr-rates.md
shruthilayaj: Ah yeah, by "metrics" here I meant the processed transactions/generic metrics - that would still be affected by client-side sampling. although come thing of it, can we even get unique users from that data? in any case yeah, the sessions data will have what we need

text/0037-anr-rates.md
shruthilayaj: Agreed that this is the more straightforward solution. I like the idea of attaching mechanism instead restrict cardinality cc: @romtsn

text/0037-anr-rates.md
marandaneto: Sessions don't do client-side sampling but error events do (such as an ANR event), so a user could technically experience 3 ANRs but maybe only 1 one of them got sampled.
If we want a more accurate rate, maybe we should expand the session metadata.

text/0037-anr-rates.md
marandaneto: ANRs don't change the `status` of the session, but rather increase the `errorCount`, the BE or FE then displays as `Errored`, since errored actually isn't a valid status.

marandaneto: Btw if the Android team works on https://github.com/getsentry/sentry-java/issues/1796 which is planned for this Q, this could affect the ANR rate calculation, let's assume that we go for the approach that only reports ANRs on App restart, that means that the session status was actually `crashed` and not only with an `errorCount`.

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
+* Start Date: 2022-10-15
+* RFC Type: decision
+* RFC PR: [#37](https://github.com/getsentry/rfcs/pull/37)
+* RFC Status: draft
+
+# Summary
+
+Calculate ANR (App Not Responding) rates with session backed data to match Google Play apps Android vitals as closely as possible. is the The more desirable option between the two presented in this RFC is introducing a new, optional tag (name TBD, but options include `mechanism`, `error.mechanism`, `freeze_mechanism`) in sessions data to track the type of ANR which will allow us to calculate ANR rate without discrepencies introduced by client-side sampling. This tag will only be set by the Android SDK for sessions that experience an ANR. We will be limiting the cardinality of the tag values; currently expecting to distinguish between user-perceived vs other ANRs, with a hope to extend to a third value to track AppHangs from iOS in the future.
```
dashed, line 8:
```suggestion
Calculate ANR (App Not Responding) rates with session backed data to match Google Play app's Android vitals as closely as possible. The more desirable option between the two presented in this RFC is introducing a new, optional tag (name TBD, but options include `mechanism`, `error.mechanism`, `freeze_mechanism`) in sessions data to track the type of ANR which will allow us to calculate the ANR rate without discrepancies introduced by client-side sampling. This tag will only be set by the Android SDK for sessions that experience an ANR event. We will be limiting the cardinality of the tag values; currently expecting to distinguish between user-perceived vs other ANRs, with a hope to extend to a third value to track AppHangs from iOS in the future. 
```

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+```bash
+(unique users who experienced error events with tag mechanism: ANR)
+___________________________________________________________________ x 100 
+                    (total unique users)
+```
+
+There are a couple problems with this:
+
+1. Total ANRs is affected by client side sampling and dropped events if org/project is close to error quota
+2. The most accurate count of total unique users will probably come from the sessions dataset
```
fpacifici, line 40: You want to replace `sessions` dataset everywhere as that dataset is being retired. Now release health data is powered by metrics.

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+  LIMIT 2
+```
+
+1. This is a more generic solution that’s extensible to "freeze mechanisms" (ex. app hangs on iOS) from other operating systems
+2. This can also be extended to track the four different types of ANRs outlined [here](https://developer.android.com/topic/performance/vitals/anr). Differentiating ANRs due to `Input dispatching timed out` separately will allow us to calculate user-perceived ANR rate
+3. Stand-alone addition that won’t require any migration, less risky, doesn’t affect existing calculations
+
+## Drawbacks
+
+1. Increases the cardinality of the sessions data, however this tag would only be set by Android SDKs for now, so would only affect Android sessions
```
fpacifici, line 79: Increasing cardinality is not an issue per se. It increases cost but it is manageable.

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+Issues outlined in 1 & 2 will result in us showing *wrong* ANR rates and 3 will limit the capabilities of ANR rate - can’t allow sort and search, can’t add it to dashboards or alerts and is not future-proof in case we want to use ANR rates in issue detection.
+
+# Supporting Data
+
+
+# Options Considered
+
+## Introduce a new tag in the sessions dataset (Option 1)
+
+Introduce a new optional tag `freeze_mechanism` in the sessions dataset and track ANRs in sessions in addition to sending the ANR error events.
```
fpacifici, line 52: IF you will go for a tag based solution please ensure you register the tag name and the most common values you expect in: https://github.com/getsentry/sentry/blob/master/src/sentry/sentry_metrics/indexer/strings.py

And get a review by somebody familiar with the indexer (SnS or ingest). 
This will reduce the impact on the indexer.

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+## Introduce a new tag in the sessions dataset (Option 1)
+
+Introduce a new optional tag `freeze_mechanism` in the sessions dataset and track ANRs in sessions in addition to sending the ANR error events. 
+
+SDK sends a session update (`freeze_mechanism:ANR`) when it hits an ANR in addition to creating an ANR error event.
+
+We will then calculate ANR rate with a query similar to counting abnormal or errored users:
+
+```sql
+SELECT
```
jan-auer, line 59: Please note that queries like these should be written directly as SnQL, but rather using the [metrics abstraction layer](https://github.com/getsentry/sentry/blob/5d0b682af5c1b71a8ae5b88f7e3ed090264ed1b5/src/sentry/snuba/metrics/datasource.py#L667-L672). 

You would define a [derived metric](https://github.com/getsentry/sentry/blob/5d0b682af5c1b71a8ae5b88f7e3ed090264ed1b5/src/sentry/snuba/metrics/fields/base.py#L1253) for this that internally evaluates the expression you're specifying here, and the metrics layer can then evaluate that expression in all places it is used (e.g. the sessions API).

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+<aside>
+💡 User-perceived ANR rate is a *core vital* meaning that it affects the discoverability of your app on Google Play.
+</aside>
+
+
+ANRs are currently surfaced as error events with tag `mechanism:ANR`. With the data we already have on hand, we can calculate ANR rate as follows:
+
+```bash
+(unique users who experienced error events with tag mechanism: ANR)
+___________________________________________________________________ x 100
```
jan-auer, line 33: This is a nitpick, but could we define it as just the ratio without `x 100`? It seems like a concern for the presentation layer to format `0.32` as 32%.

jan-auer (commented):
This proposal looks good to me, the motivation is clear and the proposed implementation makes sense with reasonable tradeoffs. 

I would just ask to incorporate the open comments before merging this. Especially, please describe the protocol changes and changes to session -> metrics conversion in slightly more detail and align the data terminology.

Given the very limited values of the added field, @fpacifici already pointed out that cardinality is not a large issue. A large number of projects sends sparse session updates, so the increase in storage should be manageable.

text/0037-anr-rates.md
```diff
@@ -0,0 +1,107 @@
...
+# Supporting Data
+
+
+# Options Considered
+
+## Introduce a new tag in the sessions dataset (Option 1)
+
+Introduce a new optional tag `freeze_mechanism` in the sessions dataset and track ANRs in sessions in addition to sending the ANR error events. 
+
+SDK sends a session update (`freeze_mechanism:ANR`) when it hits an ANR in addition to creating an ANR error event.
```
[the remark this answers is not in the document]
shruthilayaj: @jan-auer yes, you've summed it up correctly. Updated the text to reflect the detail.
> Is the new field sent for every error update or just when the status is `abnormal`?
I'm going to defer to @romtsn on this.

Since ANR rates by definition are calculated based on unique users experiencing ANRs, I am fairly certain we only need to extract this metric on `user` (but I will double check with product to make sure there isn't an expectation to count the number of sessions as well)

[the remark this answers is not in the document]
shruthilayaj: Confirming that we would only need to extract user metrics for ANR rates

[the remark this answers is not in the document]
romtsn: as agreed with @jan-auer and @shruthilayaj during our sync, let's change the SDK implementation to set session status to `abnormal` when we face an ANR

text/0037-anr-rates.md
```diff
@@ -0,0 +1,123 @@
...
+# Options Considered
+
+## Introduce a new tag in the release health metrics dataset (Option 1)
+
+Introduce a new optional tag `abnormal_mechanism` in the release health metrics dataset to track ANRs with sessions in addition to sending the ANR error events. 
+
+On the SDK side, this will result in the following changes:
+  - Add an optional top-level field called `abnormal_mechanism` with values of type `string` to the [session update payload](https://develop.sentry.dev/sdk/sessions/#session-update-payload)
+  - SDK sends a session update with `abnormal_mechanism` tag and an `abnormal` session status when it detects an ANR
+  - Allowed string values for the `abnormal_mechanism` tag are `'anr_background'` and `'anr_foreground'`
```
romtsn, line 57: @jan-auer could you check if that's fine for you? `abnormal_mechanism` seems to be a bit more generic and opens the door for new values (e.g. app hangs or OOMs from ios), if/when we need this.

PR #38, opened 2022-11-21
Scrubbing sensitive data

antonpirker:
Make scrubbing of sensitive data (like security related data (passwords, keys, et al) and PII (personal identifiable information) in Sentry smarter. 

[Rendered RFC](https://github.com/getsentry/rfcs/blob/antonpirker/scrubbing-sensitive-data/text/0038-scrubbing-sensitive-data.md)

### DIFF
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -52,7 +52,7 @@ _Cons:_
 
 ### Option B): Store data in a structured way in SDKs
 
-The content should not be a simple string but strucutred data. The content could be a template string where all the sensitive values are removed plus a list of values. (ex `select * from user where email=%s;` and `{ "email": "test@example.com" }`) This way it would be easy for Relay to remove sensitive information by field name.
+The content should not be a simple string but structured data. The content could be a template string where all the sensitive values are removed plus a list of values. (ex `select * from user where email=%s;` and `{ "email": "test@example.com" }`) This way it would be easy for Relay to remove sensitive information by field name.
 
 _Pros:_
 
@@ -60,7 +60,7 @@ _Pros:_
 
 _Cons:_
 
--
+- Could be complex for nested JSON objects or for "Array of objects" kind of data.
 
 ### Option C): Relay identifies what kind of data is present and parses it.
 
@@ -73,11 +73,11 @@ _Pros:_
 
 _Cons:_
 
--
+- Could be expensive to try multiple guesses before the right kind of data is identified. (Maybe its SQL? no. Maybe JSON? no. So it is a URL? yes.)
 
 ### Option D): Generic tozenization in Relay.
 
-Have a generic tokenizer in Relay that can not parse full fledged SQL, but can extract key/value pairs out of almost everything. This this the values of keys with potential sensitive information can be removed.
+Have a generic tokenizer in Relay that can not parse full fledged SQL, but can extract key/value pairs out of almost everything. With this the values of keys with potential sensitive information can be removed.
 
 _Pros:_
 
@@ -89,7 +89,7 @@ _Cons:_
 
 # Drawbacks
 
-(tbd)
+(none)
 
 # Unresolved questions
 
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -131,7 +131,5 @@ _Cons:_
 
 # Unresolved questions
 
-(tbd)
-
-- What parts of the design do you expect to resolve through this RFC?
-- What issues are out of scope for this RFC but are known?
+- We need to check with legal and/or security to make sure that the stuff we are planing is actually OK with existing laws and regulations
+- We need to find all places in SDKs that sensitive data could be. Places we are targeting right now: `span.description/span.data`, `breadcrumb.message/breadcrumb.data`, `logentry.message/logentry.params`, and `message.message/message.params`, local variables, request bodies, response bodies, HTTP headers, cookies, ...
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -109,6 +109,20 @@ _Cons:_
 
 -
 
+### NEW! Option E): Improved regexes
+
+Keep the logic on how data scrubbing is done right now the same, but improve the regexes to be more specific. Especially the "password regex" could be changed to the `auth` rule does ONLY match `auth` but NOT `author` or `authorize`.
+
+With this we could add data scrubbing back to `span.description` (and potentially other fields that ware marked with `pii=maybe` right now).
+
+_Pros:_
+
+-
+
+_Cons:_
+
+-
+
 # Drawbacks
 
 (none)
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -24,6 +24,15 @@ A user complained that our data scrubbing did not remove sensitive data from an 
 
 (tbd)
 
+# Conclusion
+
+Not one option will be implemented but a combination of options:
+
+1. The SDKs will try to save structured data where possible (Option B)
+2. The SDKs will try to specify what kind of data is in `span.description` (and other places) as good as possible. (Can be done by more fine grained `span.op`s or applying OTel trace semantic convention to `span.data`. Needs to be speced out)
+3. Relay uses data from 2.) to parse content fields and scrub sensitive data (Option C)
+4. If there is no structured data (so 3.) is not possible) Relay uses a more fine grained set of Regexes to remove sensitive data. Relay does not remove the complete content but only the sensitive part of it. (Option D+E)
+
 # Options Considered
 
 ### Status Quo:
@@ -37,14 +46,6 @@ A user complained that our data scrubbing did not remove sensitive data from an 
 - The regexes for data scrubbing are defined [here](https://github.com/getsentry/relay/blob/92a4b349f271963a53c8a8278acb3d4d56f0dfe5/relay-general/src/pii/regexes.rs#L103-L274)
 - Some regexes can just remove the sensitive part of the content (like IP and SSH keys regexes).
 - Some of the regexes (like password regex) will remove the complete content of a field. This is because it is unstructured data. Relay does not know if the content is a SQL query, a JSON object, an URL, a Elasticsearch/MongoDB/whatever query in JSON format, or something else.
-
-### Currently perferred solution:
-
-A combination of some of the options listed below:
-
-- Have the SDKs store structured data (Option B)
-- If Relay does not get any structured data from Option B, Relay does Option C.
-- If Relay can not figure out what kind of data is in the field, do matching like we do now.
 
 ### Option A): Remove Sensitive Data in SDKs
 
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -60,6 +60,8 @@ We need to identify all the fields we need to do this.
 
 We should model this Option after how `Logentry` does this now.
 
+Note: If we change the span.description then the hash for existing performance issues will be changed and existing performance issues will be recreated, so users would have duplicate performance issues in their list of issues. (Which we can just document in the CHANGELOG and everything should be fine.)
+
 _Pros:_
 
 - Relay would not have to reverse-engineer the semantics of the information supplied by the SDK.
--- a/text/0038-scrubbing-sensitive-data.md
+++ b/text/0038-scrubbing-sensitive-data.md
@@ -1,34 +1,96 @@
-* Start Date: 2022-11-21
-* RFC Type: feature 
-* RFC PR: https://github.com/getsentry/rfcs/pull/38 
+- Start Date: 2022-11-21
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/38
 
 # Summary
 
-One paragraph explanation of the feature or document purpose.
+Make scrubbing of sensitive data (security related data (passwords, keys, et al) and personal identifiable information (PII)) in Sentry smarter.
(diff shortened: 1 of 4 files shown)


jjbayer (commented):
I think we should do both B _and_ C, potentially also D. SDKs should provide as much structure as possible, but if they can't / for old SDKs, Relay should have some logic to parse key-value pairs out of strings.

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,96 @@
...
+(tbd)
+
+# Options Considered
+
+### Status Quo:
+
+- Right now most data scrubbing is done in Relay.
+- There is a option `sendDefaultPII` in SDKs that may or may not remove some sensitive data before sending.
+- In Relay fields have a "pii" attribute that can be `yes`, `no` or `maybe`:
+  - `yes` means all the data scrubbing Regexes defined [here](https://github.com/getsentry/relay/blob/master/relay-general/src/pii/regexes.rs#L103-L274) are applied to the fields. If the one Regex matches the whole content of the field is removed and an remark is saved telling us what regex was responsible for the removal of the data.
```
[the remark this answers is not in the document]
antonpirker: Thanks! Will update!

antonpirker:
@jjbayer suggestion is:
- SDKs store the data in a structured way (option B)
- Relay than parses this structured data and removes sensitive information (option B and kind of option C)
- If unstructured data flows in from old SDKs Relay tries to identify and scrub it (option C)
- If Relay can not identify the content, it will do plain text matching like now

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Status Quo:
+
+- Right now most data scrubbing is done in Relay.
+- There is a option `sendDefaultPII` in SDKs that may or may not remove some sensitive data before sending.
+- In Relay fields have a "pii" attribute that can be `yes`, `no` or `maybe`:
+  - `yes` means all the default data scrubbing Regexes are applied to the fields.
+  - `maybe` means that the data scrubbing Regexes are not run by default. If the user has advanced data scrubbing set (in Sentry.io under: Project Settings > Security & Privacy > Advanced Data Scrubbing) those custom rules are applied to the field and if a rule matches the whole content of the field is removed.
+  - `no` there is no possibility of data scrubbing for this field.
+- The regexes for data scrubbing are defined [here](https://github.com/getsentry/relay/blob/master/relay-general/src/pii/regexes.rs#L103-L274)
```
marandaneto, line 37:
```suggestion
- The regexes for data scrubbing are defined [here](https://github.com/getsentry/relay/blob/92a4b349f271963a53c8a8278acb3d4d56f0dfe5/relay-general/src/pii/regexes.rs#L103-L274)
```

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+_Pros:_
+
+-
+
+_Cons:_
+
+-
+
+### Option B): Store data in a structured way in SDKs
```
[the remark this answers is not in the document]
marandaneto: The other point is that it'd only work for auto instrumentation (because we apply the structured data) or when people follow our guidelines, but when they don't, there might be PII, what do we do in such cases?

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+_Pros:_
+
+- Relay would not have to reverse-engineer the semantics of the information supplied by the SDK.
+
+_Cons:_
+
+- Could be complex for nested JSON objects or for "Array of objects" kind of data.
+
+### Option C): Relay identifies what kind of data is present and parses it.
```
marandaneto, line 81: I like this option because we don't run into this issue https://github.com/getsentry/rfcs/pull/38/files#r1029153081

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+- https://github.com/getsentry/sentry/blob/68e44ed3e8343a5e69d0b0a51ad65c02ae427cd0/src/sentry/spans/grouping/strategy/base.py#L186
+- https://github.com/getsentry/sentry/blob/68e44ed3e8343a5e69d0b0a51ad65c02ae427cd0/src/sentry/spans/grouping/strategy/base.py#L142-L150
+
+_Pros:_
+
+-
+
+_Cons:_
+
+- Could be expensive to try multiple guesses before the right kind of data is identified. (Maybe its SQL? no. Maybe JSON? no. So it is a URL? yes.)
```
marandaneto, line 98: We could rely on the `op` as mentioned here https://github.com/getsentry/rfcs/pull/38/files#r1029150641
But then we run into the issue https://github.com/getsentry/rfcs/pull/38/files#r1029153081

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Option D): Generic tozenization in Relay.
+
+Have a generic tokenizer in Relay that can not parse full fledged SQL, but can extract key/value pairs out of almost everything. With this the values of keys with potential sensitive information can be removed.
```
marandaneto, line 102: Likely the most expensive one, but it is the one that is implemented on Relay (a single source of truth for data scrubbing), and likely the one that degrades the least the product (scrubbing some key/values rather than the full URL).

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+# Drawbacks
+
+(none)
+
+# Unresolved questions
+
+(tbd)
+
+- What parts of the design do you expect to resolve through this RFC?
```
marandaneto, line 134: For me its unclear in this RFC if we would like to focus on scrubbing on Relay or if we don't even the data coming into Sentry (in this case, the focus should be on SDKs).

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,135 @@
...
+
+_Pros:_
+
+-
+
+_Cons:_
+
+-
+
+### NEW! Option E): Improved regexes
```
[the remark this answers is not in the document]
marandaneto: Very similar to `Option D` IMO.

marandaneto:
In general, I believe that Relay should do its job.
SDKs can be good citizens and do tokenization when possible.
SDKs often run into issues similar to Relay (we have a String and this is the URL), but it's not a template, similar to DB queries, we have a String and this is the statement, but it's not a template.
Focusing on Relay implementations, we guarantee that data scrubbing would be applied for custom instrumentation when people don't follow the guidelines of what's put in the `desc`, `op`, `data`, etc, or if they did that manually, that does not matter? not sure.

text/0038-scrubbing-sensitive-data.md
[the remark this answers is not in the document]
krystofwoldrich: I like this option as it's the most generic one.

smeubank:
@AbhiPrasad suggests using OTEL conventions for span ops naming. So SDKs and Relay can rely on the same convention for determining what the source is `db.query.sql` vs `db.query.mongoos` for example

@cleptric ensure we clarify that users can opt out of all of this

text/0038-scrubbing-sensitive-data.md
antonpirker: In this RFC we just want to find out the way we want to move in the future. And as it seems now most of the engineers are leaning towards Relay doing the scrubbing but are also on board with the SDKs providing better formatted data to make it easier for Relay to scrub the data.

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,136 @@
...
+# Supporting Data
+
+(tbd)
+
+# Conclusion
+
+Not one option will be implemented but a combination of options:
+
+1. The SDKs will try to save structured data where possible (Option B)
+2. The SDKs will try to specify what kind of data is in `span.description` (and other places) as good as possible. (Can be done by more fine grained `span.op`s or applying OTel trace semantic convention to `span.data`. Needs to be speced out)
```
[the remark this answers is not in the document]
antonpirker: Yes, @untitaker told me that the plan was (in the long run) to have structured data everywhere, but somehow this plan was never fully executed over all sdks and data fields. So this is our chance now to make the plan come reality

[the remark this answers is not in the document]
antonpirker: But the things that relay does now for 3. and 4. need some improvements. (like you commented here: https://github.com/getsentry/rfcs/pull/38#discussion_r1029157753)

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,136 @@
...
+
+-
+
+_Cons:_
+
+-
+
+# Drawbacks
+
+(none)
```
cleptric, line 131: We should at least mention that this could have an impact on the product.

untitaker: yes, adding/removing scrubbing on spans can have impact on performance issue grouping at the very least

antonpirker: I already checked with the performance issues team and they think it should not have an impact at all: 

Screenshot 2022-11-23 at 09 42 41

antonpirker: @cleptric what impacts to you have in mind?

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,136 @@
...
+
+_Pros:_
+
+-
+
+_Cons:_
+
+- Could be expensive to try multiple guesses before the right kind of data is identified. (Maybe its SQL? no. Maybe JSON? no. So it is a URL? yes.)
+
+### Option D): Generic tozenization in Relay.
```
jjbayer, line 103:
```suggestion
### Option D): Generic tokenization in Relay.
```

text/0038-scrubbing-sensitive-data.md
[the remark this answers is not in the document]
jjbayer: I think the difference between `D` and `E` is that `E` will still replace the entire string, but it will fire less often because the matching condition (the regex) is more specific.

jjbayer (commented):
I feel like Option D and E are immediately actionable. Option B would be a nice to have.

antonpirker:
What do you think @jjbayer (and everyone else) should we:
1) make the regexes more specific and remove the ENTIRE content 
2) make the regexes more specific and remove just the sensitive part (matching some delimiter chars)
3) start with 1) and then improve to 2)

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,136 @@
...
+
+- Could be expensive to try multiple guesses before the right kind of data is identified. (Maybe its SQL? no. Maybe JSON? no. So it is a URL? yes.)
+
+### Option D): Generic tokenization in Relay.
+
+Have a generic tokenizer in Relay that can not parse full fledged SQL, but can extract key/value pairs out of almost everything. With this the values of keys with potential sensitive information can be removed.
+
+_Pros:_
+
+-
```
philipphofmann, line 109: `m`: We could add the same pros as for option C.

text/0038-scrubbing-sensitive-data.md
```diff
@@ -0,0 +1,136 @@
...
+
+_Cons:_
+
+-
+
+### NEW! Option E): Improved regexes
+
+Keep the logic on how data scrubbing is done right now the same, but improve the regexes to be more specific. Especially the "password regex" could be changed to the `auth` rule does ONLY match `auth` but NOT `author` or `authorize`.
```
philipphofmann, line 117: `l`: 
```suggestion
Keep the logic on how data scrubbing is done right now, but improve the regexes to be more specific. Especially the "password regex" could be changed to the `auth` rule does ONLY match `auth` but NOT `author` or `authorize`.
```

text/0038-scrubbing-sensitive-data.md
philipphofmann, line 32: `h`: If the SDK already uses structured data, should it also specify what kind of data is in `span.description`? If yes, how do we prevent Relay from changing the already structured data?

philipphofmann (commented):
Most of my comments are nitpicks. I used the LOGAF scale to show you how important a comment is. 

- `l`: Low - Nitpick
- `m`: Medium -  worth having a look
- `h`: High - important comment

jjbayer:
> 1. make the regexes more specific and remove the ENTIRE content
> 2. make the regexes more specific and remove just the sensitive part (matching some delimiter chars)
> 3. start with 1) and then improve to 2)

@antonpirker I think that "matching some delimiter chars" essentially comes down to Option D. It would be very hard to generalize "apply regex only to a section of the string". For example, if we declare whitespace to be a delimiter, we would not catch the token in `Authorization: Bearer <token>`.

That said, I actually think Option D will bring us more value than Option E, so I would start with D.

text/0038-scrubbing-sensitive-data.md
cleptric: This might degrade the usefulness of the product or could have unknown impact if we accidentally scrub too much.

antonpirker:
I have now updated the conclusion to point out that we are implementing option D) rather than option E): https://github.com/getsentry/rfcs/blob/antonpirker/scrubbing-sensitive-data/text/0038-scrubbing-sensitive-data.md#conclusion

@jjbayer are you confident that we can implement a generic tokenizer that can extract key/value pairs out of any unstructured data without leaking sensitive data?

jjbayer:
> @jjbayer are you confident that we can implement a generic tokenizer that can extract key/value pairs out of any unstructured data without leaking sensitive data?

@antonpirker Actually, I am confident that we *cannot*. Data scrubbing remains a best-effort deal. But I think we can improve the status quo of not scrubbing certain fields at all.

PR #40, opened 2022-11-28
rfc: SDKs report file I/O on the main thread

philipphofmann:
This RFC aims to revisit the decision that Ingest creates file I/O on the main thread performance issues and not the SDKs.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/sdk-perfomance-issues/text/0039-sdks-report-file-IO-on-main-thread.md)


text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. No need for per-SDK rollout.
+2. Less SDK overhead as they don't have to detect the issues.
+3. Changing the algorithm doesn't require SDK rollout.
+
+### Cons
+
+1. Users have to enable performance monitoring.
+2. A transaction is required to be running while a performance issue is happening.
+3. No stack traces, so no way to show where the performance problem was detected.
```
silent1mezzo, line 67: This isn't correct. The stack trace definitely makes it easier and more concrete as to where the performance issue is happening but we still highlight the spans that make up the issue so there is a way to show where it happened.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Summary
+
+This RFC aims to clarify which types of performance issues SDKs shall create.
+
+# Motivation
+
+On June 21, 2022, we decided with [DACI](https://www.notion.so/sentry/Performance-Issue-Creation-POC-e521772ebccb482b83b08f4f8a3db2cb) to create performance issues in Ingest. While implementing the file I/O on the main thread performance issue for Android, the question arose as to why SDKs don't report easy-to-detect and matured performance issues. 
+
+# Options Considered
```
silent1mezzo, line 15: Options 1 and 2 are missing some Cons which is why the Ingest side was chosen

1. The detectors must be maintained across all SDKs that detect a specific issue.

e.g. N+1 in Django, Rails, Node, Laravel, etc...
File I/O in both Android and iOS.

2. Stack traces are inherently expensive to process.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+We already decided against this option in this [DACI](https://www.notion.so/sentry/Performance-Issue-Creation-POC-e521772ebccb482b83b08f4f8a3db2cb#907db42314864ae2a4b5348835c250c9)
+
+### Pros
+
+1. SDKs can capture a stack trace which will help with actionability.
+2. No running transaction required.
+3. Can tie together transaction and error far more easily since both objects exist at the point of time the performance issue will be created
+
+### Cons
```
silent1mezzo, line 47: Need for per-SDK rollout.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Double dipping, sending the transaction and the error created within that transaction.
+2. Code on SDK could cause overhead.
+3. Not able to use dynamic thresholds, code changes would be required to update settings.
+
+## Option 3: Ingest creates the performacne issues
```
silent1mezzo, line 53:
```suggestion
## Option 3: Ingest creates the performance issues
```

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+We still need to clarify what easy to detect and matured precisely means, which is not the goal of this document. For example, file I/O on the main thread could be a candidate.
+
+To clarify the threshold and configuration for performance issues, an experimental feature phase can help to get feedback.
+
+### Pros
+
+1. SDKs can capture a stack trace which will help with actionability.
+2. No running transaction required.
+3. Sentry can tie together transaction and error far more easily since both objects exist at the point of time the performance issue will be created.
```
wmak, line 27: I don't think this is true, since errors and transactions have different quotas its possible (and common) that errors exist without a transaction

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Pros
+
+1. No need for per-SDK rollout.
+2. Less SDK overhead as they don't have to detect the issues.
+3. Changing the algorithm doesn't require SDK rollout.
+
+### Cons
+
+1. Users have to enable performance monitoring.
```
wmak, line 65: Is this a con? Don't we want "Performance" Issues to be a part of performance monitoring?

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
...
+
+This RFC aims to clarify which types of performance issues SDKs shall create.
+
+# Motivation
+
+On June 21, 2022, we decided with [DACI](https://www.notion.so/sentry/Performance-Issue-Creation-POC-e521772ebccb482b83b08f4f8a3db2cb) to create performance issues in Ingest. While implementing the file I/O on the main thread performance issue for Android, the question arose as to why SDKs don't report easy-to-detect and matured performance issues. 
+
+# Options Considered
+
+## Option 1: SDKs report easy to detect and matured performance issues
```
wmak, line 17: Do we know if the SDK can create performance issues directly, or would they be error issues?

mjq-sentry: My understanding is that currently only errors can be created, but if we go down the SDK-created road I think we should be creating performance issues, not errors. It makes more sense to me from both a logical and UX perspective - perf issues and errors have some differences, so we should categorize them correctly.

That may open another can of worms, I don't know. Maybe billing implications too? AFAIK performance issues are currently free, which is manageable given we control the creation rate ourselves internally.

text/0039-sdk-performance-issue-creation.md
romtsn: yeah, we literally say at 2) that no running transaction is required :) or we're misunderstanding the phrasing here

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
+* Start Date: 2022-11-28
+* RFC Type: decision
+* RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
+* RFC Status: draft
+* RFC Driver: [Philipp Hofmann](https://github.com/philipphofmann)
+
+# Summary
+
+This RFC aims to clarify which types of performance issues SDKs shall create.
```
mjq-sentry, line 9: Does this need to be a generic RFC? I was thinking we might focus on this case (file I/O on the main thread) in particular, and make a decision about that one alone.

I don't think trying to come up with generic performance issue implementation strategies is that meaningful. We've been asked to start from root causes, and look at the best possible detection strategy for each one individually. Trying to delimit "ingest detectors" vs "SDK detectors" is looking at the problem backwards IMO.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. SDKs can capture a stack trace which will help with actionability.
+2. No running transaction required.
+3. Sentry can tie together transaction and error far more easily since both objects exist at the point of time the performance issue will be created.
+4. As we limit this only to easy detect peformance issues the SDK overhead should be minimal.
+
+### Cons
+
+1. Need for per-SDK rollout.
+2. Double dipping quotas, sending the transaction and the error created within that transaction.
```
mjq-sentry, line 33: This might depend on how we create the performance issue, as IMO it shouldn't be an error anyway.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
...
+3. Not able to use dynamic thresholds and configurations, code changes would be required to update settings.
+4. Stack traces are inherently expensive to process
+5. Detection is mixed between ingest and SDK
```
mjq-sentry, line 36: My gut feeling is detection is going to happen in many different ways over time (starting as soon as we want to detect something that doesn't fit the span detector model) - this might be inevitable (and therefore not a large concern).

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
...
+1. No need for per-SDK rollout.
+2. Less SDK overhead as they don't have to detect the issues.
+3. Changing the algorithm, or thresholds doesn't require SDK rollout, which means we can address customer issues far faster.
+4. Consistent location for detector logic
+
+### Cons
+
+1. Users have to enable performance monitoring.
+2. A transaction is required to be running while a performance issue is happening.
+3. Transactions are high volume which means we may send too many stack traces, so we can't have stacktraces only the span that caused the issue. Which makes it harder to identify the root cause.
```
mjq-sentry, line 72: We don't have stacktraces at all in transactions, right?

text/0039-sdk-performance-issue-creation.md
wmak: Thanks! I'll add this as a note since this is a big unknown currently even if we go with this solution, while option 3 steers away from these unknowns

text/0039-sdk-performance-issue-creation.md
wmak: That's correct

text/0039-sdk-performance-issue-creation.md
wmak: 🤔 Right I think its the "far more easily" here I have issue with since its actually not that easy until Dynamic Sampling can do Error Biasing

text/0039-sdk-performance-issue-creation.md
philipphofmann: Good point, thanks 🙏 .

text/0039-sdk-performance-issue-creation.md
philipphofmann: Good point @wmak, I created an extra option. That's why I like RFCs.

text/0039-sdk-performance-issue-creation.md
philipphofmann: If we report file I/O on the main thread as an error, then it's a con.

text/0039-sdk-performance-issue-creation.md
philipphofmann: The protocol would support it, though, as transactions are events.

text/0039-sdk-performance-issue-creation.md
```diff
@@ -0,0 +1,72 @@
...
+2. No running transaction required.
+3. Sentry can tie together transaction and error far more easily since both objects exist at the point of time the performance issue will be created.
+4. As we limit this only to easy detect peformance issues the SDK overhead should be minimal.
+
+### Cons
+
+1. Need for per-SDK rollout.
+2. Double dipping quotas, sending the transaction and the error created within that transaction.
+3. Not able to use dynamic thresholds and configurations, code changes would be required to update settings.
+4. Stack traces are inherently expensive to process
```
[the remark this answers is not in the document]
philipphofmann: I agree in that specific case it's not a con. I removed it.

text/0039-sdk-performance-issue-creation.md
philipphofmann: I added an extra option 2 and added this as a pro there.

philipphofmann:
I changed the RFC quite a bit by adding an extra option. I have a hard time finding pros for option 2. @mjq-sentry, as you advocated for that (https://github.com/getsentry/rfcs/pull/40#discussion_r1033954355), could you please point out some pros?

text/0039-sdks-report-file-IO-on-main-thread.md
```diff
@@ -0,0 +1,69 @@
...
+## Option 2: SDKs report FIOMT as performance issues
+
+SDKs detect and report FIOMT as a performance issue. To achieve this need to:
+
+1. answer billing questions.
+2. make changes in Ingest to accept performance issues.
+
+### Pros
+
+1. No double dipping of quotas.
```
mjq-sentry, line 45: 2. FIOMT is correctly categorized as a Performance Issue and returned when searching inside Sentry for performance issues
2. FIOMT's issue details are presented as a Performance Issue, highlighting root causes and resources to fix the issue
2. Performance Issue-specific quotas and thresholds apply to FIOMT
2. Planned UX and workflow changes specific to performance issues also apply to FIOMT issues

With the opposite of all these things being cons of option 1 😄

philipphofmann: Thank you @mjq-sentry 💯

text/0039-sdk-performance-issue-creation.md
wmak: But its not an error, its a performance issue right?

philipphofmann: I added an extra option for that, see https://github.com/getsentry/rfcs/blob/c5c2613d9e034652d2322d40e6b97ed60d1e8169/text/0039-sdks-report-file-IO-on-main-thread.md?plain=1#L36-L56

and another one for errors
https://github.com/getsentry/rfcs/blob/c5c2613d9e034652d2322d40e6b97ed60d1e8169/text/0039-sdks-report-file-IO-on-main-thread.md?plain=1#L17-L34

text/0039-sdks-report-file-IO-on-main-thread.md
```diff
@@ -0,0 +1,72 @@
...
+2. Sentry correctly categorizes FIOMT as a performance issue and returns it when searching inside Sentry for performance issues.
+3. Sentry presents FIOMT as a performance issue, highlighting root causes and resources to fix the issue.
+4. Performance-issue-specific quotas and thresholds apply to FIOMT.
+5. Planned UX and workflow changes specific to performance issues also apply to FIOMT issues.
+
+### Cons <a name="option-2-cons"></a>
+
+1. Need for per-SDK rollout.
```
mjq-sentry, line 53: In this case is that just 2 SDKs (Android and Cocoa) or would it be more? (don't know if e.g. react-native would need its own implementation as well)

silent1mezzo:
So, there are a few things that were outlined in the original decision that never made it to long-term storage that I think is necessary for this discussion.

1. Performance issues differ from error issues in the sense that they don't cause an exception or otherwise stop the code from running and have a perceived performance impact on end-users.
2. Stack traces aren't available on most of our supported platforms, and even when it is, its generally too expensive to run against transactions. It's also not required since we can show spans to point people to the issue.
3. It's not possible to scale performance issues across the SDKs we support. Having it in Ingest (and eventually Relay) means there's a single point to build these detectors.

While we are thinking of **FIOMT** primarily on mobile it's also possible on desktop applications and javascript (node). Building the detector at the Ingest layer means it could detect any SDK that sends a span with `blocked_main_thread: true` as outlined in #36.

text/0039-sdks-report-file-IO-on-main-thread.md
philipphofmann: That applies to every mobile or desktop SDK. As far as I know, RN, Flutter, and .NET would also need implementations.

philipphofmann:
Thanks, @silent1mezzo, I added your input to the RFC with https://github.com/getsentry/rfcs/pull/40/commits/75962ad5f61dad6bc298d638c76357d378bf8fb8

silent1mezzo:
After reading through the RFC and discussion with a number of engineers the best option for File I/O on the main thread performance issues is **Option 3**, reporting it as a performance issue through the regular perf issue detectors. 

Option 1 was ruled out as this is not an error issue as per the definition I [mentioned](https://github.com/getsentry/rfcs/pull/40#issuecomment-1331532369). Option 3 was chosen over Option 2 because we can easily detect it via spans ([PR is already complete](https://github.com/getsentry/sentry/pull/41646/)) and the problem exists across multiple SDKs (iOS, Android, RN, Flutter, .NET and possibly more). There's also a likely future where FIOMT is detected via profiles and running detection in the ingestion pipeline will help combine Performance and Profiling issues.

philipphofmann:
@silent1mezzo, please review the documented decision and either give feedback or approve this PR. Thanks 🙏

silent1mezzo (approved):
👍 Looks good

PR #43, opened 2022-12-07
Introduce `instruction_addr_adjustment` Stack Trace Attribute

Swatinem:
This RFC introduces a new StackTrace Protocol field that controls adjustment of
the `instruction_addr` for symbolication.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/instruction-addr-adjustment/text/0043-instruction-addr-adjustment.md)

### DIFF
--- a/text/0043-instruction-addr-adjustment.md
+++ b/text/0043-instruction-addr-adjustment.md
@@ -101,7 +101,7 @@ frames should be.
 The above `instruction_addr_adjustment` attribute will be used in the SDK -> Sentry protocol. The internal Sentry -> Symbolicator
 protocol will add a per-frame attribute:
 
-**`needs_adjustment: Option<bool>`**, _new_
+**`adjust_instruction_addr: Option<bool>`**, _new_
 
 This is a _per-frame_ attribute which will do `instruction_addr` adjustment when set to `true`. The default depends on
 the whole stack trace. If _any_ stack frame has an explicit value set, it will default to `Some(true)`. Otherwise it
@@ -111,7 +111,7 @@ implementation.
 For example:
 
 ```rust
-let default_adjustment = if frames.iter().any(|frame| frame.needs_adjustment.is_some()) {
+let default_adjustment = if frames.iter().any(|frame| frame.adjust_instruction_addr.is_some()) {
   Some(true)
 } else {
   None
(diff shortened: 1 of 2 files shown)


mitsuhiko:
I though the deduplication actually happens on the server side. Either way though it sounds like profiling probably wants to do the heuristic manually. That said I wonder if the new adjustment is a question for the ingestion protocol *and* symbolicator or if in symbolicator this should be handled separately.

I am wondering if form the perspective of the symbolicator interface it wouldn't make more sense to add a `is_first` flag into every single frame into the symbolication payload. That way profiling could deduplicate into `(is_first, addr)` and still use the `auto` mode.

mitsuhiko:
Just to be clear I would only make it a per-frame attribute on the symbolication side (sentry -> symbolicator).

Swatinem:
After a discussion with the profiling team, we came to the following conclusion:

- Implement the SDK -> Sentry protocol as proposed in this RFC.
- Implement a per-frame "needs adjustment" boolean flag in the internal Sentry -> Symbolicator protocol.
- For backwards compatibility, the `instruction_addr_adjustment: "auto"` (or `undefined`) will *not* add any per-frame attributes in the Sentry -> Symbolicator call.
- If none of the frames has the per-frame attribute set, symbolicator will default to the existing `"auto"` behavior. (eg. `let use_auto = frames.iter().all(|frame| frame.should_adjust.is_none());`)
- The default will be "needs adjustment", so ideally only the first frame in Sentry->Symbolicator stack traces needs to be marked.
- Profiling should mark frames as "does not need adjustment" by marking every top frame in the stacks, eg: `stack_traces.iter().flat_map(|frames| frames.first()).for_each(|frame_idx| global_frame_list[frame_idx].needs_adjustment = Some(false))`
- Sorry for the Rust-style oneliners :-)

Swatinem:
Another thing I just noticed: We are only doing the adjustment for absolute addresses, and *NOT* for relative addressing mode (aka WASM).
I believe that is the correct thing to do, right?

Swatinem:
@bitsandfoxes we had a look at most of the events you sent us, they look consistent across unity versions on one platform, but different across platforms.

Here is what we found out:

- On Android arm64 (across all editor versions), Unity reports already adjusted addresses, so no further adjustment on our end needs to happen
  -> SDK should report `instruction_addr_adjustment: "none"`
- On macOS arm64 and iOS x86_64 (simulator build?), Unity reports return addresses for all the frames, so symbolicator needs to adjust all of them
  -> SDK should report `instruction_addr_adjustment: "all"`

It would be nice to have one example of a Windows event / binary as well to take a look at the reported address in relation to the disassembly. Then we know which adjustment mode the SDK should report.

PR #44, opened 2022-12-09
feat(rfc): Heartbeat monitoring

priscilawebdev:
Through “heartbeat” monitoring, we want a way for the SDKs to communicate with Sentry so that they are at the bare minimum connected.

This will also be expanded in the future to include richer diagnostic information about SDK configuration, such as the configuration of products 'error monitoring', 'performance monitoring', 'profiling', 'replays' and potentially more.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/3cc099f287506641dbacb4770e364dfaa6abbcab/text/0044-heartbeat.md)


priscilawebdev:
@smeubank .... @iker-barriocanal asked the following question:

> also, if we are trying to solve “the sdk can send events to sentry”, why not a regular event, triggered by the sdk, with some special field in the context? why do we need a new type of event?

Jesse-Box (approved):

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+# Motivation
+
+As part of the [SDK onboarding improvements initiative](https://www.notion.so/sentry/SDK-Onboarding-Improvements-261a3d1deed94522bcff1361fc8bd756) (employees only), user tests have recently been done, where we collect data about the user experience when trying to integrate Sentry for the first time with their applications.
+
+Several discoveries were made, but one in particular that caught our attention is that many users, after following all the installation steps and sending the first error, were confused and frustrated if they did not see it in the Sentry's application.
+
+After refreshing the Sentry's application page a few times to see if they get an error and not getting it, they were left wondering if they missed something while configuring Sentry, which led them to open the onboarding documentation again.
+
+We also noticed that when they successfully sent an issue to Sentry, they were confused as they were taken to the issue stream page without receiving any feedback in the UI, which was initially awaiting for an event.
```
iker-barriocanal, line 18: > they were taken to the issue stream page without receiving any feedback in the UI

What feedback are we looking for specifically? I assume we are looking for some specific feedback, besides the issue stream page changing from "no new issues" to a list with an issue from an event a user just sent.

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
+- Start Date: 2022-12-09
+- RFC Type: feature
+- RFC PR: N/A
+- RFC Status: draft
+
+# Summary
+
+Introduce an infrastructure monitoring agent in all SDKs supported by us. This agent will be triggered every time the user launches the application and will send us information about the state of their integration with Sentry.
```
iker-barriocanal, line 8: > Introduce an infrastructure monitoring agent

I think this sentence is making a decision on what the RFC is about, instead of focusing on the problem.

iker-barriocanal, line 8: > send us information about the state of their integration

Does "the state of their integration" refer to the configuration of the SDK? What information are we looking for specifically?

iker-barriocanal (commented):
Before getting into options, we first should consider what information we need from this event and how requirements will evolve over time to expand it. From reading this document, I don't see any benefit in introducing a heartbeat over sending a simple (log) message.

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+Sentry when collecting this information will be able to give feedback to the user through the user interface, such as that he can now send an error or perform other actions.
+
+This is [one of the designs](https://www.figma.com/file/4EkecQGYtbpOY1G6oEir4X/Exploration%3A-User-Journey?node-id=215%3A1147&t=fGJbtdmq2tDAI8iE-0) we have in mind for the UI (employees only).
+
+In the future, we can also use this mechanism to send other information to Sentry, such as whether the user has configured performance, profiling, etc.
+
+## Session
+
+We're still not sure about this option and are planning to work on a Spike, but the idea is that we'll be able to use the user's session to find out if Sentry's integration with an SDK is working properly.
```
jjbayer, line 50: This approach makes sense to me, because sessions already represent what we are looking for. It would give us the heartbeat infrastructure for free. Not sure if all SDKs send session updates unconditionally though, and whether the latency would be acceptable (sessions are aggregated in relay instances along the way, so it might take up to one minute for it to show up in the UI).

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+Introduce an infrastructure monitoring agent in all SDKs supported by us. This agent will be triggered every time the user launches the application and will send us information about the state of their integration with Sentry.
+
+# Motivation
+
+As part of the [SDK onboarding improvements initiative](https://www.notion.so/sentry/SDK-Onboarding-Improvements-261a3d1deed94522bcff1361fc8bd756) (employees only), user tests have recently been done, where we collect data about the user experience when trying to integrate Sentry for the first time with their applications.
+
+Several discoveries were made, but one in particular that caught our attention is that many users, after following all the installation steps and sending the first error, were confused and frustrated if they did not see it in the Sentry's application.
+
+After refreshing the Sentry's application page a few times to see if they get an error and not getting it, they were left wondering if they missed something while configuring Sentry, which led them to open the onboarding documentation again.
```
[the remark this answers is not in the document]
priscilawebdev: refers to an event that did not hit Sentry. It may happen that instrumentation with Sentry has not been configured correctly

text/0044-heartbeat.md
priscilawebdev: Do you have any suggestion on how we could better describe it?

priscilawebdev: state means:

On -> The connection of the user's SDK with Sentry was established
Off -> There is no connection between the user's SDK and Sentry

text/0044-heartbeat.md
priscilawebdev: Users could get confused if they don't get immediate feedback. Unless we inform them about it in the UI

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Heartbeat/Pulse endpoint
+
+Create an endpoint called heartbeat or pulse that would be used via a request whenever the user successfully launches their app.
+
+The request body would initially be quite simple, containing only the following:
+
+`{type: 'heartbeat'}`
+
+Sentry when collecting this information will be able to give feedback to the user through the user interface, such as that he can now send an error or perform other actions.
```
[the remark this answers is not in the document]
priscilawebdev: We saw during the user testing sessions that we did, that after the user went through all the onboarding steps and sent an error, he would get frustrated if the error wasn't coming, making him wonder if he missed something during the configuration, for instance, if the DSN he used was correct. 

We just want to give the user the feedback that a connection was successfully established so he can discard, for instance the DSN,  from his troubleshooting list

text/0044-heartbeat.md
priscilawebdev: is this also the case of the "envelop"?

text/0044-heartbeat.md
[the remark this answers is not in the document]
olksdr: I get this, but we also can have here same problem. 

If we don't, in the implementation, introduce some kind of QoS for these heartbeats, they could be stuck somewhere in pipeline, can be delayed and won't give expected (fast) feedback. 

I'm talking about e.g: 
* network latency (which could be difficult to account for in some cases)
* or maybe the ingesting/processing queues are full with other types of messages and heartbeats won't be forwarded just right away

I'm not sure if this, QoS-related discussion, should be the part this RFC or later the part of the implementation. But I wanted just to mention it here so it won't get lost.

text/0044-heartbeat.md
danielkhan, line 8: I would not call this "infrastructure monitpring agent". This has a different meaning in the industry. What we want is a heartbeat that carries diagnostic data.

text/0044-heartbeat.md
jjbayer: As discussed in-person: we would probably always build this on top of our metrics infrastructure to make it scalable. So we would have the same delays as with sessions if we build something new.

To me it seems that the questions "Is this SDK alive?" and "Did this SDK have a session?" are equivalent, so we might as well build this feature with sessions.

RaduW:
# Summary:
Question:  
Whose infrastructure are we monitoring with this? 
If the application is a mobile phone we are effectively monitoring the up statuses of the apps running the user's mobile phone. Similarly if the application is a web app, we are monitoring the status of the app running on a user's browser ?

Not clear to me what's the purpose of this, for server side projects I see how that would be useful for client side projects ( mobile or web apps) not so much.

# Motivation:
The motivation seems to be to make life easier for users (that is our users) when they first setup a project. A heart-beat, if introduced, will continue working past this initial 5 minute user confusion.

The problem seems to be that the user does a `Sentry.init(...)`  (as instructed by the script in the project configuration wizard) and then it is confused that nothing has happed. So basically copies and pastes the instruction and nothing seems to happen.

The simplest fix would be to change the snippet displayed and add after `Sentry.init(...) ` another line `Sentry.captureMessage("Hello, this is my first message")` and then the user will see the message.

If this is too hard (I don't think it is), we can probably add a flag to the `Sentry.init()`  called `sendMessageOnInit:true` that would do that automatically ( I think that would be a terrible idea).

# Hartbeat/Pulse endpoint:
Not clear (at least to me) where are we proposing to add this endpoint (Relay or Sentry) ?
If Relay it seams to me to be quite a lot of work that it is not justified in the motivation. In the motivation we are exclusively talking about the intial setup. In the summary we are talking about monitoring infrastructure (without exactly specifing whose infrastructure). Are we going to introduce another type of item in the envelope, how is this going to propagate through Relay and into sentry ( new kafka topic ? ...)

If we are talking about adding an endpoint directly to Sentry we are efectivelly breaking what we did in the last few years by isolating Sentry app from the ingestion pipepline with Relay. Currently Sentry does not ingest anything through an HTTP endpoint, Relay pushes the data into a kafka queue, so adding  the endpoint to Sentry doesn't look to me like something we would consider.    

# Conclusion:

Firstly the Summary and Motivation seem to talk about two very different issues, what are we addressing here?
If we address both let's have them in the motivation... we need A and B and with this we kill 2 birds with one stone. 

The document does not address some basic high level questions (like where are we proposing to add the API), it may be clear to the author but not everybody's assumptions are the same and it makes a huge difference. 

To me the complexity of the proposal does not justify the stated goal (motivation), it is simply a matter of better documentation . By the way I found it frustrating to find the documentation for
`Sentry.captureMessage()` and I sympathise with new users on this but this could be achieved by simply saying ... this is how you set up Sentry `Sentry.intit(...)` ... now let's see if you set it up correctly `Sentry.captureMessage("I'm the first message")`. If you don't receive this in a few seconds then your `Sentry.init( ...)` has a problem.

text/0044-heartbeat.md
priscilawebdev: will update this...thanks for the feedback

priscilawebdev, line 8:
```suggestion
Introduce a heartbeat in all SDKs supported by us. The heartbeat will carry diagnostic data and will be triggered every time the user launches the application.
```

iker-barriocanal:
@RaduW that's pretty good feedback and I agree with the questions about building it. I have a follow-up question on

> If this is too hard (I don't think it is), we can probably add a flag to the Sentry.init() called `sendDebugTestMessageOnInit:true` that would do that automatically ( I think that would be a terrible idea).

Why do you think this is a terrible idea? I think this is better than sending the session every time the SDK is initialized. I recognize the irony of adding a new param to the `init`, considering that the user could just send the event as you described.

text/0044-heartbeat.md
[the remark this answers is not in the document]
smeubank: yes I think a session for example would also be a good fit

we want to know that the SDK is connected with right DSN and the init is fundamentally correct

>  they were left wondering if they missed something

the UI reacting to session received at init for example would solve this. "we see you're connected now waiting for errors and txns" for example

text/0044-heartbeat.md
smeubank: entirety of instrumentation being correct would be difficult to state

but we could simply know the init is correct for now, and what the future of this would be is adding more information about what is configured in the init. What options have they set for example, ideally even options which have default values

the product could then know what _should_ be working, so these initial steps with a simple event are the POC stage


next step the product has more information prior to ever requiring an error or transaction about what all that project should do, and what the user already did so we don't waste their time

text/0044-heartbeat.md
smeubank: the problem statement is something like this I think:

> Today Sentry is relies on receiving errors and transactions from SDKs to take users into the rest of their onboarding experience. This reliance seems to put a weighted importance on pushing the user to send a transaction or error to confirm they have done things correctly. Pushing them through an early process which is an opportunity loss for Sentry and the user to appreciate what they can do with the product.

I'd say that is more or less the problem. And the solution proposal or direction is something like we want SDKs to be able to send a functionally useful event/message which is not dependant on an error or transaction trigger from a specific application/service.

smeubank: @iker-barriocanal for now it will be rather simple - boolean initialized or not basically

but in the future that could scale and be dependent on platform. Thinking of scaling serverside services and their "alive" status or time alive. Versus client side where that type feature would not be so interesting. Every browser opened for example. But we are getting past scope of the initial phase, But I think you were correct above it is important to consider when determining what event makes sense for future planning of the payload.

text/0044-heartbeat.md
smeubank: Agreed. Sessions make a lot of sense to start

> so it might take up to one minute for it to show up in the UI).

I don't think this is so bade TBH. Is it better we push them to send some error and eject them from onbaording for example? 1 minute isn;t that long when someone is presented with new content to digest

smeubank: > Not sure if all SDKs send session updates unconditionally though

which conditions are you referring to?

there are definitely differences across SDKs though whether due to technical need or difference in implementation. Some are auto, some don't send at init (so I heard), others manual. But I think many top prio SDKs aupport auto and at init, which would support this POC work and help justifiy the need to scale should there be proof positive

text/0044-heartbeat.md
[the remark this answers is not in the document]
smeubank: great points

> For me, ability to send the errors / events / envelopes only shows that everything is working

i don't think we want to get to a place where the product can state "everything is configured correctly" that is a bold statement without knowledge of which frameworks and DB integrations they need, then if that SDK supports auto instrumentation for them etc.

But the product can at least start to say "hey you are definitely connected" as a step 1. Further iterations of this would include info about the options which have been configured in the init and set by default, sentry would know a bit more.all without the requirement of pushing them to send an error or txn

> QoS
yes I think we can say this out of scope for the time being :D

text/0044-heartbeat.md
jjbayer: > which conditions are you referring to?

I wasn't sure if all SDKs send sessions by default, and whether they start sending sessions right after startup. So basically what you said.

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+This is [one of the designs](https://www.figma.com/file/4EkecQGYtbpOY1G6oEir4X/Exploration%3A-User-Journey?node-id=215%3A1147&t=fGJbtdmq2tDAI8iE-0) we have in mind for the UI (employees only).
+
+In the future, we can also use this mechanism to send other information to Sentry, such as whether the user has configured performance, profiling, etc.
```
jjbayer, line 46: I think this sentence might decide what we're gonna build eventually. For the simple "one message after startup without data", I think sessions make sense. If we want diagnostics data, and maybe even send these data periodically, then we should build a separate mechanism.

text/0044-heartbeat.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Session
+
+We're still not sure about this option and are planning to work on a Spike, but the idea is that we'll be able to use the user's session to find out if Sentry's integration with an SDK is working properly.
+
+# Drawbacks
+
+We would like to know through this RFC if there are any drawbacks that we should consider
+
+# Unresolved questions
```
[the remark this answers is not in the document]
danielkhan: They are onboarding. If they are using an old SDK for that, something went wrong in the first place already.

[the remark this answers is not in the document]
lforst: I can think of cases where you might run into that case:
- Needing to use an old SDK version because of compatibility issues.
- Migrating orgs/projects.

RaduW:
@iker-barriocanal 
> Why do you think this is a terrible idea? I think this is better than sending the session every time the SDK is initialized. I recognize the irony of adding a new param to the `init`, considering that the user could just send the event as you described.

I think it is a bad idea because it is absolutely unwarranted for our use case. Sure it is easier to add a flag in the `Sentry.init()` function than adding another message type (for example) but my point was that the user confusion could be easily sorted out with a bit of documentation. 

The point I was trying to make is that IMO we don't really need to go past better documentation & a more complete user snippet that does send a test message to resolve the problem stated in the motivation.

RaduW:
@smeubank 
> But what we are also looking to do is lay foundations for features that provide Sentry with more information across releases, well past the initial project onboarding. Today Sentry is actually pretty naive to what a users project/application/service. Then Sentry can start being a little bit more clever about what a user needs to do.

Understood I would say that we should first figure out what we want to achieve before we can start thinking about adding features. So if we have a use case for something that cannot be covered by our current messages then we can think about what else do we need to do/add. 

My point was that IMO the current use case does not warrant/require additional backend work.

text/0044-heartbeat.md
priscilawebdev: yes, for now we are mainly focusing on JavaScript Browser SDKs and according to the [doc](https://docs.google.com/spreadsheets/d/1TzLYBTzFFKnPaR4IqGt-u6ypYU2iaxIGQG--4mdi618/edit#gid=337502117) all of them send sessions automatically

jan-auer:
Thanks @priscilawebdev, this summary makes sense - we can go ahead with this. Please update the RFC to include this information and move all the other considered options into an appendix. 

An additional note to "Sentry connection monitoring": As this is written now, it suggests the same requirements and solution as the first point "Initial Project Setup". Note that neither sessions nor client reports would allow us to track specific installations over time. For example, we cannot track with this whether a specific server instance is no longer able to connect. From reading this, I believe this is not needed. If it is, I suggest to exclude it from the scope of this RFC and list it as a future extension.

priscilawebdev:
@jan-auer I updated the RFC 😉 can you review and approve if you agree with the new updates? I would appreciate more approvals before merging and closing this...thanks!

text/0044-heartbeat.md
```diff
@@ -0,0 +1,83 @@
+- Start Date: 2022-12-09
+- RFC Type: feature
+- RFC PR: N/A
+- RFC Status: Approved
+
+# Summary
+
+Through “heartbeat” monitoring, we want a way for the SDKs to communicate with Sentry so that they are at the bare minimum connected.
+
+This will also be expanded in the future to include richer diagnostic information about SDK configuration such as configured products (bug tracking, performance profiling, replay) or features (e.g. releases, customer sample rates, frameworks, integrations ) and potentially more.
```
ale-cota, line 10: Very minor detail, but I would suggest to rephrase the configured products list to "error monitoring, performance monitoring, profiling, replays"

ale-cota (approved):

jan-auer (approved):

PR #45, opened 2022-12-15
fix: Improve copy/paste experience

armenzg:
It helps when copy pasting into JIRA.

### DIFF
--- a/text/0027-manual-disabling-of-flaky-tests.md
+++ b/text/0027-manual-disabling-of-flaky-tests.md
@@ -36,15 +36,15 @@ https://sentry.io/organizations/sentry/dashboard/7997/?project=5350637&project=4
     - subject: `disabled flaky test: <testname>`
     - body: (something like this)
 
+~~~
+The following test is flaky and it has been disabled. Please fix it and re-enable it.
+
+Here is the Sentry issue: <<< link here >>>
+and the test output:
 ```
-The following test is flaky so it is being disabled: <name of tests>
-
-Here is the sentry issue: <<< link here >>>
-
 <<< test failure output here >>>
-
-Please fix this test and re-enable it.
 ```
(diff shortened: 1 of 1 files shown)


text/0027-manual-disabling-of-flaky-tests.md
```diff
@@ -28,25 +28,23 @@ https://sentry.io/organizations/sentry/dashboard/7997/?project=5350637&project=4
...
-          the following test is flaky so it is being disabled:
+```
+The following test is flaky so it is being disabled: <name of tests>
 
-          here is the sentry issue: <<< link here >>>
+Here is the sentry issue: <<< link here >>>
 
-          ```
-          <<< test failure output here >>>
-          ```
```
asottile-sentry, line 46: I think a different trick that can be used here is to alternate backticks and ~ for the code fences -- I believe github supports both so it'd use the ~ on the outer fence and backticks on the inner fence?

text/0027-manual-disabling-of-flaky-tests.md
```diff
@@ -28,25 +28,23 @@ https://sentry.io/organizations/sentry/dashboard/7997/?project=5350637&project=4
 **a test should be disabled if it has flaked on the primary branch**
 
 - identify the owner of the test in question
-    - often `.github/CODEOWNERS` can help
+    - often [.github/CODEOWNERS](https://github.com/getsentry/sentry/blob/master/.github/CODEOWNERS) can help
```
asottile-sentry, line 31: this procedure isn't specific to `sentry` (for example `getsentry` also)

armenzg:
I've addressed your last comment.

asottile-sentry (approved):

armenzg:
Happy New Year!

PR #46, opened 2022-12-16
proposal for time-to-full-display span in ui transactions

stefanosiano:
We want to provide a new span to the automatic ui transactions.
The ttfd (time-to-full-display) span is a way for the user to notify the sdk that a ui has been fully loaded.
That is, after all data is retrieved, either by database calls or rest apis, and set into the ui.
We would create a new span to the ui automatic transactions to measure it, for all screens of the application.
This rfc is mostly about how we should design the new api.

Also please edit the "Rendered RFC" link so one can quickly get to the rendered markdown file.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/ui-transaction-ttfd-span/text/0046-ttfd-automatic-transaction-span.md)

### DIFF
--- a/text/0046-ttfd-automatic-transaction-span.md
+++ b/text/0046-ttfd-automatic-transaction-span.md
@@ -2,53 +2,58 @@
 - RFC Type: feature
 - RFC PR: https://github.com/getsentry/rfcs/pull/46
 - RFC Status: draft
-
-# Summary
-
-We want to provide a new span to the automatic ui transactions.
-The ttfd (time-to-full-display) span is a way for the user to notify the sdk that a ui has been fully loaded.
-That is, after all data is retrieved, either by database calls or rest apis, and set into the ui.
-We would create a new span to the ui automatic transactions to measure it, for all screens of the application.
-This rfc is mostly about how we should design the new api.
-
+  
+# Summary. 
+  
+We want to provide a new span to the automatic ui transactions.  
+The ttfd (time-to-full-display) span is a way for the user to notify the sdk that a ui has been fully loaded.  
+That is, after all data is retrieved, either by database calls or rest apis, and set into the ui.  
+We would create a new span to the ui automatic transactions to measure it, for all screens of the application.  
+This rfc is mostly about how we should design the new api.  
+  
 # Motivation
 
-This is useful especially for mobile developers, but could also be useful for web.
-There is no reliable way to automatically detect when a ui is fully drawn, as the "fully drawn" concept itself is dependant upon the developer.
-
-# Background
-
-We have to add a new api to the sdk to allow the user to notify that a ui was fully drawn.
-
+This is useful especially for mobile developers, but could also be useful for web.  
+There is no reliable way to automatically detect when a ui is fully drawn, as the "fully drawn" concept itself is dependant upon the developer.  
+  
+# Background. 
+  
+We have to add a new api to the sdk to allow the user to notify that a ui was fully drawn.  
+  
 # Options Considered
-
-These options were considered for Android, but the same apply to other sdks, too.
-1) Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the api is called.
-Pros: This resembles the system api `Activity.reportFullyDrawn()`, making it immediate to use.
+  
+These options were considered for Android, but the same apply to other sdks, too.  
+1) Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the api is called.  
+  
+Pros:  
+This resembles the system api `Activity.reportFullyDrawn()`, making it immediate to use.  
 Cons:
- - We need the activity this api is called for, and passing an Activity instance to an api is not ideal.
-E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the api is called.
-At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.
- - We need to add an api to `SentryAndroid`, instead of the `Sentry` class used everywhere else, due to Activity dependency.
- - This is not ideal for single activity apps, as it wouldn't work for fragments
- - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.
-2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created. 
-This api would return the span or a custom object to allow the user to finish it autonomously.
-Pros: We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it
-Cons: 
-- Returning the span would allow the user to perform "dangerous" operations. We could solve this by returning a stripped interface to allow only the `finish()` method, or an entirely custom object.
-- If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span. 
-3) Add a `Sentry.monitorFullyDrawn()` and a `Sentry.reportFullyDrawn(ISpan)` api. We would start the span automatically when an Activity is being created. 
-This api would return a uid used by the other api to stop the span.
-Pros: 
-- We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it
-- We don't return any "dangerous" object to the user
-Cons: 
-- We would add and force the user to use 2 apis
-- If the user doesn't call the second api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.
-
+ - We need the activity this api is called for, and passing an Activity instance to an api is not ideal.  
+E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the api is called.  
+At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.  
+ - We need to add an api to `SentryAndroid`, instead of the `Sentry` class used everywhere else, due to Activity dependency.  
+ - This is not ideal for single activity apps, as it wouldn't work for fragments.  
+ - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:  
+We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.  
+Cons:
+- Returning the span would allow the user to perform "dangerous" operations. We could solve this by returning a stripped interface to allow only the `finish()` method, or an entirely custom object.  
+- If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+3) Add a `Sentry.monitorFullyDrawn()` and a `Sentry.reportFullyDrawn(ISpan)` api. We would start the span automatically when an Activity is being created.  
+This api would return a uid used by the other api to stop the span.  
+  
+Pros:
+- We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.  
+- We don't return any "dangerous" object to the user.  
+Cons:
+- We would add and force the user to use 2 apis.  
+- If the user doesn't call the second api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+  
 # Unresolved questions
-
-- How long should the timeout be? The Facebook app developers consider a "bad start" a ttfd of 2.5 seconds or more, or an unsuccessful start. 
-This is only for the first screen, as the other screens are usually faster.
+  
+- How long should the timeout be? The Facebook app developers consider a "bad start" a ttfd of 2.5 seconds or more, or an unsuccessful start.  
+This is only for the first screen, as the other screens are usually faster.  
 https://android-developers.googleblog.com/2021/11/improving-app-startup-facebook-app.html
--- a/text/0046-ttfd-automatic-transaction-span.md
+++ b/text/0046-ttfd-automatic-transaction-span.md
@@ -36,6 +36,7 @@ These options were considered for Android, but the same apply to other SDKs, too
 
 
 ## 1. SentryAndroid.reportFullyDrawn(Activity) <a name="option-1"></a>
+
 Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the API is called.  
   
 ### Pros
@@ -68,7 +69,7 @@ This API would return the span or a custom object to allow the user to finish it
 
 ## 3. Sentry.monitorFullyDrawn() with UUID <a name="option-3"></a>
 
-Add a `Sentry.monitorFullyDrawn()` and a `Sentry.reportFullyDrawn(UUID)` API. We would start the span automatically when an Activity is being created.  
+Add a `Sentry.monitorFullyDrawn()` and a `Sentry.reportFullyDrawn(UUID)` API. We would start the span automatically when the SDK creates an auto-generated transaction.  
 This API would return a UUID used by the other API to stop the span.  
   
 ### Pros
@@ -98,7 +99,7 @@ Add `monitorFullyDrawn()` and `reportFullyDrawn()` to ISpan. The user gets acces
 ### Cons
 
 - Extra APIs to call.  
-- Keeping a reference of transaction.  
+- Keeping a reference of transaction.
 
 
 ## 5. reportFullyDrawn() on ISpan <a name="option-5"></a>
@@ -112,8 +113,9 @@ Add `reportFullyDrawn()` toISpan. The user gets access to the APM UI transaction
 - User can add more spans via the same API Sentry.span.  
 
 ### Cons
+
 - Extra APIs to call.  
-- Keeping a reference of transaction.
+- Keeping a reference of a transaction.
 - Not knowing when to wait for fully drawn.
 
 
@@ -131,6 +133,7 @@ We would use use a callback from that.
 - Only available from `androidx.activity` library version 1.7, currently in alpha.  
 - This is not ideal for single activity apps, as it wouldn't work for fragments.  
 - Not knowing when to wait for fully drawn.  
+- Only works on Android.
   
 # Unresolved questions
   
--- a/text/0046-ttfd-automatic-transaction-span.md
+++ b/text/0046-ttfd-automatic-transaction-span.md
@@ -134,12 +134,6 @@ If the user calls `reportFullyDrawn` and the option wether to wait for calling `
 
 1. [Cons 1-2 of option 5](#option-5-cons)
 
-### Cons
-
-- Extra APIs to call.  
-- Keeping a reference of a transaction.
-- Not knowing when to wait for fully drawn.
-
 
 ## 7. Hook into Android's `FullyDrawnReporter` <a name="option-7"></a>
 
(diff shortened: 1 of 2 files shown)


text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the api is called.  
+At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.  
+ - We need to add an api to `SentryAndroid`, instead of the `Sentry` class used everywhere else, due to Activity dependency.  
+ - This is not ideal for single activity apps, as it wouldn't work for fragments.  
+ - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:  
+We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.
```
philipphofmann, line 41: What do you mean by `we can flag the span`?

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+ - We need the activity this api is called for, and passing an Activity instance to an api is not ideal.  
+E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the api is called.  
+At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.  
+ - We need to add an api to `SentryAndroid`, instead of the `Sentry` class used everywhere else, due to Activity dependency.  
+ - This is not ideal for single activity apps, as it wouldn't work for fragments.  
+ - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:
```
philipphofmann, line 40: `m`: Another pro is that it doesn't depend on Android and can be used on multiple UI platforms.

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+ - We need to add an api to `SentryAndroid`, instead of the `Sentry` class used everywhere else, due to Activity dependency.  
+ - This is not ideal for single activity apps, as it wouldn't work for fragments.  
+ - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:  
+We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.  
+Cons:
+- Returning the span would allow the user to perform "dangerous" operations. We could solve this by returning a stripped interface to allow only the `finish()` method, or an entirely custom object.
```
philipphofmann, line 43: What are `"dangerous" operations`?

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+ - This is not ideal for single activity apps, as it wouldn't work for fragments.  
+ - If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:  
+We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.  
+Cons:
+- Returning the span would allow the user to perform "dangerous" operations. We could solve this by returning a stripped interface to allow only the `finish()` method, or an entirely custom object.  
+- If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.
```
philipphofmann, line 44: `m`: Another con is that we can't reliably map `Sentry.monitorFullyDrawn()` to the correct APM transaction.

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+2) Add a `Sentry.monitorFullyDrawn()` api. We would start the span automatically when an Activity is being created.  
+This api would return the span or a custom object to allow the user to finish it autonomously.  
+  
+Pros:  
+We can flag the span when the api is called, so that if the user doesn't call the api, we know we can cancel it.  
+Cons:
+- Returning the span would allow the user to perform "dangerous" operations. We could solve this by returning a stripped interface to allow only the `finish()` method, or an entirely custom object.  
+- If the user doesn't call the api, we would have a span that runs forever. We would have to add a timeout to automatically cancel the span.  
+3) Add a `Sentry.monitorFullyDrawn()` and a `Sentry.reportFullyDrawn(ISpan)` api. We would start the span automatically when an Activity is being created.  
+This api would return a uid used by the other api to stop the span.
```
philipphofmann, line 46: `l`:
```suggestion
This api would return a UUID used by the other api to stop the span.  
```

philipphofmann, line 46: `m`: `Sentry.reportFullyDrawn(ISpan)` takes an `ISpan` as a parameter, but here we write about a UUID. That's a bit confusing.

philipphofmann:
Futher options

4) Add `monitorFullyDrawn()` and `reportFullyDrawn()` to `ISpan`.
The user gets access to the APM UI transaction by calling `Sentry.span`, calls `span.monitorFullyDrawn` and `span.monitorFullyDrawn()`.

Pros

1.  Correlate fully drawn to correct APM transaction
2. User can add more spans via the same API `Sentry.span`
3. Knowing when to wait for fully drawn

Cons

1. Extra APIs to call
2. Keeping a reference of transaction

5) Add reportFullyDrawn()` to `ISpan`.
The user gets access to the APM UI transaction by calling `Sentry.span`, and calls `span.monitorFullyDrawn()`.

Pros

1. Same as 1-2 of option 4

Cons

1. Same as 1-2 of option 4
2. Opposite of pro of option 4

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: I mean that we can have a mechanism to store whether the user called the api for the current transaction, so that we know if we can cancel the ttfd span or keep it and let the user finish it when he wants

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: that should be uuid. My fault, fixed it

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: Operations like creating a new span as child of the ttfd span, or set a different status.
I'm not sure if that would be a real problem, though, or just paranoia 😅

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: You're right, even if maybe forcing the user to call it in a specific callback (like onActivityCreated) could work. But that would be another con 😄

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,59 @@
...
+  
+We have to add a new api to the sdk to allow the user to notify that a ui was fully drawn.  
+  
+# Options Considered
+  
+These options were considered for Android, but the same apply to other sdks, too.  
+1) Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the api is called.  
+  
+Pros:  
+This resembles the system api `Activity.reportFullyDrawn()`, making it immediate to use.
```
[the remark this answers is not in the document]
armcknight: I think what you want to say is "making it *obvious* how to use it"

armcknight (commented):
I'm not sure if this is meant to be Android specific, but looking at this through a `sentry-cocoa` lens, I think the API would look like so:

- Add a new option to `SentryOptions` to indicate that TTFD should be measured, and add a new static method `+[Sentry reportFullyDrawn]` to stop the span, with a timeout for if it's never called and some way to call that out to customers in the front end. It could still have child spans attached and otherwise behave as any other span, but the static API means customers don't have to manage the handle to it, which can be complicated to thread through various async APIs they use while initializing their app.
    - This will not include pre-main time. We could add some static API that allows measuring pre-main time and excludes pre-warm pauses. We already have some prewarming logic, but in order to measure the actual pre-main work done by an app, we'd need to instruct users on how to add some code to their main.m for ObjC, or for Swift and SwiftUI apps, how to call the API from a @main-annotated construct or a swift.main file.

stefanosiano:
> * Add a new option to `SentryOptions` to indicate that TTFD should be measured, and add a new static method `+[Sentry reportFullyDrawn]` to stop the span, with a timeout for if it's never called and some way to call that out to customers in the front end

This is basically option 1 of this rfc. But how would you understand the screen the static api was called for?

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,90 @@
...
+  
+We have to add a new API to the SDK to allow the user to notify that the UI was fully drawn.  
+Also, we need the user to specify the span to finish through a parameter, as it cannot be done automatically.  
+E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the API is called.  
+At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.  
+  
+# Options Considered
+  
+These options were considered for Android, but the same apply to other SDKs, too.  
+1) Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the API is called.
```
marandaneto, line 29: Added in [API level 19](https://developer.android.com/guide/topics/manifest/uses-sdk-element#ApiLevels)
So this feature wont work < 19

marandaneto:
There's a new API for `reportFullyDrawn` from AndroidX Core IIRC, I don't remember the name nor the link, but might be worth investigating this, allows hooking into the Activity `reportFullyDrawn` via callbacks.

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,90 @@
+- Start Date: 2022-12-16
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/46
+- RFC Status: draft
+  
+# Summary.
```
marandaneto, line 6: Is that going to be a transaction, a span or a measurement?
How is this going to play along with https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-activity-instrumentation ?
Do we intend to do that for Fragments as well? https://docs.sentry.io/platforms/android/performance/instrumentation/automatic-instrumentation/#androids-fragment-instrumentation

marandaneto:
> There's a new API for `reportFullyDrawn` from AndroidX Core IIRC, I don't remember the name nor the link, but might be worth investigating this, allows hooking into the Activity `reportFullyDrawn` via callbacks.

Here we go https://developer.android.com/reference/androidx/activity/FullyDrawnReporter
Maybe another option would be hooking into this API and calling the `span.finish()` automatically.

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: The idea is to create a new span to attach to the automatic transactions of activities.
We could also later add the same for fragment transactions as well

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: Nope, we would just add a new API to our sdk, without calling system functions. I think it's better to let the user have control over it

stefanosiano:
> > There's a new API for `reportFullyDrawn` from AndroidX Core IIRC, I don't remember the name nor the link, but might be worth investigating this, allows hooking into the Activity `reportFullyDrawn` via callbacks.
> 
> Here we go https://developer.android.com/reference/androidx/activity/FullyDrawnReporter Maybe another option would be hooking into this API and calling the `span.finish()` automatically.

Yeah, I had a look at that, even if that library version is still in alpha. 
Thanks for the catch, I'm going to add it to the options with pros and cons, and I'd appreciate some comments on them.
The biggest con i think is that it seems to be intended to be called only the first launch of the first Activity, as it says `You can safely call this method any time after first launch as well, in which case it will simply be ignored.` https://developer.android.com/reference/android/app/Activity.html#reportFullyDrawn()

text/0046-ttfd-automatic-transaction-span.md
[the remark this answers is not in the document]
stefanosiano: > I think what you want to say is "making it _obvious_ how to use it"

yeah 😄

text/0046-ttfd-automatic-transaction-span.md
marandaneto: Mobile vitals is based on measurements, don't we intend to leverage our mobile vitals with this?
What if the transaction auto finishes before the user called `Sentry.reportFullyDrawn`?

marandaneto:
> > > There's a new API for `reportFullyDrawn` from AndroidX Core IIRC, I don't remember the name nor the link, but might be worth investigating this, allows hooking into the Activity `reportFullyDrawn` via callbacks.
> > 
> > 
> > Here we go [developer.android.com/reference/androidx/activity/FullyDrawnReporter](https://developer.android.com/reference/androidx/activity/FullyDrawnReporter) Maybe another option would be hooking into this API and calling the `span.finish()` automatically.
> 
> Yeah, I had a look at that, even if that library version is still in alpha. Thanks for the catch, I'm going to add it to the options with pros and cons, and I'd appreciate some comments on them. The biggest con i think is that it seems to be intended to be called only the first launch of the first Activity, as it says `You can safely call this method any time after first launch as well, in which case it will simply be ignored.` [developer.android.com/reference/android/app/Activity.html#reportFullyDrawn()](https://developer.android.com/reference/android/app/Activity.html#reportFullyDrawn())

Found more info here https://android-developers.googleblog.com/2022/10/whats-new-in-jetpack-ads-22.html

> Yeah, I had a look at that, even if that library version is still in alpha

Fair point.

> i think is that it seems to be intended to be called only the first launch of the first Activity

Im not sure about that, it depends what they call app start cold/hot/warm state, maybe they consider app start just by bringing the app from the background which is not even the 1st screen anymore, so it'd be fine to be called or it will be ignored if already, hard to tell without reading the source code/more docs.

text/0046-ttfd-automatic-transaction-span.md
```diff
@@ -0,0 +1,101 @@
...
+  
+We have to add a new API to the SDK to allow the user to notify that the UI was fully drawn.  
+Also, we need the user to specify the span to finish through a parameter, as it cannot be done automatically.  
+E.g. Activity A starts -> Activity B starts -> Activity A finishes loading data and the API is called.  
+At this point without the activity it was called on, we wouldn't know which span to finish, because the activity B would be at the top of the stack.  
+  
+# Options Considered
+  
+These options were considered for Android, but the same apply to other SDKs, too.  
+1) Add a `SentryAndroid.reportFullyDrawn(Activity)` static method. We would start the span automatically when an Activity is being created and we would finish it when the API is called.
```
marandaneto, line 29: `reportFullyDrawn` is an Android API, if our feature is called TTFD, I'd suggest naming it to something similar.
I'd also try to add this API to  `SentryAndroid`, also if possible, to not require `Activity`, so it can't be Android less classes.
Why do we need the activity?

text/0046-ttfd-automatic-transaction-span.md
romtsn: I feel like we could have both, right? The reason for spans is that we can detect performance issues server-side (which works based on spans). But having it as a measurement wouldn't hurt either?

text/0046-ttfd-automatic-transaction-span.md
stefanosiano: > Im not sure about that, it depends what they call app start cold/hot/warm state, maybe they consider app start just by bringing the app from the background which is not even the 1st screen anymore, so it'd be fine to be called or it will be ignored if already, hard to tell without reading the source code/more docs.

I tried to call `Activity.reportFullyDrawn()` multiple times in multiple activities, and it gets called once per activity, so it would be fine. There is still the question about single activity apps, tough. However, i updated the rfc.

the name would be `time-to-full-display`. The activity is needed (in some options) because we wouldn't know which activity is the one that was fully drawn, in case two activities are started and the previous one is not destroyed.

text/0046-ttfd-automatic-transaction-span.md
philipphofmann: I think that's not a problem, but a feature.

text/0046-ttfd-automatic-transaction-span.md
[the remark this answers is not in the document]
philipphofmann: Yes please rephrase it to what @armcknight pointed out.

philipphofmann (commented):
We decided on naming the spans for TTFD and TTID, `ui.load.initial_display` and `ui.load.full_display`. I think our API should reflect that naming decision. 

> Report to the system that your app is now fully drawn, for diagnostic and optimization purposes. The system may adjust optimizations to prioritize work that happens before reportFullyDrawn is called, to improve app startup. Misrepresenting the startup window by calling reportFullyDrawn too late or too early may decrease application and startup performance. - https://developer.android.com/reference/android/app/Activity.html#reportFullyDrawn()

When reading the docs above on `reportFullyDrawn` I think you can only use it for the first Activity when the app is fully drawn. So I would agree with @stefanosiano.

philipphofmann:
We agreed on adding an API `Sentry.reportFullDisplayed` (name to be defined), which will try to find the last active screen load transaction and set the time for full-displayed. The active screen load transaction needs to wait for a to-be-defined timeout if the user calls this API. We still need to spec out all the edge cases.

We didn't go with any of the other options because we don't think that users will use the more complicated API. `Sentry.reportFullDisplayed` seems simple enough for users to try. If we find out users are not happy with the API cause it has some problems with multiple screens we are happy cause it means they are using the API.

Furthermore, the SDKs should somehow automatically detect fully drawn, cause we see a higher adoption with that on JavaScript. Users are reluctant to use manual APIs.

We can add an option to enable or disable this behavior.

Participants: @kahest, @AbhiPrasad, @stefanosiano, @philipphofmann

gggritso:
Extremely late feedback, but out of curiosity, why a span? Time-to-full-display sounds like it's a good fit for a measurement. The Performance team has gotten feedback before that spans are not appropriate for these use cases, since they don't represent a specific unit of work. Plus, with measurements, users would be able to search and filter transactions in Discover and so on. Long time-to-full-display spans might show up as suspect spans, would show up in Performance span breakdowns, etc.

stefanosiano:
> why a span?

One of the main reasons was to allow the usage of Time-to-full-display for performance issues, which are built over spans.
Of course, adding a measurement other than a span makes sense, and we could add both

mjq-sentry:
> One of the main reasons was to allow the usage of Time-to-full-display for performance issues, which are built over spans.
Of course, adding a measurement other than a span makes sense, and we could add both

Is there anything preventing you from creating performance issues using the measurement alone? You have access to the entire event when creating an issue. We use the FCP measurement as a factor in the Large Render Blocking Asset detector, for example.

stefanosiano:
Just throwing an update: 
we are tracking the work on adding `time-to-initial-display` and `time-to-full-display` as measurements [here](https://github.com/getsentry/team-mobile/issues/86)

PR #47, opened 2022-12-20
feat(rfc): Introduce profiling context

Zylphrex:
Today, a profile contains a transaction ID to indicate the transaction that occurred within it. We would like to create the inverse relationship from a transaction to a profile to enable queries that were not feasible using only the relationship from profile to a transaction.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/txiao/feat/introduce-profiling-context/text/0047-introduce-profile-context.md)

### DIFF
--- a/text/0047-introduce-profile-context.md
+++ b/text/0047-introduce-profile-context.md
@@ -12,9 +12,9 @@ transactions so we can easily relate a group of transactions back to the respect
 
 We already have a way to relate a group of profiles back to the respective transaction. And up to
 this point, we have been using this relationship to map single transaction at a time to the
-respective profile. This is now insufficient as we would like to be able to relate a group of
-transactions back to the respective profile. This is to enable various touch points into profiling
-in the existing performance product.
+respective profile. This is now insufficient as we would like to be able to relate a set of
+transaction ids back to the respective profile ids. This is to enable various touch points into
+profiling in the existing performance product.
 
 The intended use cases include
 
--- a/text/0047-introduce-profile-context.md
+++ b/text/0047-introduce-profile-context.md
@@ -34,7 +34,7 @@ Performance to Profiling and vice versa.
 
 ## SDK
 
-### Option A: Add a new context type
+### Add a new context type
 
 Introduce a key in the transaction payload under contexts called `profile` to include information
 related to the profile.
@@ -45,22 +45,6 @@ related to the profile.
   "contexts": {
     ...,
     "profile": {
-      "profile_id": "<id>"
-    }
-  }
-}
-```
-
-### Option B: Use the existing trace context
-
-Use the existing trace context and add a new key for the `profile_id`.
-
-```json
-{
-  ...,
-  "contexts": {
-    ...,
-    "trace": {
       "profile_id": "<id>"
     }
   }
--- a/text/0047-introduce-profile-context.md
+++ b/text/0047-introduce-profile-context.md
@@ -1,0 +1,110 @@
+- Start Date: 2022-12-20
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/47
+- RFC Status: draft
+
+# Summary
+
+We want to introduce the concept of a profiling context, much like a tracing context, starting with
+transactions so we can easily relate a group of transactions back to the respective profile.
+
+# Motivation
+
+We already have a way to relate a group of profiles back to the respective transaction. And up to
+this point, we have been using this relationship to map single transaction at a time to the
+respective profile. This is now insufficient as we would like to be able to relate a group of
+transactions back to the respective profile. This is to enable various touch points into profiling
+in the existing performance product.
+
+The intended use cases include
+
+- fetching a list of transaction with the profile ID
+- filtering transactions to only those that have a profile
+
+# Background
+
+Profiles will be closely associated to transactions and creating this relation will enable us to
+build a richer integration between the Performance and Profiling products.
+
+Since the Profiling product will be initially tied to the Performance product, we want to ensure
+that there is a smooth integration through various touch points to allow users to navigate from
+Performance to Profiling and vice versa.
+
+# Options Considered
+
+## SDK
+
+### Option A: Add a new context type
+
+Introduce a key in the transaction payload under contexts called `profile` to include information
+related to the profile.
+
+```json
+{
+  ...,
+  "contexts": {
+    ...,
+    "profile": {
+      "profile_id": "<id>"
+    }
+  }
+}
+```
+
+### Option B: Use the existing trace context
+
+Use the existing trace context and add a new key for the `profile_id`.
+
+```json
+{
+  ...,
+  "contexts": {
+    ...,
+    "trace": {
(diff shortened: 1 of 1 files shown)


text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,110 @@
+- Start Date: 2022-12-20
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/47
+- RFC Status: draft
+
+# Summary
+
+We want to introduce the concept of a profiling context, much like a tracing context, starting with
+transactions so we can easily relate a group of transactions back to the respective profile.
```
philipphofmann, line 9: On the Cocoa SDK, @armcknight opened a draft PR to change the relationship between transactions and profiles because of problems with DS. Currently, a profile can relate to a group of transactions. The [draft PR](https://github.com/getsentry/sentry-cocoa/pull/2545) proposes to change this so that a profile always relates to one transaction only. If there are multiple transactions in parallel, each one gets its own profiling data, which will be duplicated for the time they overlap. @Zylphrex, I think it's best to sync with @armcknight before continuing with this RFC.

Zylphrex: Yes, the change @armcknight is making is expected and will be the way all SDKs handle parallel transactions (by duplicating the profile data).

The end result of this is expected to be 
- each transaction will have the corresponding profiling data from the same time window
- the transaction and the profile should be sent in the same envelope (to deal with the DS problem)
- if there are parallel transactions, we will duplicate the profiling data for each transaction (by trimming the profile for the transaction's time window)

The *new* change from this RFC is that the transaction will contain a profile context that contains the information for the corresponding profile so that it can be used to relate a transaction back to the profile easily.

Zylphrex: That is correct. There will be a 1:1 relationship between transactions and profiles. And this context will power the relation from transactions to profiles.

text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+## Storage
+
+To support querying for this relationship, we have to store the profile ID, extracted from the
+profile context set by the SDK, on the existing transactions dataset. The types of queries we want
+to do are as follows:
+
+1. Selecting the profile ID
+2. Selecting rows that have a profile ID since not all transactions will have a profile
```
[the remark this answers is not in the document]
Zylphrex: So one common workflow that has come up is a user is looking at transactions in the performance product. Then they want to see a profile associated with it.

But because profiles can be sampled at a lower rate than transactions, not all transactions will have a profile associated with it. So we would like to provide a way for users to focus on transactions that **HAVE** a profile associated.

One *possible* flow for this is perhaps the user will type `has:profile_id` into the search bar to filter for just transactions that are known to have an associated profile.

text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,110 @@
...
+
+1. Have to add a new column to the transactions dataset
+
+## Option B:
+
+1. Is there bloom filter index on contexts? Will it be performant enough?
+
+# Unresolved questions
+
+- Which option should we use for storing the profile ID on the transaction?
```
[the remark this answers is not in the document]
Zylphrex: That sounds good, then we'll likely lean towards adding a new profile id column on the transactions for this.

text/0047-introduce-profile-context.md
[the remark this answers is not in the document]
nikhars: Thanks for the explanation. In the workflow you mentioned, since the user is interested in **any** transaction with a profile, I would recommend that you limit your search to a fixed number of transactions. There isn't really a need to search for all transactions within a time range for your use case.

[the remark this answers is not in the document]
Zylphrex: Yep understood.

text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,110 @@
...
+# Summary
+
+We want to introduce the concept of a profiling context, much like a tracing context, starting with
+transactions so we can easily relate a group of transactions back to the respective profile.
+
+# Motivation
+
+We already have a way to relate a group of profiles back to the respective transaction. And up to
+this point, we have been using this relationship to map single transaction at a time to the
+respective profile. This is now insufficient as we would like to be able to relate a group of
```
jjbayer, line 15: Given the thread above about the 1:1 relationship between transactions and profiles, I'm confused by the mention of "a group of transactions" here. Is the relationship 1:n in theory but 1:1 in practice?

text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,110 @@
...
+related to the profile.
+
+```json
+{
+  ...,
+  "contexts": {
+    ...,
+    "profile": {
+      "profile_id": "<id>"
+    }
```
jjbayer, line 49: Apart from `profile_id`, do we expect any other fields in this context in the future? If not, I would prefer option B or even a top-level `profile_id` field.

text/0047-introduce-profile-context.md
Zylphrex: Ahh sorry, I'll update the wording but the intention here is that we want to be map a set of N transaction ids to a set of N profile ids

text/0047-introduce-profile-context.md
Zylphrex: We don't have any plans right now to add more fields to this context in the right now but that may change as we move towards a continuous profiling approach.

text/0047-introduce-profile-context.md
```diff
@@ -0,0 +1,94 @@
...
+
+# Drawbacks
+
+## Option A:
+
+1. Have to add a new column to the transactions dataset
+
+## Option B:
+
+1. Is there bloom filter index on contexts? Will it be performant enough?
```
fpacifici, line 89: There is no bloom filter index on contexts and we do not plan to add it. 
Queries on contexts tend to be the heaviest queries we have because contexts can be really large.

jjbayer (approved):
Not reviewing the storage part, but the addition of a context to the event protocol looks good to me. Please also open a PR to add the new schema to Relay, similar to https://github.com/getsentry/relay/pull/1617.

AbhiPrasad:
> Please also open a PR to add the new schema to Relay

Aside from this we'll also need to update the develop docs: https://github.com/getsentry/develop/blob/master/src/docs/sdk/event-payloads/contexts.mdx

AbhiPrasad (approved):
If the profiles end up on the scope, then theoretically any event can just look on the scope, grab the profile ID and generate a new profile context (perhaps we want to quickly add this context to the RFC). 

✅ on Option A for me on the SDK side.

sl0thentr0py (approved):

text/0047-introduce-profile-context.md
Zylphrex: Understood. In that case, I think we'll go with option A and add a new profile id column as @nikhars mentioned. Does that make sense?

evanh (approved):

PR #48, opened 2023-01-05
Add replay_id to context, change storage in snuba

JoshFerge:
Move replay_id out of tags in event schema, and likely in clickhouse storage. very similar to https://github.com/getsentry/rfcs/pull/47

[Rendered RFC](https://github.com/getsentry/rfcs/blob/jferg/elevate-replayid/text/0048-move-replayid-out-of-tags.md)


text/0048-move-replayid-out-of-tags.md
```diff
@@ -0,0 +1,63 @@
...
+    "replay": {
+      "replay_id": "<id>"
+    }
+  }
+}
+```
+
+### Other Options
+
+I suppose we could add the "replay_id" top level to contexts. Not sure what other options there could be.
```
bruno-garcia, line 40: The protocol defines that `contexts` only take an object, so this isn't an option I don't think

Lms24:
From an SDK perspective, I think this change makes sense and I'd prefer SDK option A. Sounds more future-proof to me and it's more consistent with how we store other entries in `event.contexts`.

bruno-garcia (approved):
From the PoV of SDK it LGTM.
Storage I imagine Option A or B would do.

text/0048-move-replayid-out-of-tags.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Summary
+
+This RFC is almost exactly the same as https://github.com/getsentry/rfcs/pull/47.
+
+Right now replay_ids are added to the error/transaction event tags object in the [event schema](https://sentry.io/organizations/sentry/performance/summary/replays/?project=11276&query=&statsPeriod=24h&transaction=%2Fsettings%2Faccount%2Femails%2F). This isn't ideal as tags should be reserved for user use only.
+
+We'd like to create a context object `replay` which for now would contain the singular field `replay_id`.
+
+We also want to move this potentially out of the tags columns within clickhouse for errors/txs. The replay_id can be stored as its own column or as a key in contexts.
```
evanh, line 14: This will happen automatically once you stop writing the tag value I believe.

text/0048-move-replayid-out-of-tags.md
```diff
@@ -0,0 +1,63 @@
...
+    }
+  }
+}
+```
+
+### Other Options
+
+I suppose we could add the "replay_id" top level to contexts. Not sure what other options there could be.
+
+## Storage
```
evanh, line 42: The difference between A and B (I don't think C is a good idea) will depend on how you plan to query on this column. Can you give some examples of how the column will be used?

JoshFerge: common use cases:

1. from a list of issue_ids, select and return the event_id and replay_id if a replay_id exists on the event
2. from a list of transaction_names, select and return the event_id and replay_id if a replay_id exists on the event
3. from a replay_id, fetch all associated issues
4. from a replay_id, fetch associated transactions 
5. select the column as a default field in the performance page, and on discover, for errors/txs

evanh: My take on this would be to suggest option B then, based on queries 3 and 4. If you are filtering by replay ID having a dedicated column would be best. It's also what we recommended to the profiling team. @nikhars do you have any other thoughts?

nikhars: The `replay_id` column seems to be a loose linking column between two different datasets. I would also recommend a dedicated column based on its usage. So I would agree with Evan.

nikhars (approved):

text/0048-move-replayid-out-of-tags.md
JoshFerge: got it, thanks!

PR #49, opened 2023-01-10
Add filename spec and Android visibility example to View Hierarchy

markushi:
During development on iOS/Flutter/Android we agreed on setting the attachment filename to `view-hierarchy.json`.
On top of that I added an example how the visibility attribute for Android looks like.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/fix/view-hierarchy-add-missing-specs/text/0033-view-hierarchy.md)

### DIFF
--- a/text/0033-view-hierarchy.md
+++ b/text/0033-view-hierarchy.md
@@ -20,7 +20,7 @@ Example View Hierarchies from PoC implementations:
 
 # Proposal
 
-Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.view_hierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
+Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The filename is `view-hierarchy.json`, the `attachment_type` is set to `"event.view_hierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
 
 The attachment payload is a JSON structure containing the view hierarchy, here's an example:
 ```json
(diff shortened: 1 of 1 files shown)


text/0033-view-hierarchy.md
```diff
@@ -38,6 +38,7 @@ The attachment payload is a JSON structure containing the view hierarchy, here's
             "y": 0.0,
             "z": 2.0, // if applies.
             "visible": true|false,
+            "visibility": "visible|invisible|gone", // Android specific visibility property
```
marandaneto, line 41: Shouldn't this be part of the `{extra_properties}`? since this is not applicable to every platform.

marandaneto (approved):

text/0033-view-hierarchy.md
```diff
@@ -20,7 +20,7 @@ Example View Hierarchies from PoC implementations:
 
 # Proposal
 
-Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The `attachment_type` is set to `"event.view_hierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
+Capture the view hierarchy in our front-end SDKs and convert it to a common JSON representation. Add it as an [attachment](https://develop.sentry.dev/sdk/envelopes/#attachment) to the envelope. The filename is `view-hierarchy.json`, the `attachment_type` is set to `"event.view_hierarchy"` and the `content_type` is set to the `"application/json"` mime-type.
```
narsaynorath, line 23: Do we currently expect to have to support multiple view hierarchies, similar to how we may have multiple screenshots? In that case I wonder if we should call out that the filename will be something like `view-hierarchy-{i}.json`

narsaynorath (approved):

text/0033-view-hierarchy.md
markushi: Good point! I don't think it's required right now as we already have a top level `windows` property inside the `view-hierarchy.json` which should cover your use-case.

text/0033-view-hierarchy.md
markushi: I think the idea back then was to have a single flat list for all properties since all of them are optional, can you remember @brustolin?

PR #60, opened 2023-01-13
rfc(decision): Linking Backend Errors With Replays

JoshFerge:
See rendered RFC. 

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/linking-backend-errors-with-replays/text/0060-linking-backend-errors-with-replays.md)


mitsuhiko:
@ale-cota @smeubank given the conversions about session based sampling, I think you two should probably have a look at this.

mydea:
Would we also adjust the frontend-error linking to work the same in this case? I think this would be valuable as well, as the current SDK-site linking is a bit brittle as well.

JoshFerge:
> Would we also adjust the frontend-error linking to work the same in this case? I think this would be valuable as well, as the current SDK-site linking is a bit brittle as well.

What did you have in mind for this // how is it brittle?

mydea:
> > Would we also adjust the frontend-error linking to work the same in this case? I think this would be valuable as well, as the current SDK-site linking is a bit brittle as well.
> 
> What did you have in mind for this // how is it brittle?

Basically we link errors locally, but this does not account for errors that did not reach Sentry, for whatever reason. So we may still send errorIds that actually do not exist on sentry.io.

JoshFerge:
> > > Would we also adjust the frontend-error linking to work the same in this case? I think this would be valuable as well, as the current SDK-site linking is a bit brittle as well.
> > 
> > 
> > What did you have in mind for this // how is it brittle?
> 
> Basically we link errors locally, but this does not account for errors that did not reach Sentry, for whatever reason. So we may still send errorIds that actually do not exist on sentry.io.

Ah, right! Yes, that would be possible. I'll write some more details on that but we'd get that with our `post_process` changes. Related: https://github.com/getsentry/replay-backend/issues/247

text/0060-linking-backend-errors-with-replays.md
```diff
@@ -0,0 +1,74 @@
...
+
+We'd add a new baggage http header `sentry-replay_id`. This value would automatically make it onto the trace envelope header, and from there we could add it onto the event's payload if it exists.
+
+- in `post_process`, if a replay_id exists on the event (we've previously defined a context for this) emit a message to our snuba processer which would write the `event_id`/`replay_id` to our replays table. [Link to post_process code where we'd emit kafka message](https://github.com/getsentry/sentry/blob/b1c6aa7b1a4ca0bfa2f402df61bf5d23b169e7ed/src/sentry/tasks/post_process.py#L452)
+
+### Option B: sentry-trace Header
+
+[Link to sentry-trace header information](https://develop.sentry.dev/sdk/performance/#header-sentry-trace)
+
+- The current sentry-trace header format is `sentry-trace = traceid-spanid-sampled`. If we want to maintain compatability with traceparent/zipkin-b3, we'd extend this to add the flags field. he format would be `traceid-spanid-sampled-flags`, of which flags can contain 8 bits. We'd add a "has_replay" bit to the flags field.
```
billyvg, line 32:
```suggestion
- The current sentry-trace header format is `sentry-trace = traceid-spanid-sampled`. If we want to maintain compatibility with traceparent/zipkin-b3, we'd extend this to add the flags field. The format would be `traceid-spanid-sampled-flags`, of which flags can contain 8 bits. We'd add a "has_replay" bit to the flags field.
```

text/0060-linking-backend-errors-with-replays.md
```diff
@@ -0,0 +1,74 @@
...
+- This is unlikely to be done, but there has been talk before of creating a "graph" type datastore of events, where you could pass a list of events/traces/issues and quickly retrieve back a list of associated events. If we had a store like this, we'd simply pass it a transaction_name/issue_id/list of traces, and get back a list of replay_ids.
+- This store would not help with combined queries like "show me replays greater than 1 min that have a backend error associated", as I assume we could not join it with a clickhouse query easily. It would only be for showing associated replays on a backend issue / transaction / event, and
+
+### Option E:
+
+... There may be other options I haven't considered and am open to more discussion.
+
+# Drawbacks
+
+### Option A
```
lforst, line 56: Something that came up in TSC

```suggestion
# Drawbacks

### Option A

A thing to keep in mind when adding a replay ID to the Dynamic Sampling Context is the immutability of DSC and traces starting in the backend.

Simple example: A trace starts in the backend for a server-side rendered website, DSC is frozen there, if we assume that DSC cannot be unfrozen, the frontend cannot add a replay ID as part of the pageload transaction.

### Option A
```

text/0060-linking-backend-errors-with-replays.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Option B
+
+This in every way feels inferior to option A, as the backend is still quite complicated as there's a race condition.
+
+## Option C
+
+This is very brittle and likely wouldn't scale super well. A naive solution if SDK changes aren't possible.
+
+## Option C
```
ryan953, line 93:
```suggestion
## Option D
```

JoshFerge:
Picking this back up. meeting notes from Client TSC Below:

- Replay-backend connection RFC
    - We currently link to replays and transactions with a Replay ID.  This is impossible for back-end errors.
    - We need a way to go from a back-end error to a front-end replay.
    - Options listed in RFC.  Option A is likely the best one.  Would add the replay id to the baggage header.
    - Neel - might be limits on baggage size.  Steven - that’s a known limitation we have accepted.  Abhijeet - we’re not that close to it, but we could consider using another header instead of baggage.
    - Bruno - There’s an advantage using baggage, in that it’s a well-know header.
    - Would need this implemented in the JavaScript SDK
    - Luca - a trace could start in the back end that leads to a page-load transaction that has a replay (Bruno - yes, such as Server-Side Rendering)
    - Luca - why can’t we do this in the back end?  Josh - We can’t do joins in Clickhouse, but options C and D in the RFC explore various possibilities to deal with this in the backend
    - Followup outside of TSC to continue

JoshFerge:
Note from @benvinegar we should consider the case where a user hasn't adopted tracing.

bruno-garcia:
> Note from @benvinegar we should consider the case where a user hasn't adopted tracing.

We could split this in 2 concerns: Lets get the propagation done.
Separately, we get traces to start and propagate without performance monitoring.

text/0060-linking-backend-errors-with-replays.md
```diff
@@ -0,0 +1,93 @@
...
+> XX does a good job of displaying the session replays. To me the biggest differentiator that Session Replay has is its connectivity to the rest of the Sentry ecosystem.
+> XX is perfect for determining front-end only issues, but doesn't give enough information to find root causes when the issue is deeper in the system. and It would be great if we could link replays to BE errors. This is our current setup. 1) FE Web App; 2) BE Web App.
+
+# Options Considered
+
+## Option A: Dynamic Sampling Context / Baggage Header
+
+[Link To relevant docs about dynamic sampling context / baggage](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#baggage)
+
+We'd add a new baggage http header `sentry-replay_id`. This value would automatically make it onto the trace envelope header, and from there we could add it onto the event's payload if it exists.
```
antonpirker, line 35: Just for clarification. There is only one `baggage` HTTP header and it contains multiple items. A example baggage header looks like this: `"other-vendor-value-1=foo;bar;baz, sentry-trace_id=771a43a4192642f0b136d5159a501700, sentry-public_key=49d0f7386ad645858ae85020e393bef3, sentry-sample_rate=0.01337, sentry-user_id=Am%C3%A9lie, other-vendor-value-2=foo;bar;"`

antonpirker:
My gutt tells me that the `baggage` header is the way to go, because it was designed for this. If we add a new `sentry-something` key into the baggage it should be propagated everywhere and it also should not be break dynamic sampling, because we have a list of what items should go into the dynamic sampling context.

bruno-garcia:
> My gut feeling is that the `baggage` header is the way to go, because it was designed for this. If we add a new `sentry-something` key into the baggage it should be propagated everywhere and it also should not be break dynamic sampling, because we have a list of what items should go into the dynamic sampling context.

Are all backend SDKs already reading `sentry-something` and adding as a tag to the scope? I imagine it's not, so we'll need to add that feature for `replayId`

text/0060-linking-backend-errors-with-replays.md
JoshFerge: right, makes sense. will clarify "we add a new baggage item"

JoshFerge:
> > My gut feeling is that the `baggage` header is the way to go, because it was designed for this. If we add a new `sentry-something` key into the baggage it should be propagated everywhere and it also should not be break dynamic sampling, because we have a list of what items should go into the dynamic sampling context.
> 
> Are all backend SDKs already reading `sentry-something` and adding as a tag to the scope? I imagine it's not, so we'll need to add that feature for `replayId`

this could also be done in relay in theory, but do not know feasibility of implementation

antonpirker:
> Are all backend SDKs already reading `sentry-something` and adding as a tag to the scope? I imagine it's not, so we'll need to add that feature for `replayId`

Oh, I read "tag with replay_id" as that the replay_id just hast to be somewhere in the event, did not realize that it means actually set a tag on the scope. 

No, this tagging is not done by SDKs, so we will need to implement this.

JoshFerge:
> Oh, I read "tag with replay_id" as that the replay_id just hast to be somewhere in the event, did not realize that it means actually set a tag on the scope.

I think its ambiguous whether or not the SDK has to add it to the event, I will dig into implementation options.

JoshFerge:
After doing some prototyping, I've found that each individual backend SDK will have to be modified in order to set the replay_id on the event. this is because the dynamic sampling context is not sent along to ingest on error events.

I also found that at least on the python SDK, [the dynamic sampling values are hardcoded](https://github.com/getsentry/sentry-python/blob/2c8d27788c7e78a2e24e264d0e2d2f221e157658/sentry_sdk/tracing_utils.py#L253), which means that we'll have to update SDKs to add another value.

I'm curious because from reading [the spec](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#baggage-header), it seems like we wanted more flexibility in the propagation.

> Being able to simply copy key-value pairs from the baggage header onto the trace envelope header gives us the flexibility to provide dedicated API methods to propagate additional values using Dynamic Sampling Context. This, in return, allows users to define their own values in the Dynamic Sampling Context so they can sample by those in the Sentry interface.

Given these details, I believe we can take the following course of action to accomplish the initial goal of backend errors being tagged w/ replay_id when tracing is enabled:


- [ ] Modify the sentry-javascript replay SDK to expose the current replayId on the [integration](https://github.com/getsentry/sentry-javascript/blob/766a00b68a5f6f7ad4f1ca80b353a01520392788/packages/replay/src/integration.ts#L22)
- [ ] Modify the [sentry-javascript DSC function](https://github.com/getsentry/sentry-javascript/blob/766a00b68a5f6f7ad4f1ca80b353a01520392788/packages/tracing/src/transaction.ts#L238) to grab the replayId on the replay integration if it exists, and add it to the DSC obj.
- [ ] For each Backend SDK:
  - [ ] let replay_id propagate on the DSC
  - [ ] Set the replay_id on the scope like we do for trace_id currently
  - [ ] When we send an event, enrich it with the scope's replay_id
  - [ ] TODO: prioritize SDK implementations by adoption/importance
- [ ] From there, we shouldn't need any changes on relay
- [ ] Given this needs and SDK upgrade, we'd likely want to put SDK CTA upgrades for this feature in non-intrusive manner in the replays parts of the product.

This would allow users to navigate to replays from backend errors / issues. The extra work I described in option A of the RFC would allow users to search for replays by a specific backend error, but I don't think that's where the majority of the value lies in this integration, so for the initial feature release, I believe the above is the minimum set we'd need.

Given the amount of SDK work and friction in getting users to upgrade their SDKs, I do think it would be worthwhile to prototype what some kind of query engine for linking all of this via trace_id instead (Option D in the RFC). I will plan to spend a couple of days on this to explore feasibility before moving forward with the implementation / delegation of these SDK changes.

JoshFerge:
Update: after having internal discussion, the path forward has been decided on and we'll be adding replay_id to dynamic sampling context, which, eventually, _all_ sdks will send in _all_ envelopes. there will be related SDK work that happens this year which will 

a) send all events using envelopes
b) send the dynamic sampling context baggage headers along with all envelopes
c) allow any baggage header prefixed with `sentry-` to be propagated along.

We will pluck the replay_id out of these headers in relay and add it to the event context. If this final note makes sense to all we can go ahead and approve / merge this RFC

bruno-garcia:
Is it correct to say we're going with Option A? If so can we set this as the decided option and merge this?

JoshFerge:
Yes! Option A it is! decided!

PR #62, opened 2023-01-17
rfc(feature): Controlling PII and Credentials in SDKs

sl0thentr0py:
all my homies hate leaky pipelines

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/controlling-pii-and-credentials-in-sd-ks/text/0062-controlling-pii-and-credentials-in-sd-ks.md)

resolves https://github.com/getsentry/team-webplatform-meta/issues/8

### DIFF
--- a/text/0062-controlling-pii-and-credentials-in-sd-ks.md
+++ b/text/0062-controlling-pii-and-credentials-in-sd-ks.md
@@ -1,0 +1,207 @@
+- Start Date: 2023-01-17
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/62
+- RFC Status: draft
+
+# Summary
+
+Proposals for new SDK-side configs/APIs to exhaustively control both PII (Personally Identifiable Information) and security credentials that are sent to Sentry.
+
+# Motivation
+
+As we get larger orgs with privacy, security and GDPR concerns using Sentry, it becomes more and more important to solve these use cases well with proper abstractions.  
+We've also had incidents in the past where we got collateral flak publicly for breaches / leaks on self-hosted instances.
+
+Controlling data exposed or collected on any SDK in a centralized and consistent manner with full transparency would be the goal of this RFC.
+
+I will also make the case for (optionally) distinguishing between PII and Security semantics.  
+Security credentials should **never** be sent to Sentry, while PII would be left as a choice for the user.
+
+# Background
+
+## Existing PII handling mechanisms
+
+Currently, PII is mainly stripped in Relay along with some ad-hoc logic of what data to include in each (SDK, language, integration) combination.
+
+### Relay
+
+Relay implements a [`PiiProcessor`](https://github.com/getsentry/relay/blob/37f965b9747fd00ba4f3b01be378279bb6704687/relay-general/src/pii/processor.rs#L55) that scrubs
+the event payload recursively for things it can identify including emails, creditcards, IP addresses, SSNs, etc.  
+The configuration format is documented [here](https://develop.sentry.dev/pii/) and a full list of the current builtin rules can be found [here](https://github.com/getsentry/relay/blob/37f965b9747fd00ba4f3b01be378279bb6704687/relay-general/src/pii/builtin.rs#L34).
+
+While we should continue doing server-side scrubbing in any case, having more control on the SDK is still desired by users who want to be completely sure
+that no sensitive data ever leaves their application and crosses the network.
+
+### SDKs
+
+#### `send_default_pii`
+
+The Unified API has a boolean config `send_default_pii` that is used ad-hoc in various integrations to conditionally add certain fields to the event payload.  
+Each (SDK, language, integration) combination internally uses this flag to optionally add sensitive data such as headers, cookies or user login data.  
+Thus, each integration author has to decide what data is potentially sensitive and the user has no further control on what is included.  
+PII handling is also missing in many SDKs in performance related logic such as database span recording integrations.
+
+The following table will illustrate usage examples from various SDKs highlighting various problems with the current way of doing things.
+
+| SDK        | Integration             | Links                                                                                                                                                                                               | Remarks                                                                      |
+| ---        | ---                     | ---                                                                                                                                                                                                 | ---                                                                          |
+| python     | WSGI                    | [cookies](https://github.com/getsentry/sentry-python/blob/de84c4c7d65830e91f3af8555ebb1f7385a2d7cc/sentry_sdk/integrations/_wsgi_common.py#L63-L64)                                                 | all cookies including session are included which is a security issue         |
+| python     | WSGI                    | [headers](https://github.com/getsentry/sentry-python/blob/de84c4c7d65830e91f3af8555ebb1f7385a2d7cc/sentry_sdk/integrations/_wsgi_common.py#L18-L27)                                                 | hardcoded sensitive headers list                                             |
+| python     | sql spans & breadcrumbs | [params](https://github.com/getsentry/sentry-python/blob/de84c4c7d65830e91f3af8555ebb1f7385a2d7cc/sentry_sdk/tracing_utils.py#L169-L179)                                                            | not using `send_default_pii` but custom config                               |
+| javascript | httpclient              | [cookies, headers](https://github.com/getsentry/sentry-javascript/blob/7316b850988df64467a7052082db84461f8a2cd2/packages/integrations/src/httpclient.ts#L136-L154)                                  | no control in which cookies and headers                                      |
+| ruby       | request interface       | [headers](https://github.com/getsentry/sentry-ruby/blob/bbfb6514d95adda77676d4a0823924e789db0909/sentry-ruby/lib/sentry/interfaces/request.rb#L84-L104)                                             | random hardcoded lists                                                       |
+| android    | okhttp                  | [headers](https://github.com/getsentry/sentry-java/blob/9b6c37f33376556ce0f66e526541a04e9e3aa60f/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L223-L242) | no headers at all if `send_default_pii` false, filter sensitive even if true |
+
+As demonstrated by the collection above, each SDK does slightly different things and the decisions on what is gated behind `send_default_pii` are mostly left to the integration authors.  
+The user can merely choose to follow the author's implementation with this flag but not really be sure if and what data is leaking without reading the integration source, which is an undesirable state of affairs.
+
+
+#### `before_send` and `add_event_processor`
+
+In cases where users are not happy with `send_default_pii`, we suggest using either `before_send` for errors or `add_event_processor` for all events for controlling PII on the final event payload.  
+This, while it works in theory has several problems:
+
+* there is an ordering problem when a user's added global event processor runs before some other event processor added by an integration, thus not giving the user the actual final event
+  * there are [documented **unsupported** use cases](https://github.com/getsentry/sentry-python/issues/1226#issuecomment-978806984) here such as the user never being able to remove a certain request header since the headers are actually added [after](https://github.com/getsentry/sentry-python/blob/de84c4c7d65830e91f3af8555ebb1f7385a2d7cc/sentry_sdk/integrations/wsgi.py#L306-L325) the user's event processor runs
+  * this will be partially fixed by `before_send_transaction` but I still believe PII scrubbing is a key concern that should not be shoe-horned into catch-all callbacks that run at the end
+* writing either of these callbacks requires the user to know the whole event schema which even our own SDK devs don't know fully
+
+# Options Considered
+
+The high-level proposal is to introduce a new `config.event_scrubber` with a default implementation that users can tweak with a simple `denylist` in most cases.  
+The end-user only cares about hiding certain information quickly, so a single list is the ideal interface.  
+Anything more complex than that is asking them to think about things they don't care about.  
+The advanced user can still override the implementation (similar to custom transports) if they wish to do so by using their own class implementation.
+
+
+```python
+# ===========================================================================
+# event_scrubber.py
+
+DEFAULT_DENYLIST = ["password", "HTTP_COOKIE"]
+
+class EventScrubber:
+    def __init__(self, denylist=None):
+        self.denylist = DEFAULT_DENYLIST if denylist is None else denylist
+
+    def scrub_event(self, event):
+        # type: (Event) -> Event
+        # ... scrubbing logic discussed below
+        return event
+
+
+# ===========================================================================
+# client.py
+
+def capture_event(self, event):
+    # ... existing logic
+    if not config.send_default_pii:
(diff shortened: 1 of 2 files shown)


text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+    def scrub_event(self, event):
+        # type: (Event) -> Event
+        # ... scrubbing logic discussed below
+        return event
+
+
+# ===========================================================================
+# client.py
+
+def capture_event(self, event):
```
cleptric, line 96: So here we would expect an SDK to process an event like event processors -> event scrubber -> before send?

sl0thentr0py: yep, since we use event processors internally for integrations as well, I would say so.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+
+sentry_sdk.init(
+    send_default_pii=False,
+    event_scrubber=EventScrubber(denylist=["custom_sensitive_field"]),  # has default impl
+)
+```
+
+The actual scrubbing logic can be implemented in two ways with trade-offs outlined below.
+
+## A: Walk whole tree
```
cleptric, line 116: While this might be an expensive option in terms of computation, I still prefer it over option B. 

Do we expect users to set `password`, `Password`, `PASSWORD`, or should there be some some convenience built in?

sl0thentr0py: good point, I guess checking lowercase doesn't hurt so we cover more cases?

cleptric:
Overall I'm 👍 on this! While someone could argue that this is just a default impl of `before_send´, I like the simplicity.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+* `event.logentry.params`
+* `event.user`
+* `event.breadcrumbs.values -> value.data`
+* `event.exception.values -> value.stacktrace.frames -> frame.vars`
+* `event.spans -> span.data`
+
+## (Optional) C: Add separate security denylist
+
+A related proposal is to split the `denylist` into `pii_denylist` and `security_denylist` where stuff in `security_denylist` is **always** scrubbed
+irrespective of `send_default_pii` and stuff in `pii_denylist` is only scrubbed if `send_default_pii` is `False`.
```
antonpirker, line 192: I am for option A) (because it is secure by default and will not rot as fast)

But with two deny lists for PII and security data. 

To make it impossible to send session data by setting `sendDefaultPII` to true. It could be though that some people are depending on having session data for debugging, but if a lot of people complain we can introduce `sendDefaultSecurityData`

text/0062-controlling-pii-and-credentials-in-sd-ks.md
antonpirker: Yes, lowercase matching is the way.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Drawbacks
+
+* Option A: we add one more pass through the event object which adds some non-trivial computational overhead
```
adinauer, line 196: If code obfuscation is turned on (e.g. ProGuard on Android) Option A could break. Assuming we check typed fields we'd have to use reflection to get the name of the field. This name might have been obfuscated thus not matching the key in the deny list. We'd have to apply the filtering to the serialized event or some step between but that would cause issues with ordering vs `beforeSend` as that again expects a non serialized event (or transaction for `beforeSendTransaction`) at the moment. Maybe a filter method could internally iterate fields considering serialization keys.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+                d[k] = self.scrub_dict(v)
+            elif isinstance(v, list):
+                d[k] = self.scrub_list(v)
+        return d
+
+    def scrub_event(self, event):
+        return self.scrub_dict(event)
+```
+
+Here we would walk the whole event dict and whenever we encounter a key from the denylist, we would replace the value with `[Filtered]`.
```
adinauer, line 142: Could false positives bite use here? I might want to filter a property with the same name in HTTP headers but not in some other `data` field or similar.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+* `event.request.headers`
+* `event.request.cookies`
+* `event.request.data`
+* `event.logentry.params`
+* `event.user`
+* `event.breadcrumbs.values -> value.data`
+* `event.exception.values -> value.stacktrace.frames -> frame.vars`
+* `event.spans -> span.data`
+
+## (Optional) C: Add separate security denylist
```
adinauer, line 189: Is this intended as a second line of defense against leaking security related things? I assume we'd still prefer not to put the data into any of our data structures in the first place.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
sl0thentr0py: yes I believe most typed languages would have to go with option B.
The dict without interfaces seems to be mostly a python thing.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
sl0thentr0py: if we implement the scrubber in this RFC, I personally would like to move away from all those `send_default_pii` checks scattered around the codebase, but that's up for debate.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
mdtro: +1

A separation between the two would be great here. It would significantly reduce the risks one has to consider when toggling the `sendDefaultPII` option.

Some may gravitate to using a sensitive value like a session cookie to track through their apps, but there are more secure ways of doing this with an identifier not inherently tied to auth.

sl0thentr0py: :O thanks for stopping by on my RFC @mdtro! <3

text/0062-controlling-pii-and-credentials-in-sd-ks.md
adinauer: I meant more on the security side of things where we add data we never want to send to Sentry only to be removed again by the security denylist.

text/0062-controlling-pii-and-credentials-in-sd-ks.md
```diff
@@ -0,0 +1,207 @@
...
+| android    | okhttp                  | [headers](https://github.com/getsentry/sentry-java/blob/9b6c37f33376556ce0f66e526541a04e9e3aa60f/sentry-android-okhttp/src/main/java/io/sentry/android/okhttp/SentryOkHttpInterceptor.kt#L223-L242) | no headers at all if `send_default_pii` false, filter sensitive even if true |
+
+As demonstrated by the collection above, each SDK does slightly different things and the decisions on what is gated behind `send_default_pii` are mostly left to the integration authors.  
+The user can merely choose to follow the author's implementation with this flag but not really be sure if and what data is leaking without reading the integration source, which is an undesirable state of affairs.
+
+
+#### `before_send` and `add_event_processor`
+
+In cases where users are not happy with `send_default_pii`, we suggest using either `before_send` for errors or `add_event_processor` for all events for controlling PII on the final event payload.
```
philipphofmann, line 61: We seem to be fighting the symptom of not knowing which PIIs our SDKs are sending instead of solving the underlying problem. Users should trust our SDKs not to transmit any PII data to Sentry per default. The SDK maintainers are fully responsible for the data our SDKs send to Sentry. Asking our users to implement some of the options below is loading the problem off to them. How can they be sure a newly added feature by us doesn't start sending PII? Do they have to adapt their logic for new features? Even with the endeavors below, we can't guarantee to scrub all PII from our payloads.
Users are responsible if they manually add any PII to Sentry data. We can't check all their data to ensure they don't send PII to us. Of course, we could go the extra mile to implement one of the options, but that's just being polite, and I wonder if that's required.
Instead, we have to go through every SDK to ensure they don't send any PII if send_default_pii is off. Furthermore, we have to find a way to ensure no new feature starts transmitting PII by accident.
One step in this direction could be to have a checkbox in PR templates to validate that no PII gets sent by accident. We recently added this to some of the mobile SDKs: https://github.com/getsentry/sentry-cocoa/pull/2651/files

sl0thentr0py: we don't always know what is PII in a user's application?

philipphofmann: @sl0thentr0py, but we know which data we collect automatically and send to Sentry. If we can't ensure it doesn't contain PII, we should only send it if `send_default_pii` is on. There is a big difference between the user manually actively passing PII to Sentry, like `scope.user.email = 'my@user.com'` , or we intercepting HTTP requests and collecting headers and parameters.

HazAT: @philipphofmann I believe it's part of the exercise to double-check what we send and what we can catch.
Adding a checkbox to the PR template at least reminds you to think about it.
As always, a 100% guarantee doesn't exist, we can try our best tho.

sl0thentr0py: > If we can't ensure it doesn't contain PII, we should only send it if send_default_pii is on.

this is impractical, headers or stack variables can always potentially contain sensitive stuff. So we'd have to just stop sending headers in this case, which is a severe functionality regression.

> Asking our users to implement some of the options below is loading the problem off to them. 

We are not asking them to implement anything, everything in this RFC has defaults that then can be _easily_ tweaked in increasing complexity if they so wish to do so.

I'm not sure if the proposal came across effectively, so I will attempt a simpler version.

The current control for PII is `send_default_pii`. This RFC expands those controls to a combination of `send_default_pii` and a new `denylist` (that has defaults). 

This way the user can *additionally* let us know what in *their application* could be PII on top of what we think is PII. This is a more comprehensive handling of the state of affairs than currently exists. We are expanding the semantics of what is PII and giving the user a say in choosing those semantics.

Everything else in this RFC is implementation details for expanding that basic semantic.

philipphofmann: > this is impractical, headers or stack variables can always potentially contain sensitive stuff. So we'd have to just stop sending all headers in this case, which is a severe functionality regression.

Can we narrow it down to headers that don't contain any PII? If not, we have to ensure to scrub all data out of the headers to not send any PII. 

> We are not asking them to implement anything, everything in this RFC has defaults that then can be easily tweaked in increasing complexity if they so wish to do so.

Sorry, my bad, then. It was late, and maybe I didn't read the RFC thoroughly.

> This way the user can additionally let us know what in their application could be PII on top of what we think is PII. This is a more comprehensive handling of the state of affairs than currently exists. We are expanding the semantics of what is PII and giving the user a say in choosing those semantics.

What I don't like about this approach is that even if `send_default_pii` is disabled, we will send PII to Sentry if users put PII into HTTP headers, not following a proper naming pattern and even if we have good data scrubbers. For example, how can we ensure users are not doing
```
Nammee-Addressi: Philipp Hofmann And My Address and because I don't know what I'm during here is also my social security number 123456789
```

Why don't we instead only collect headers not meant to contain PII? So basically go with an allowlist of HTTP headers that users can add headers to? Like for example:

```
Age: 2318192
Cache-Control: public, max-age=315360000
Connection: keep-alive
Date: Mon, 18 Jul 2016 16:06:00 GMT
Server: Apache
Vary: Accept-Encoding
X-Backend-Server: developer6.webapp.scl3.mozilla.com
X-Cache: Hit from cloudfront
...
```

sl0thentr0py: > Can we narrow it down to headers that don't contain any PII? 

No we cannot which is why we let users tell us what is PII via a `denylist`.

> how can we ensure users are not doing

we cannot, which is why we rely on users to tell us and add `Namee-Addressi` to the `denylist`.

>   basically go with an allowlist

the default of product design should be adding value (which here is having most headers by default) with the _optionality_ to remove when it is not appropriate. The `allowlist` is taking away the feature by default and selectively turning it on.

I will stick to the `denylist`, others are welcome to chime in on what they prefer.

philipphofmann: > the default of product design (and a big reason our SDKs are successful) should be adding value

Yes, but privacy and security should be more important than customer value. Is it clear to customers that we send PII if they disable `send_default_pii`? How common is it that PII ends up in HTTP headers?

Your suggested approach might work if sending PII is an edge case. So the approach would be that we try to avoid sending PII, but we can't guarantee that. Here is the list of features that might collect PII, so please double-check. You can configure an extra deny list to ensure PII doesn't end up in Sentry. If the docs explain this, it might be acceptable.

cleptric: > Yes, but privacy and security should be more important than customer value.
Why would someone use any service that doesn't add value? I feel like this is not a good argument to make here.

Having a pragmatic way of removing Pii/sensitive data from the data we collect, or set by the user, seems to be missing in today's SDKs.
A on/off toggle doesn't cut it, hence the approach outlined in this RFC is necessary.

Our users should still be able to send us their request headers/bodies, with an easy-to-use API that can perform client-side scrubbing.

sl0thentr0py: >  If it's turned off, and we still send PII

The name of the option is `send_default_pii`, not `send_pii`. Without the user's input about what PII is in *their application*, we can not *deterministically* tell ourselves what is PII or not.

> Off. We don't send any PII, period.

This is not possible with the current state of affairs, even if you remove headers and such. We are trying to improve the current handling, but we cannot still end up in a system where only a boolean controls this. The only way to guarantee this 100% is to not use Sentry.

philipphofmann: > The name of the option is send_default_pii, not send_pii. Without the user's input about what PII is in their application, we can not deterministically tell ourselves what is PII or not.

I'm confused by the name, then. What's default PII?

> Off. We don't send any PII, period.

I'm only referring to data we automatically collect. If the user sends an email address to the scope, it's their responsibility. 

Maybe it's better to hop on a call, @sl0thentr0py, to discuss this.

sl0thentr0py: I will add it to TSC.

cleptric (approved):

PR #63, opened 2023-01-18
rfc(feature): 0063 SDK Crash Monitoring

philipphofmann:
Proposal for detecting crashes caused by our SDKs to improve reliability.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/sdk-crash-monitoring/text/0063-sdk-crash-monitoring.md)

### DIFF
--- a/text/0063-sdk-crash-monitoring.md
+++ b/text/0063-sdk-crash-monitoring.md
@@ -1,0 +1,87 @@
+# SDK Crash Monitoring
+
+- Start Date: 2023-01-18
+- RFC Type: feature
+- RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
+- RFC Status: draft
+- RFC Driver: [Philipp Hofmann](https://github.com/philipphofmann)
+- RFC Approver: tbd
+
+## Summary
+
+This RFC aims to detect crashes caused by our SDKs to improve reliability.
+
+## Motivation
+
+As an APM company, the reliability of our SDKs is one of our most essential quality goals. If our SDK breaks the customer, we fail. Our SDK philosophy refers to this as [degrade gracefully](https://develop.sentry.dev/sdk/philosophy/#degrade-gracefully).
+
+For some SDKs, like mobile SDKs, we primarily rely on users to reach out when our SDKs cause crashes cause we don't operate them in production. If users don't report them, we are unaware. Instead, we should detect crashes caused by our SDKs when they happen so we can proactively fix them.
+
+https://develop.sentry.dev/sdk/client-reports/
+
+### No APM for SDKs
+
+The proposed solution is not an APM solution for our SDKs. Our SDKs should not send extra events to Sentry to inform us about a problem cause of PII and the overhead of the extra requests.
+
+## Background
+
+Google Play SDK console only works for Android.
+
+## Options Considered
+
+- [Option 1: SDKs report FIOMT as errors](#option-1)
+- [Option 2: SDKs report FIOMT as performance issues](#option-2)
+- [Option 3: Ingest reports FIOMT as performance issues](#option-3)
+
+### Option 1: Detect during event processing <a name="option-1"></a>
+
+During event processing, after processing the stacktrace, the server detects if a crash stems from any of our SDKs by looking at the top frames of the stacktrace. If the server detects that it does, it could duplicate the event and store it in a special-cased sentry org where each SDK gets its project.
+
+A good candidate to add this functionality is the `event_manager`. Similarly, where we call [`_detect_performance_problems`](https://github.com/getsentry/sentry/blob/4525f70a1fb521445bbb4c9250b2e15e05b059c3/src/sentry/event_manager.py#L2461), we could add an extra function called `detect_sdk_crashes`.
+
+As we’d only make this visible to Sentry employees, we might not have to strip out any data cause of data privacy reasons, as employees could view the original events anyways. Of course, employees would need to go through the _admin portal and form.
+
+### Pros <a name="option-1-pros"></a>
+
+1. No need for per-SDK rollout.
+2. No extra events or data sent to Sentry.
+
+### Cons <a name="option-1-cons"></a>
+
+1. Please add your cons.
+
+### Option 2: Detect in SDKs <a name="option-2"></a>
+
+When the SDK sends a crash event to Sentry, it checks the stacktrace and checks if the crash stems from the SDK itself by looking at the top frames of the stacktrace. If it does, the SDK duplicates the event and sends it to a special-cased sentry org where each SDK gets its project.
+
+### Cons <a name="option-2-cons"></a>
+
+1. Opposite of [Pro 1 of option 1](#option-1-pro).
+2. Extra events sent from SDKs to Sentry.
+3. Changing the DSN needs an SDK update.
+
+### Option 3: Client Reports <a name="option-3"></a>
+
+Similar to option 2, we use client reports instead of sending events. We would need the entire event to get enough context to fix a crash. So basically, we would add the crash event to client reports.
+
+### Pros
+
+1. Opposite of [Con 2 of option 2](#option-2-cons).
+
+### Cons
+
+1. Opposite of [Pro 1 of option 1](#option-1-pro).
+2. [Con 3 of option 2](#option-2-cons).
+3. Extend the protocol of client reports.
+4. `The client reports feature doesn't expect 100 percent correct numbers, and it is acceptable for the SDKs to lose a small number of client reports`, [develop docs](https://develop.sentry.dev/sdk/client-reports/#sdk-side-recommendations). So the SDK might drop some crashes, but maybe that's acceptable.
+5. `Bugs in our SDKs are out of scope for client reports and are not being tracked using client reports at the moment`, [develop-docs](https://develop.sentry.dev/sdk/client-reports/#basic-operation).
+
+
+## Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+## Unresolved questions
+
+- Do we need to strip sensitive values from the events because of PII or is going through the _admin portal enough?
--- a/text/0063-sdk-crash-monitoring.md
+++ b/text/0063-sdk-crash-monitoring.md
@@ -23,6 +23,8 @@ The Google Play SDK Console provides insights into crashes for SDK maintainers. 
 
 ## Options Considered
 
+For every solution, the server or the SDK has to strip all irrelevant data for us to have enough information to solve an SDK crash to reduce privacy implications and risk. They should strip stacktrace frames from our customers, remove most of the context, etc.
+
 - [Option 1: Detect during event processing](#option-1)
 - [Option 2: Detect in SDKs](#option-2)
 - [Option 3: Client Reports](#option-3)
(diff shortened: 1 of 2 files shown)


text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+As we’d only make this visible to Sentry employees, we might not have to strip out any data cause of data privacy reasons, as employees could view the original events anyways. Of course, employees would need to go through the _admin portal and form.
+
+### Pros <a name="option-1-pros"></a>
+
+1. No need for per-SDK rollout.
+2. No extra events or data sent to Sentry.
+
+### Cons <a name="option-1-cons"></a>
+
+1. Please add your cons.
```
marandaneto, line 45: Some SDKs remove the sentry frames, such as [Dart](https://github.com/getsentry/sentry-dart/blob/21d4150814ba25a944c98452c814af5d9b49ef28/dart/lib/src/sentry_stack_trace_factory.dart#L33-L37), Java and likely more.
The reason is, our frames are always part of the stack trace since its responsible for capturing and parsing it, we could start sending them and hiding them on the product tho.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+
+### Option 2: Detect in SDKs <a name="option-2"></a>
+
+When the SDK sends a crash event to Sentry, it checks the stacktrace and checks if the crash stems from the SDK itself by looking at the top frames of the stacktrace. If it does, the SDK duplicates the event and sends it to a special-cased sentry org where each SDK gets its project.
+
+### Cons <a name="option-2-cons"></a>
```
marandaneto, line 51: If the bug is somewhere in the sending pipeline or in this function, we might cause a recursive/endless bug in the SDK.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Options Considered
+
+- [Option 1: SDKs report FIOMT as errors](#option-1)
+- [Option 2: SDKs report FIOMT as performance issues](#option-2)
+- [Option 3: Ingest reports FIOMT as performance issues](#option-3)
+
+### Option 1: Detect during event processing <a name="option-1"></a>
+
+During event processing, after processing the stacktrace, the server detects if a crash stems from any of our SDKs by looking at the top frames of the stacktrace. If the server detects that it does, it could duplicate the event and store it in a special-cased sentry org where each SDK gets its project.
```
mdtro, line 32: Can we maintain data locality with this approach (ie. the hybrid cloud project)?

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,81 @@
...
+
+
+## Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+## Unresolved questions
+
+- Do we need to strip sensitive values from the events because of PII or is going through the _admin portal enough?
```
mdtro, line 81: If the sensitive data is not relevant to the problem, we should strip the data to reduce privacy implications and risk.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Background
+
+The Google Play SDK Console provides insights into crashes for SDK maintainers. We regularly use it for the Android/Java SDK. While it would be great also to build something similar for SDK maintainers within Sentry, it's a bit complicated cause of PII and such. Narrowing down the scope to only Sentry SDKs makes the problem easier to solve.
+
+## Options Considered
+
+- [Option 1: SDKs report FIOMT as errors](#option-1)
+- [Option 2: SDKs report FIOMT as performance issues](#option-2)
+- [Option 3: Ingest reports FIOMT as performance issues](#option-3)
```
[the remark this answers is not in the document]
philipphofmann: Good catch thanks.

text/0063-sdk-crash-monitoring.md
philipphofmann: We also do something similar on Cocoa, but not for unhandled errors. Do you do that as well for unhandled errors, @marandaneto?

text/0063-sdk-crash-monitoring.md
philipphofmann: Added thanks 🙏

text/0063-sdk-crash-monitoring.md
philipphofmann: Excellent point, @mdtro. That's a problem to solve. @jjbayer, can you maybe answer that question ⬆️ ?

text/0063-sdk-crash-monitoring.md
philipphofmann: Added here https://github.com/getsentry/rfcs/blob/1545d847c999be3751da7e90e231784cfffbe004/text/0063-sdk-crash-monitoring.md?plain=1#L26

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+When the SDK sends a crash event to Sentry, it checks the stacktrace and checks if the crash stems from the SDK itself by looking at the top frames of the stacktrace. If it does, the SDK also sends the event to a special-cased sentry org, where each SDK gets its project.
+
+### Cons <a name="option-2-cons"></a>
+
+1. Opposite of [Pro 1 of option 1](#option-1-pros).
+2. Extra events sent from SDKs to Sentry.
+3. Changing the DSN needs an SDK update.
+4. The SDK might end up in an endless loop, if there is a bug in this functionality.
+
+### Option 3: Client Reports <a name="option-3"></a>
```
antonpirker, line 62: This option feels like the best one to me. 

The client reports dev docs state "Client reports are a protocol feature that let clients send status reports about themselves to Sentry". And a crash is clearly a status report of a Sentry Client.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+
+1. Opposite of [Con 2 of option 2](#option-2-cons).
+
+### Cons
+
+1. Opposite of [Pro 1 of option 1](#option-1-pros).
+2. [Con 3-4 of option 2](#option-2-cons).
+3. Extend the protocol of client reports.
+4. `The client reports feature doesn't expect 100 percent correct numbers, and it is acceptable for the SDKs to lose a small number of client reports`, [develop docs](https://develop.sentry.dev/sdk/client-reports/#sdk-side-recommendations). So the SDK might drop some crashes, but maybe that's acceptable.
+5. `Bugs in our SDKs are out of scope for client reports and are not being tracked using client reports at the moment`, [develop-docs](https://develop.sentry.dev/sdk/client-reports/#basic-operation).
```
antonpirker, line 76: I do not see 3. and 5. as strictly as cons. Times and requirements change, software evolves.

About 4.: I think it is fine to loose some crashes. We have more than enough customers to catch all of the important bugs in the clients.

text/0063-sdk-crash-monitoring.md
jjbayer: Does Option 3 have any advantages over Option 2? The nice thing about Option 2 is that it would not require any backend changes.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+
+- [Option 1: Detect during event processing](#option-1)
+- [Option 2: Detect in SDKs](#option-2)
+- [Option 3: Client Reports](#option-3)
+
+### Option 1: Detect during event processing <a name="option-1"></a>
+
+During event processing, after processing the stacktrace, the server detects if a crash stems from any of our SDKs by looking at the top frames of the stacktrace. If the server detects that it does, it duplicates the event and stores it in a special-cased sentry org where each SDK gets its project.
+
+A good candidate to add this functionality is the `event_manager`. Similarly, where we call [`_detect_performance_problems`](https://github.com/getsentry/sentry/blob/4525f70a1fb521445bbb4c9250b2e15e05b059c3/src/sentry/event_manager.py#L2461), we could add an extra function called `detect_sdk_crashes`.
```
jjbayer, line 40: Option 1 seems to be the only viable option if we want this to evolve into a general by-product for library / SDK maintainers (see "Background"). However, if all we care about in the foreseeable future are sentry SDK crashes, then Option 2 is probably easier to implement.

text/0063-sdk-crash-monitoring.md
jjbayer: I guess it could be solved by actually posting HTTP requests to `sentry.sentry.io` to create the new issues (i.e. all derived errors go to the same region), but that might be problematic for performance. We can always heavily sample the number of derived errors we send though.

text/0063-sdk-crash-monitoring.md
cleptric: Not all SDKs implemented Client reports just yet, and in PHP we will likely never do it.

HazAT:
My 2 cents are, I vote for option 1.

Both options 2 & 3 produce a lot of busy work, that yet again is very hard to align across all SDKs and even then will not get it right a 100%.

I think one con to add to Option 1 is: no 100% guarantee - but one can argue this is the case for all options.

The way this could work is as simple as having a check:
```daniel
if sdk == sentry.javascript.browser 
and
frames.contain "@sentry"
return true
```

HazAT:
re @mdtro do we really tho?
we are already processing their stacktraces, we are just increasing a counter if the the top frame is ours?
I don't see why we need to ask anyone for permission on that.

text/0063-sdk-crash-monitoring.md
philipphofmann: @mdtr @jjbayer, why do we need data locality?

text/0063-sdk-crash-monitoring.md
philipphofmann: > Does Option 3 have any advantages over Option 2

Yes, the SDKs don't need to send extra events to Sentry. Client reports are attached to already outgoing events.

text/0063-sdk-crash-monitoring.md
philipphofmann: > to evolve into a general by-product for library / SDK maintainers 

Yes, one of the main ideas would be that Sentry SDK maintainers get it for free.

text/0063-sdk-crash-monitoring.md
marandaneto: Nope, Sentry frames are never there.

text/0063-sdk-crash-monitoring.md
```diff
@@ -0,0 +1,102 @@
+# SDK Crash Monitoring
+
+- Start Date: 2023-01-18
+- RFC Type: feature
+- RFC PR: [#33](https://github.com/getsentry/rfcs/pull/33)
```
kahest, line 5:
```suggestion
- RFC PR: [#63](https://github.com/getsentry/rfcs/pull/63)
```

kahest:
Looking at this strictly from the _SDK monitoring_ PoV, a big _pro_ of option 1 IMO is that detection isn't performed on the SDK side. Large part of the rationale for this initiative is to find out if SDKs do something wrong which we are not aware about, and adding code in the SDK to solve this seems a bit counter intuitive and could have a lot of limitations and drawbacks.

marandaneto:
One drawback that is not written yet is that this would only work for errors that the SDK is still able to operate/send the event to Sentry.
When the SDK has a major bug that is in the transport layer or something, the SDK won't be able to even send the event.
Just trying to understand if this is really beneficial and would help a lot or just for minor issues.

philipphofmann:
> One drawback that is not written yet is that this would only work for errors that the SDK is still able to operate/send the event to Sentry. When the SDK has a major bug that is in the transport layer or something, the SDK won't be able to even send the event. Just trying to understand if this is really beneficial and would help a lot or just for minor issues.

@marandaneto, I think our automated tests must surface such severe bugs. The goal of this proposal is definitely not to help to surface such severe bugs. This solution will be useful for edge case crashes hard to detect during CI.

kahest (approved):
Great outcome, thanks to all participants!

PR #64, opened 2023-01-18
Proposal to modify RFC workflow state process

armcknight:
This PR proposes relying more on GitHub PR status for normal RFC statuses vs requiring authors to manually update it throughout the proposal process

[Rendered RFC](https://github.com/getsentry/rfcs/blob/armcknight/meta/workflow-status-recommendations/text/0064-workflow-status.md)


philipphofmann:
We could create a GH project for this repository, add the RFC statuses to the project, and remove them from the RFC.

mwarkentin:
I wonder if we could use https://shields.io/ to inject a badge with the status pulled from the source (PR state, label, etc) into the doc itself.

armcknight:
Thanks for linking that previous attempt @chadwhitacre, I hadn't seen it. While you rightly saw the same problem as me, I do prefer the PR-based process. And it looks like you and @mitsuhiko came to the same conclusion as I do, in these comments: https://github.com/getsentry/rfcs/pull/8#issuecomment-1258185266 and https://github.com/getsentry/rfcs/pull/8#issuecomment-1258209380.

I'm not sure why you closed that PR though @chadwhitacre ?

chadwhitacre:
As I recall @mitsuhiko was [-1 on the PR approach](https://github.com/getsentry/rfcs/pull/8#issuecomment-1257761464) and so I dropped it.

PR #68, opened 2023-01-24
rfc(feature): Option for organizations to disallow use of user API tokens

mdtro:
Provide a toggle option for organizations to disallow the use of user API tokens when authenticating to their organization.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/mdtro/org-user-api-token-options/text/XXXX-org-user-token-restriction.md)


text/0068-org-user-token-restriction.md
```diff
@@ -0,0 +1,42 @@
...
+
+An organization may have stricter requirements or a lower risk appetite compared to a user. We provide options for organizations to require SSO and 2FA to be enabled on a user account that wishes to interact with their organization. However, a user can authenticate with a personal API token, which inherently does not require controls like a 2FA prompt.
+
+As a result, protecting the user API token lies solely on the user. Leaking secrets in source control, personal computers infected with malware, or vendor breaches can lead to these tokens being compromised, exposing an organization to increased risk in Sentry.
+
+If a personal user API token is leaked, an organization owner does not have the ability to revoke the token and must coordinate with the user to do so -- further increasing the time an organization is potentially exposed.
+
+# Solution
+
+1. Add a toggle in the Organization Settings to deny use of personal API tokens for authentication against their organization.
```
bukzor-sentryio, line 22: I don't know if this is feasible or even useful, but more-generalized solution would be

```
1. Add a Organization Setting to limit the maximum lifetime of acceptable tokens. Existing organizations will default to "unlimited".
1. Update authentication middleware to compare token's lifetime versus the targeted organization's preference.
```

The boolean in this document is equivalent to setting maximum-lifetime to either `0s` or `unlimited`. 

For comparison, GCP access tokens are only valid for (up to) one hour.

text/0068-org-user-token-restriction.md
```diff
@@ -0,0 +1,42 @@
+- Start Date: 2023-01-18
+- RFC Type: feature
+- RFC PR: -
+- RFC Status: `draft`
+
+# Summary
+
+Provide an option to organization owners to disallow all user interaction with their organization that is using a user API token for authentication.
```
bukzor-sentryio, line 8: The mention of "user API token" here feels overly specific / brittle to future changes.

I believe the key feature of these tokens that makes you single them out for rejection is not their type (user token) but rather the fact that they have no expiry. If that's the case, it would improve this feature to be focused on expiry lifetime, apart from token type.

mdtro: I'm targeting the user API tokens specifically here. This is where the concern is.

I specify "user" here because we also have "sentry app" API tokens that are similar, but different enough. 🙂

text/0068-org-user-token-restriction.md
mdtro: I believe this is out of scope. The issue isn't that an expired token is being used, it's that a user token is being used.

While a policy type system that limits the lifetime of a token would be more flexible, I haven't seen any asks for this. An on/off toggle is more inline with what I have seen requested by customers.

These tokens are unknown and outside of the control organization owners. This can unnecessarily and unknowingly expose an organization to increased risk if a user of their org accidentally leaks their token.

mitsuhiko:
Generally a big fan of this idea. However I think we probably need to revisit some of our user experience around the tokens. The work for source maps recently also surfaced again that user bound tokens are also not idea from a user experience point of view.

text/0068-org-user-token-restriction.md
```diff
@@ -0,0 +1,86 @@
...
+Owners should have more control on how user API tokens
+interact with their organization.
+
+In the majority of cases organization owners have requested the ability
+to disallow interaction with their organization via a user API
+token completely. These organizations typically follow our recommended
+practice of using a custom integration for automation tasks.
+
+**I'm proposing we add a toggle option under the organization settings
+to disallow usage of user API tokens for authentication to their organization.**
```
markstory, line 49: Is this a setting that we would enable by default for new orgs?

PR #70, opened 2023-01-27
rfc(decision): Document sensitive data collected

antonpirker:
We need an exact but concise documentation on what sensitive data our SDKs collect. This should be available in the SDKs documentation on docs.sentry.io and be specific to all the integrations that each SDK supports.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/document-sensitive-data-collected/text/0070-document-sensitive-data-collected.md)

### DIFF
--- a/text/0070-document-sensitive-data-collected.md
+++ b/text/0070-document-sensitive-data-collected.md
@@ -26,8 +26,13 @@ After a data incident and a meeting with legal, we said that we need to take dat
 
 Have a hand written (and maintained) table in the description that shows people in an easy to grasp way what data is collected. It also shows how the data collection is changed when certain options (like `sendDefaultPII`) are changed.
 
-Here a example on how this could look like:
-https://sentry-docs-git-antonpirker-python-fastapi-sensitive-data.sentry.dev/platforms/python/guides/fastapi/#data-collected
+Here a example on how this could look like. (After talking with our designer Jesse, having two tables makes it way easier to ingest the information)
+
+For issues:
+https://sentry-docs-git-antonpirker-python-fastapi-sensitive-data.sentry.dev/platforms/python/guides/fastapi/#data-collected-by-issue-reporting
+
+And for performance:
+https://sentry-docs-git-antonpirker-python-fastapi-sensitive-data.sentry.dev/platforms/python/guides/fastapi/#data-collected-measuring-performance
 
 The elements in the table can be different for different kinds (frontend, backend, mobile) of SDKS.
 
@@ -42,6 +47,7 @@ Here a list of all sensitive data that is collected:
 - Breadcrumb Values (`event.breadcrumbs.values -> value.data`)
 - Local vars in Exceptions (`event.exception.values -> value.stacktrace.frames -> frame.vars`)
 - Span Data (`event.spans -> span.data`)
+- ... more to be defined ...
 
 Pros:
 
--- a/text/0070-document-sensitive-data-collected.md
+++ b/text/0070-document-sensitive-data-collected.md
@@ -1,0 +1,35 @@
+- Start Date: 2023-01-27
+- RFC Type: decision
+- RFC PR: https://github.com/getsentry/rfcs/pull/70
+- RFC Status: draft
+
+# Summary
+
+One paragraph explanation of the feature or document purpose.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Background
+
+The reason this decision or document is required. This section might not always exist.
+
+# Supporting Data
+
+[Metrics to help support your decision (if applicable).]
+
+# Options Considered
+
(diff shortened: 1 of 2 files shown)


text/0070-document-sensitive-data-collected.md
```diff
@@ -0,0 +1,81 @@
...
+
+If you have another idea on how to document this, please and an option here.
+
+# Drawbacks
+
+People tend to forget about documentation and then we end up with outdated documentation, which is kind of worse than having no documentation at all.
+
+# Unresolved questions
+
+- How do we guarantee, that the documentation stays up to date with the implementation?
```
philipphofmann, line 79: A checkbox in the PR template, similar to Cocoa could help: https://github.com/getsentry/sentry-cocoa/blob/df756be1176654218cbcbb1589c4af20833c56ef/.github/pull_request_template.md?plain=1#L18

text/0070-document-sensitive-data-collected.md
```diff
@@ -0,0 +1,81 @@
...
+- Documentation need to be kept up to date with seperate PR in `sentry-docs` repo when changes to SDK are made
+- Documentation for different versions of the SDK not solved yet
+
+## B) Automatic documentation creation
+
+If we go with _Option B)_ in [RFC-0062 Controlling PII and Credentials in SDKs](https://github.com/getsentry/rfcs/pull/62) we could add doc strings in the code of the implemented `EventScrubber` and then generate documentation from this code to render a table similar to the one in Option A) in this RFC.
+
+Pros:
+
+- Generated from code, so it should be always up to date
```
philipphofmann, line 61: We need a process to ensure that developers mark code collecting PII as such. A PR template checkbox as suggested in could help.

text/0070-document-sensitive-data-collected.md
```diff
@@ -0,0 +1,81 @@
...
+Pros:
+
+- Easy understandable and nice to read documentation
+
+Cons:
+
+- Documentation need to be kept up to date with seperate PR in `sentry-docs` repo when changes to SDK are made
+- Documentation for different versions of the SDK not solved yet
+
+## B) Automatic documentation creation
```
philipphofmann, line 55: I would love to work on such a project not just for this specific case. Our docs are often not updated when we add new options / features to our SDKs. Ideally, they would be linked so they are always up to date, but I think that's a hard problem to solve.

philipphofmann (commented):
I vote for A) to get started. B) would be great in general for docs and not only for PII, but it's way harder to set up.

text/0070-document-sensitive-data-collected.md
```diff
@@ -0,0 +1,87 @@
...
+This documentation will make our customers lifes way easier while creating these records.
+This will probably be a big selling point for larger customers.
+
+# Background
+
+After a data incident and a meeting with legal, we said that we need to take data issues to the next level.
+
+# Options Considered
+
+## A) Table in docs of each integration
```
AbhiPrasad, line 25: I like A a lot - also for the reason that we could use this to play around with the idea of keeping a "feature list" of our SDKs documented somewhere, so people can compare which SDKs have what features.

B seems like a lot of effort which may not be worth it at the current time. Perhaps we can re-evaluate when we have more docs personnel?

tonyo, line 25: It feels like it'll be hard to get a nice B without a properly implemented A, so IMO option A is a good start, and then we might consider implement B for languages/frameworks that have relevant tooling and conventions (versus starting to build a one-fits-all solutions and then regretting it).

text/0070-document-sensitive-data-collected.md
```diff
@@ -0,0 +1,87 @@
...
+- Easy understandable and nice to read documentation
+
+Cons:
+
+- Documentation need to be kept up to date with seperate PR in `sentry-docs` repo when changes to SDK are made
+- Documentation for different versions of the SDK not solved yet
+
+## B) Automatic documentation creation
+
+If we go with _Option B)_ in [RFC-0062 Controlling PII and Credentials in SDKs](https://github.com/getsentry/rfcs/pull/62) we could add doc strings in the code of the implemented `EventScrubber` and then generate documentation from this code to render a table similar to the one in Option A) in this RFC.
```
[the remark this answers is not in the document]
antonpirker: This RFC was created because people were asking for documentation on what data is collected, so only having it in the code is not enough. :-) 

But I guess when we have the EventScrubber we can have a documentation on how the event scrubbing works in general and then point to the default `EventScrubber`s code to the people that want to know the details.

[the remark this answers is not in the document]
adinauer: Yeah I was referring to the other RFCs outcome of having a default filtering logic which could then also serve as docs (at least for people able to read the code).

PR #71, opened 2023-02-01
rfc(feature): Continue trace over process boundaries

antonpirker:
To make it possible to continue a trace over process boundaries (think of one program starting another one) For this to work we will use environment variables as the carrier. 

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/continue-trace-over-process-boundaries/text/0071-continue-trace-over-process-boundaries.md)

### DIFF
--- a/text/0071-continue-trace-over-process-boundaries.md
+++ b/text/0071-continue-trace-over-process-boundaries.md
@@ -31,7 +31,7 @@ See Appendix A if you want to know how this works in the Python SDK.
 
 # Decision
 
-We will implement Option A.
+We will implement **Option A**.
 
 In the [develop docs](https://develop.sentry.dev/) a specification of this RFC will be added and afterwards it will be first implemented in Python.
 
--- a/text/0071-continue-trace-over-process-boundaries.md
+++ b/text/0071-continue-trace-over-process-boundaries.md
@@ -64,6 +64,22 @@ See [Dynamic Sampling Context Payloads](https://develop.sentry.dev/sdk/performan
 
 The integrations that patch functions that are used for spawning new processes (`StdlibIntegration` in Python) should be changed so they use the parsed tracing information (if any) and serialize it to the environment variables (`SENTRY_TRACING_BAGGAGE`, `SENTRY_TRACING_SENTRY_TRACE`) for the newly spawned process. The variable `SENTRY_TRACING_USE_ENVIRONMENT` should also be set to `true` so the receiving process is picking up the information.
 
+## Option B) Passing tracing information via files
+
+While initializing, SDK checks for environment variables: `SENTRY_TRACING_FILE_BAGGAGE` and `SENTRY_TRACING_FILE_SENTRY_TRACE`. Each of those variables might contain a file path that points to a file containing the tracing information.
+
+For example, `SENTRY_TRACING_FILE_SENTRY_TRACE=/dir1/file1` indicates that `sentry-trace` value should be read from the file located at `/dir/file1`.
+
+**Note**: the actual file object might also be, for example, a named pipe.
+
+After the values are read from the files, the same considerations as in Option A apply.
+
+Main differences from Option A:
+
+* Using files, tracing information can be passed not only from a parent process to a child process, but also between processes that don't have direct parent-child relationships.
+* Files used for transport can be updated by any other process (as far as filesystem permissions allow), and the SDK can technically re-read the file multiple times during the lifetime of the host application. However, I don't have a good use case for this right now.
+* Message passing via files is generally more cumbersome and might have performance/reliability drawbacks, depending on the type of file (plain file, named pipe) and the backing filesystem.
+
 # Drawbacks
 
 Because the serialization format of the bagagge (and thus the dynamic sampling context) stays the same, just the carrier is a new one, it should be 100% compatible to other integrations and will not break dynamic sampling.
--- a/text/0071-continue-trace-over-process-boundaries.md
+++ b/text/0071-continue-trace-over-process-boundaries.md
@@ -35,7 +35,7 @@ See Appendix A if you want to know how this works in the Python SDK.
 
 ### Retrieving tracing information via environment variables:
 
-When the SDK starts up it reads an environment variable `SENTRY_TRACING_USE_ENVIRONMENT` (defaults to `False`).
+When the SDK starts up it reads an environment variable `SENTRY_TRACING_USE_ENVIRONMENT` (defaults to `True`).
 If `SENTRY_TRACING_USE_ENVIRONMENT` is set to `true` (or `true|True|TRUE|yes|Yes|YES|y|1|on`) then the SDK reads the following environment variables:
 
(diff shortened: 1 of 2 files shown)


text/0071-continue-trace-over-process-boundaries.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Retrieving tracing information via environment variables:
+
+When the SDK starts up it reads an environment variable `SENTRY_TRACING_USE_ENVIRONMENT` (defaults to `False`).
+If `SENTRY_TRACING_USE_ENVIRONMENT` is set to `true` (or `true|True|TRUE|yes|Yes|YES|y|1|on`) then the SDK reads the following environment variables:
+
+- `SENTRY_TRACING_BAGGAGE` (similar to `baggage` HTTP header)
+- `SENTRY_TRACING_SENTRY_TRACE` (similar to `sentry-trace` HTTP header)
+
+The environment variables contain the same strings that the respecitve HTTP headers would contain.
```
tonyo, line 44: We should also define what happens when multiple mechanisms are enabled, e.g. when `SENTRY_TRACING_USE_ENVIRONMENT` is set to `true`, but also we have incoming HTTP/other requests with different `sentry-trace` values. 
It raises some questions: for example, should the trace-id from `SENTRY_TRACING_SENTRY_TRACE` "override" all other trace-ids, received via HTTP? Probably not. 

Or do we want to limit this RFC to purely non-server environments?

evanpurkhiser:
This would allow errors to be associated to trace IDs as well correct? That's probably the more immediate case for the Cron monitoring feature, is being able to trace errors in sentry to specific monitor checkin failures.

text/0071-continue-trace-over-process-boundaries.md
```diff
@@ -0,0 +1,93 @@
...
+If `SENTRY_TRACING_USE_ENVIRONMENT` is set to `true` and no information can be found in `SENTRY_TRACING_BAGGAGE` or `SENTRY_TRACING_SENTRY_TRACE` then the current process is the head of trace and a dynamic sampling context should be created.
+
+See [Unified Propagation Mechanism](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#unified-propagation-mechanism) for details.
+
+See [Dynamic Sampling Context Payloads](https://develop.sentry.dev/sdk/performance/dynamic-sampling-context/#payloads) to see what data needs to be included in the dynamic sampling context.
+
+### Propagating/sending tracing information via environment variables:
+
+The integrations that patch functions that are used for spawning new processes (`StdlibIntegration` in Python) should be changed so they use the parsed tracing information (if any) and serialize it to the environment variables (`SENTRY_TRACING_BAGGAGE`, `SENTRY_TRACING_SENTRY_TRACE`) for the newly spawned process. The variable `SENTRY_TRACING_USE_ENVIRONMENT` should also be set to `true` so the receiving process is picking up the information.
+
```
tonyo, line 66: For the sake of having an alternative option, here's another one: using files. I personally don't see it a superior, but it opens some interesting possibilities.

```suggestion

## Option B) Passing tracing information via files

While initializing, SDK checks for environment variables: `SENTRY_TRACING_FILE_BAGGAGE` and `SENTRY_TRACING_FILE_SENTRY_TRACE`. Each of those variables might contain a file path that points to a file containing the tracing information.

For example, `SENTRY_TRACING_FILE_SENTRY_TRACE=/dir1/file1` indicates that `sentry-trace` value should be read from the file located at `/dir/file1`.

**Note**: the actual file object might also be, for example, a named pipe.

After the values are read from the files, the same considerations as in Option A apply.

Main differences from Option A:

* Using files, tracing information can be passed not only from a parent process to a child process, but also between processes that don't have direct parent-child relationships.
* Files used for transport can be updated by any other process (as far as filesystem permissions allow), and the SDK can technically re-read the file multiple times during the lifetime of the host application. However, I don't have a good use case for this right now.
* Message passing via files is generally more cumbersome and might have performance/reliability drawbacks, depending on the type of file (plain file, named pipe) and the backing filesystem.

```

text/0071-continue-trace-over-process-boundaries.md
```diff
@@ -0,0 +1,93 @@
...
+See Appendix A if you want to know how this works in the Python SDK.
+
+# Options Considered
+
+## Option A)
+
+### Retrieving tracing information via environment variables:
+
+When the SDK starts up it reads an environment variable `SENTRY_TRACING_USE_ENVIRONMENT` (defaults to `False`).
```
mitsuhiko, line 38: I think the default for `SENTRY_TRACING_USE_ENVIRONMENT` should be `True`. Otherwise most processes will not have that feature enabled and typically you control your spawning context anyways.

text/0071-continue-trace-over-process-boundaries.md
mitsuhiko: There can only ever be once trace so when an incoming HTTP request is coming in, it starts a new trace in a new hub, shadowing the process one. We do not have a trace to trace relationship so they will be disconnected which is reasonable.

text/0071-continue-trace-over-process-boundaries.md
mitsuhiko: I'm not a huge fan of this idea but I will admit that it has some interesting properties.

text/0071-continue-trace-over-process-boundaries.md
```diff
@@ -0,0 +1,93 @@
...
+
+How trace propagation is working in the Python SDK right now:
+
+When a request from another service is received a transaction is created. (So each request-response cycle is a transaction).
+The process of creating a transaction that attaches to an existing trace received from the calling service is done as follows:
+
+- data from the `baggage` header is parsed. If it contains values with a `sentry-` prefix the parsed baggage is frozen (marked as immutable).
+- data from the `sentry-trace` header is parsed. (`trace_id`, `parent_span_id` and `parent_sampled`) If this header exists and contains values the parsed baggage from step is is frozen (marked as immutable).
+- DEPRECATED and will be removed soon: data from the `tracestate` header is parsed. (`sentry_tracestate` and `third_party_tracestate`).
+- With all this information a new transaction is created.
```
mitsuhiko, line 87: Can we have a recommendation of what we call this transaction?

mitsuhiko (commented):
I believe a non covered omission here is if we want a flag on a transaction that says "do not emit me" but do consider me sampled (to avoid charging customers unnecessary transactions). I don't think it should be part of the RFC, but it should me mentioned in a section that refers to out of scope discussions.

text/0071-continue-trace-over-process-boundaries.md
antonpirker: this is just a description on how existing integrations work. So there is not a generic term for this transaction because for every SDK it is something different, a request-response-cycle on the server, a user-interaction in the browser and something else on mobile.

antonpirker:
> This would allow errors to be associated to trace IDs as well correct? That's probably the more immediate case for the Cron monitoring feature, is being able to trace errors in sentry to specific monitor checkin failures.

Yes, this is possible @evanpurkhiser

PR #72, opened 2023-02-01
rfc(feature): Kafka schema registry

lynnagara:
This RFC proposes introducing a centralized schema repository for the Kafka topics in use at Sentry. It provides mappings between the Kafka topics we are using to the encoding format and schemas for messages on that topic.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/schema-registry/text/0072-kafka-schema-registry.md)

### DIFF
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -67,10 +67,10 @@ topic: events
 schemas:
 	- version: 1
 	type: json
-    resource: events.schema.json
-    - version: 2
+	resource: events.schema.json
+	- version: 2
 	type: avro
-    resource: events_v2.avsc
+	resource: events_v2.avsc
 ```
 
 In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present. The version bump is only necessary for major breaking changes of the schema (like changing the encoding). Most changes should be done in backward compatible manner, and keep the same version number even if the schema changes.
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -65,12 +65,12 @@ Topic data will be stored as yaml. For example:
 ```
 topic: events
 schemas:
-  - version: 1
-  type: json
-  resource: events.schema.json
-  - version: 2
-  type: avro
-  resource: events_v2.avsc
+- version: 1
+type: json
+    resource: events.schema.json
+    - version: 2
+    type: avro
+    resource: events_v2.avsc
 ```
 
 In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present. The version bump is only necessary for major breaking changes of the schema (like changing the encoding). Most changes should be done in backward compatible manner, and keep the same version number even if the schema changes.
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -21,6 +21,7 @@ The goals of the centralized schema repository are to:
 - provide more stability by enforcing backward compatibility of changes via automation
 - explicitly enumerate ownership of schemas
 - make it easier for consumers to identify and reject bad messages without pausing the whole pipeline
+- provide examples of messages for each schema type, which would help with writing consumer tests
 
 # Supporting Data
 
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -1,35 +1,111 @@
 - Start Date: 2023-02-01
 - RFC Type: feature
-- RFC PR: <link>
+- RFC PR: n/a
 - RFC Status: draft
 
 # Summary
 
-One paragraph explanation of the feature or document purpose.
+This RFC proposes introducing a centralized schema repository for the Kafka topics in use at Sentry. It provides mappings between the Kafka topics we are using to the encoding format and schemas for messages on that topic.
+
+Topic registration and schema validation of messages will be optional for the foreseeable future.
 
 # Motivation
 
-Why are we doing this? What use cases does it support? What is the expected outcome?
+Kafka is increasingly used as the message bus between services at Sentry. A Kafka topic and its’ schema not owned by any one service, but is part of the contract between services.
 
-# Background
+As Sentry grows, the number of event types and topics in use is exploding. We are marching further towards increasingly distributed ownership of data by product and infrastructure engineering teams. We have already seen an increase in the number of incidents related with invalid data and schema issues at Sentry. It is reasonable to expect this trend to continue as the number of topics, data types, teams and engineers contributing to Sentry grows.
 
-The reason this decision or document is required. This section might not always exist.
+The goals of the centralized schema repository are to:
+
+- provide an explicit, single source of truth about how data in any given topic should look, and the contract that all consumers and producers of a topic must adhere to
+- provide more stability by enforcing backward compatibility of changes via automation
+- explicitly enumerate ownership of schemas
+- make it easier for consumers to identify and reject bad messages without pausing the whole pipeline
 
 # Supporting Data
 
-[Metrics to help support your decision (if applicable).]
+Over the last quarter, we have seen many incidents related to schema disagreements at Sentry. Examples include the following: (note: inc numbers are internal to Sentry)
+
+- Post process forwarder (INC-218)
+- Snuba’s transactions consumer (INC-210)
+- Super big consumers (INC-220)
+- Replays consumer (INC-281)
+
+Many of these incidents are P1. Since messages are in order, an invalid message often halts consumers and requires manual intervention. This is disruptive to both Sentry engineers and our users (it takes us much longer to recover).
 
 # Options Considered
 
-If an RFC does not know yet what the options are, it can propose multiple options. The
-preferred model is to propose one option and to provide alternatives.
+### **Option A (preferred): Publish a library**
+
+Schemas will be made available as a library for multiple programming languages (Python and Rust to start).
+
+**Client usage example:**
+
+```python
+from sentry_kafka_schemas import get_schema
+
+topic = "events"
+schema = get_schema(topic)
+# => {
+# 	"type": "json"
+# 	"schema": {
+# 		"$schema": "http://json-schema.org/draft-07/schema#",
+# 		"type": "object",
+# 		...
+# 	}
+# }
+
+# Get a specific version of the schema
+v1_schema = get_schema(topic, version=1)
+```
+
+**How schemas are stored:**
+Topic data will be stored as yaml. For example:
+
+```yaml
+topic: events
+schemas:
+	- version: 1
+	type: json
+    resource: events.schema.json
+    - version: 2
+	type: avro
+    resource: events_v2.avsc
+```
+
+In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present. The version bump is only necessary for major breaking changes of the schema (like changing the encoding). Most changes should be done in backward compatible manner, and keep the same version number even if the schema changes.
+
+**Data governance**
+The schemas repository will have checks in place to ensure schemas are valid and non backwards compatible changes are not being introduced with a same version number.
+
+### **Option B (alternative, non-preferred option): Deploy a separate service **
+
+No library is provided. Clients fetch data from the schemas service and need to know how to parse the schema from the response by themselves. The schemas service could either be built from scratch or we could deploy an existing open source implementation such as Confluent schema registry.
+
+**Example usage:**
+
+```bash
+GET /schemas/events
+GET /schemas/events?version=1
+```
+
+Or, using the Confluent schema registry API
+
+```bash
+GET /subjects/events/versions
+GET /schemas/ids/123/schema
+
+# Creating schemas. Unlike other options presented, schemas are not checked into code
+POST /subjects/events/version/
+{
+    "schema": {"type": .........}
+}
+```
+
+Unlike Option A, consumer and producer code should be updated automatically when a new version of the schemas service is deployed. This requires consumers and producers to frequently check for updates to the schema to stay in sync. If using the Confluent API, there are libraries to manage this. There are additional complexities of adding another service, like network access to the schema service and schema service capacity to figure out.
 
 # Drawbacks
 
-Why should we not do this? What are the drawbacks of this RFC or a particular option if
-multiple options are presented.
-
-# Unresolved questions
-
-- What parts of the design do you expect to resolve through this RFC?
-- What issues are out of scope for this RFC but are known?
+- We expect a significant performance hit if we were to validate every message in Python consumers (and producers). This has not yet been benchmarked for the various shapes of messages we process. Even with multiprocessing, we may have to sample or not validate in every scenario
+- Additional step involved in making schema changes to Kafka topics
+- Option B in particular adds quite a bit of additional overhead to Sentry’s infrastructure and would need to be shipped in all environments: all regions, open source, dev, CI and single tenant installations.
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -68,17 +68,25 @@ Topic data will be stored as yaml. For example:
 topic: events
 schemas:
     - version: 1
+    compatibility_mode: backward
     type: json
     resource: events.schema.json
     - version: 2
+    compatibility_mode: none
     type: avro
     resource: events_v2.avsc
 ```
 
-In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present. The version bump is only necessary for major breaking changes of the schema (like changing the encoding). Most changes should be done in backward compatible manner, and keep the same version number even if the schema changes.
+In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present.
 
-**Data governance**
-The schemas repository will have checks in place to ensure schemas are valid and non backwards compatible changes are not being introduced with a same version number.
+**Compatibility modes:**
+Each schema version defines it's compatibility mode. There will be 2 to start but more can be added if we want to change or tighten the rules.
+
+- `none` - Any changes are allowed. Generally used if a feature is in dev to allow for fast iteration and breaking changes.
+- `backward` - Allows adding optional fields, removing optional fields, and changing from optional to required and required to optional. Required field cannot be added at once, it must be split into 2 separate releases.
+
+If `backward` is selected, CI in the schemas repository will ensure changes that are not allowed are not being introduced with a same version number
+
 
 ### **Option B (alternative, non-preferred option): Deploy a separate service**
 
--- a/text/0072-kafka-schema-registry.md
+++ b/text/0072-kafka-schema-registry.md
@@ -63,6 +63,7 @@ v1_schema = get_schema(topic, version=1)
 ```
 
 **How schemas are stored:**
+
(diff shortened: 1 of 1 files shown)


text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+
+# Summary
+
+This RFC proposes introducing a centralized schema repository for the Kafka topics in use at Sentry. It provides mappings between the Kafka topics we are using to the encoding format and schemas for messages on that topic.
+
+Topic registration and schema validation of messages will be optional for the foreseeable future.
+
+# Motivation
+
+Kafka is increasingly used as the message bus between services at Sentry. A Kafka topic and its’ schema not owned by any one service, but is part of the contract between services.
```
markstory, line 14:
```suggestion
Kafka is increasingly used as the message bus between services at Sentry. A Kafka topic and its’ schema are not owned by any one service, but is part of the contract between services.
```

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+
+As Sentry grows, the number of event types and topics in use is exploding. We are marching further towards increasingly distributed ownership of data by product and infrastructure engineering teams. We have already seen an increase in the number of incidents related with invalid data and schema issues at Sentry. It is reasonable to expect this trend to continue as the number of topics, data types, teams and engineers contributing to Sentry grows.
```
markstory, line 16:
```suggestion
As Sentry grows, the number of message types and topics in use is exploding. We are marching further towards increasingly distributed ownership of data by product and infrastructure engineering teams. We have already seen an increase in the number of incidents related with invalid data and schema issues at Sentry. It is reasonable to expect this trend to continue as the number of topics, data types, teams and engineers contributing to Sentry grows.
```

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+Topic data will be stored as yaml. For example:
+
+```
+topic: events
+schemas:
+    - version: 1
+    type: json
+    resource: events.schema.json
+    - version: 2
+    type: avro
```
markstory, line 72: We also have grpc/protobufs floating around in our codebase. Do we need json, avro, and protobufs?

text/0072-kafka-schema-registry.md
fpacifici, line 14: On ownership. I think we can have owners for schemas. The issue we have is that the contract is not stated anywhere, not that there never is an owner.

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+    resource: events_v2.avsc
+```
+
+In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present. The version bump is only necessary for major breaking changes of the schema (like changing the encoding). Most changes should be done in backward compatible manner, and keep the same version number even if the schema changes.
```
fpacifici, line 76: Are we going to expect each message to carry the version of the schema it complies with ?

fpacifici (commented):
I suggest you run some profiling to assess the impact of the validation on a consumer before we discuss this at the Infra TSC

text/0072-kafka-schema-registry.md
lynnagara: Reworded to `internal` - the point I was attempting to make is that it's part of the external contract. Hopefully that is clearer.

text/0072-kafka-schema-registry.md
lynnagara: No - primarily because I don't want to enforce making this change across the board. I think we can just default to the latest version.

Now if the version upgrade is actually required on a topic, we would probably need to add that metadata (likely in the header) at that point so we can make the change without downtime and support both versions temporarily.

text/0072-kafka-schema-registry.md
lynnagara: Good question. We will most likely focus on supporting json to start. The reason for this is not technical, just that we are focussing on golden paths and don't necessarily want to encourage too many different encodings and schema formats at Sentry. But this type field will be here from the start, and the process to add other formats should be pretty straightforward.

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+
+Topic registration and schema validation of messages will be optional for the foreseeable future.
+
+# Motivation
+
+Kafka is increasingly used as the message bus between services at Sentry. Kafka topics and their schemas are usually not internal to any one service, but part of the contract between services.
+
+As Sentry grows, the number of message types and topics in use is exploding. We are marching further towards increasingly distributed ownership of data by product and infrastructure engineering teams. We have already seen an increase in the number of incidents related with invalid data and schema issues at Sentry. It is reasonable to expect this trend to continue as the number of topics, data types, teams and engineers contributing to Sentry grows.
+
+The goals of the centralized schema repository are to:
```
nikhars, line 18: One thing which I have found useful with other schema repositories in the past is having example messages for each schema type. Looking at schemas versus actual example messages complying with the schema are two different things. Once example messages are there for schemas its easy for consumers to write tests using example messages. Should we consider adding an example message for each schema in the schema repository a requirement to make it easier for consumers to write tests?

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,111 @@
...
+
+Schemas will be made available as a library for multiple programming languages (Python and Rust to start).
+
+**Client usage example:**
+
+```python
+from sentry_kafka_schemas import get_schema
+
+topic = "events"
+schema = get_schema(topic)
```
nikhars, line 48: Wanted to add something which might not be relevant on which option to choose. But more around how option A should look like. IMO, we shouldn't allow getting a schema without a version. Most folks won't know what this would mean. Is it the first version, or is it the latest version? If its the latest version, how can one know whether updating the schema library version causes backward incompatible change. Forcing versions in the api makes it explicit which version of schema you use. It also makes it easier finding consumers who might be on older versions more easily.

text/0072-kafka-schema-registry.md
lynnagara: Added to the list of goals

text/0072-kafka-schema-registry.md
lynnagara: I have been thinking about this quite a bit. I prefer to omit by default- though this is not a strong preference and I agree there are some benefits to being explicit. My thinking around this was:
- Looking at our usage of Kafka so far, I think at Sentry it's actually likely that we will never bump a version number *ever* on any topic. Though I added this example to this RFC, I don't actually expect our usage to look like this (unless we completely switch out the message format or something which would be a larger undertaking). I think it's more likely that we will continue to evolve our schema in a backward compatible way and none of the versions will ever go past 1
- Given that we never update the schema in a non backward compatible way, none of our messages actually contain version numbers. I would like to keep it this way rather than start forcing everyone to put extra metadata inside their messages.
- The alternative to putting the version in the message could be hardcoding the version number inside all of our codebases next to every topic. But I think this is a bit of extra work if we're never really bumping major versions anyway.
- The other option I considered was to remove `version` out of here entirely so that only backwards compatible changes are allowed. However, I think it's might be better to future proof even if it's not that likely we'll use it - it's still there if we want it later and we actually need to make a breaking change.

markstory: If we don't have versioned payloads, would we consider having versioned topics instead? For example, when a producer needs to make a breaking change to a message format, the producer starts publishing to a new topic. This makes breaking changes more expensive but I think that's a benefit.

lynnagara: The payload is still supposed to be versioned (just fixed the example to actually include this field) and the user can always request a specific version of the schema. Here we are just debating whether the schema version should default to latest if none is passed.

About the decision whether to publish an entire new topic or bump the major version instead - I think that is up to the developer to decide. The current design of the registry would allow for both options and doesn't really take an opinion about it either way.

fpacifici: A case for storing the version on each message (possibly as a header - headers are for metadata, this seems a good use), applying a version to every schema change no matter on the compatibility model we choose and making the consumer declare the highest version they support. 

This approach would allow us to log warning and collect metrics about schema mismatches (even for backwards compatible changes). 
I define a schema mismatch as a consumer dealing with a message that has a higher version than the highest version it supports, even if the changes are all backwards compatible and the consumer can still deal with the data it receives.

I believe we are interested in highlighting these changes because they can hide semantic incompatibilities. For instance, let's imagine I introduce an data category on outcomes to separate dynamic sampling data from others, a consumer may ignore the new field and, for what concerns schema validation all would be good.
The application may be broken in a subtle way.

If, instead, I start seeing warning about a schema mismatch after adding this new field and forgetting to adapt the consumer business logic to take it into account, I have a chance to react and fix the application.

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,113 @@
...
+- Super big consumers (INC-220)
+- Replays consumer (INC-250, INC-281)
+
+Many of these incidents are P1. Since messages are in order, an invalid message often halts consumers and requires manual intervention. This is disruptive to both Sentry engineers and our users (it takes us much longer to recover).
+
+# Options Considered
+
+### **Option A (preferred): Publish a library**
+
+Schemas will be made available as a library for multiple programming languages (Python and Rust to start).
```
fpacifici, line 41: A concern was raised during the TSC discussion about the inadequacy of the tooling we have today to handle cross repo dependencies. (think about the pain when cutting new arroyo releases).

I think we need to mitigate that issue and, specifically for this RFC we should outline the process that we need to follow when applying schema changes in order to propagate them to all clients of the library.

I would try to answer these questions:
- versioning and releasing. Are we going to follow the same releasing process we have for other internal libraries like Arroyo? How do we make it seamless?
- How do we encourage clients to stay up to date? Are we going to build automation in all consumers? Or are we going to make arroyo provide tools to fetch the recent schema?
- Is there a revert process in case we sneak a bad change into the repo by mistake?


I would focus on keeping the process as light and as seamless as possible. If there is too much bureaucracy to begin with, there will be resistance in adopting the schema itself.

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,113 @@
...
+# Motivation
+
+Kafka is increasingly used as the message bus between services at Sentry. Kafka topics and their schemas are usually not internal to any one service, but part of the contract between services.
+
+As Sentry grows, the number of message types and topics in use is exploding. We are marching further towards increasingly distributed ownership of data by product and infrastructure engineering teams. We have already seen an increase in the number of incidents related with invalid data and schema issues at Sentry. It is reasonable to expect this trend to continue as the number of topics, data types, teams and engineers contributing to Sentry grows.
+
+The goals of the centralized schema repository are to:
+
+- provide an explicit, single source of truth about how data in any given topic should look, and the contract that all consumers and producers of a topic must adhere to
+- provide more stability by enforcing backward compatibility of changes via automation
```
fpacifici, line 21: Two questions about backwards compatibility:
- I think we should define it a bit more formally. Maybe not everybody has the same concept of what can and cannot be done in mind.
- Do you plan to stick with one model of backward compatibility for the entire registry or is the compatibility model a property of a topic/development phase?

For the second question, for the sake of keeping development lean and quick, I think we should not force strict backward compatibility rule early on when a schema is being defined. 
Chances are, at this stage, that nobody is using it yet and more fundamental changes are required before a product starts actually relying on it. Think about the breaking changes we do on Clickhouse schemas before activating the migrations in CI and self hosted. 
I think we should have a way to tweak the schema easily early on and locking it down to strict backward compatibility rules as soon as we consider it viable.
Maybe a `dev` phase for each schema before it graduates to `stable` ?

text/0072-kafka-schema-registry.md
lynnagara: I think this is a fair concern, but I think it may be a bit heavy handed for size of the problem today. Some arguments against creating new version for every schema change and putting the versions inside the messages:

- This behaviour (libraries out of sync in producer and consumer) is expected to happen under normal circumstances. It is impossible that all producers and consumers get rolled out in prod at the same moment, so any logging or metrics here will be incredibly noisy. It can even be normal for one to be updated even days ahead of another as the unfortunate reality at Sentry is that approval / deployment processes do end up taking a while. This is not really an issue as long as the changes are done in a backward compatible order.
- The problem of updating the library scales with the number of services to be updated. Today there is not too many. While it is possible in the future, I think this is a distant problem and not one we have today.
- I am concerned this repository will become too polluted with copies of almost identical schemas. There will be more friction involved in adding and changing schemas, and they will be harder to review. The library will grow too big.
- I would prefer to avoid enforcing changes to consumers/producers/message format in order for clients to adopt this repository. It will be harder to migrate onto the schema repository.
- I think bumping the version for every schema change adds complexity and lots of potential for edge cases in how the versioning system works. There would need to be more of a distinction between backwards compatible and non backwards compatible changes. We may need major and minor versions.

How about the following:
- In the short term, ensure we have an aggressive dependabot configuration set up on all services that use the schema repository. Pull request should be opened soon after the new version is available.
- If the number of services at Sentry grows, we can add the library version (or some other checksum) into the header to identify differences between producer and consumer versions without changing the whole schema versioning system.

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,113 @@
...
+- Post process forwarder (INC-218)
+- Snuba’s transactions consumer (INC-210)
+- Super big consumers (INC-220)
+- Replays consumer (INC-250, INC-281)
+
+Many of these incidents are P1. Since messages are in order, an invalid message often halts consumers and requires manual intervention. This is disruptive to both Sentry engineers and our users (it takes us much longer to recover).
+
+# Options Considered
+
+### **Option A (preferred): Publish a library**
```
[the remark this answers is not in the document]
lynnagara: As far as this RFC goes, only the schemas are in scope.

However, I do think we can also use it for the source of truth for other things like topic creation parameters.
If we wanted to add this in the future I think it would not be a difficult addition. It could be something like:
```
topic: events
schemas:
    - version: 1
    type: json
    resource: events.schema.json
topic_creation_config:
    - message.timestamp.type: LogAppendTime
```

Unfortunately I see no easy way to handle any kind of override values here as these are values that can be changed per installation.

text/0072-kafka-schema-registry.md
lynnagara: > * versioning and releasing. Are we going to follow the same releasing process we have for other internal libraries like Arroyo? How do we make it seamless?

The plan is to use Sentry standard releasing processes. Yes, there is some friction with approvers, etc that already exists today. But given these problems already exist today, I think we need to focus on solving this for all Sentry libraries not build anything special here.

> * How do we encourage clients to stay up to date? Are we going to build automation in all consumers? Or are we going to make arroyo provide tools to fetch the recent schema?

As mentioned above, I think in the short term we can just add some aggressive dependabot config on all the services that use the library.

> * Is there a revert process in case we sneak a bad change into the repo by mistake?

I don't think we should have any special revert process. Since this is a library and not a service, we don't need the kind of "fast revert" that some other Sentry repositories have. I worry that it would complicate the rules around compatibility and versioning, so I'd vote to just roll forward to undo the change.

text/0072-kafka-schema-registry.md
lynnagara: I think the idea of having various compatibility rules is a good one, as it allows for different types of development styles that may be appropriate at different products or times in product life cycle.

I will update the proposal to add a `compatibility_mode` key to the schema, and a couple of modes to start (perhaps "any" which will allow any change and "backward" or something like that).

The idea is that we could add additional modes in the future, if we want to, say, start enforcing stricter or different rules on specific topics

text/0072-kafka-schema-registry.md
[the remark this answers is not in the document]
fpacifici: ok.

text/0072-kafka-schema-registry.md
fpacifici: > But given these problems already exist today, I think we need to focus on solving this for all Sentry libraries not build anything special here.

I don't think I agree that this is a problem that exists today. We do not have a schema registry today, changing the schema does not require any external approval just to perform the release. Rely on the current process can introduce this additional step that does not add any value to the quality of the schema. This will be unneeded friction. 

Could you please explore whether we can allow every owner of a schema to be allowed to cut releases ? I think that would be an ok process.
Imagine if the replays team adds a field to their schema. If they can cut a release without external approval we would be ok with the current tool. If they have to wait for somebody potentially in a different time zone that is a non starter.

text/0072-kafka-schema-registry.md
fpacifici: > In the short term, ensure we have an aggressive dependabot configuration set up on all services that use the schema repository. Pull request should be opened soon after the new version is available.

I'd be ok with this if we confirm that every owner of a schema can cut a release on their own without external approval.  Dependabot requires us to cut releases whenever we make changes to the schema.
If the process it annoying people will start depending on github dependency on specific commits in their requirements file instead of importing the package from pypi. At that point dependabot will be useless.

text/0072-kafka-schema-registry.md
fpacifici: ok

text/0072-kafka-schema-registry.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+In this example scenario, we decided to make a change to schemas on the “events” topic. The `events.schema.json` and `events_v2.avsc` files must be present.
+
+**Compatibility modes:**
+
+Each schema version defines it's compatibility mode. There will be 2 to start but more can be added if we want to change or tighten the rules.
+
+- `none` - Any changes are allowed. Generally used if a feature is in dev to allow for fast iteration and breaking changes.
+- `backward` - Allows adding optional fields, removing optional fields, and changing from optional to required and required to optional. Required field cannot be added at once, it must be split into 2 separate releases.
```
fpacifici, line 89: Why do we consider switching a field from optional to required a backward compatible change? 
I think that should be a breaking change that generate a new version.

lynnagara: Fair enough, removed that. Fwiw, I think it's ok if most of our schemas are `none` as I think we do want it to be easy to go from optional to required and vice versa. I think we do want to be able to move quickly even for the more mature datasets.

text/0072-kafka-schema-registry.md
lynnagara: Confirmed with devinfra that internal packages will get autoapproval and not require the sam release process as SDKs and customer installed packages

text/0072-kafka-schema-registry.md
untitaker: we can probably hammer msgpack payloads into a json-shaped hole if we have to. grpc is another world

lynnagara: Agreed. We are using msgpack in Sentry in at least one place so we may need to add it sooner rather than later. I think msgpack can still use the jsonschema though, just the decoder would be different.

text/0072-kafka-schema-registry.md
fpacifici: Mind writing your last statement in the RFC as well. Let's make it clear that we target a lighter release process than what we do today.

text/0072-kafka-schema-registry.md
fpacifici: Mind writing the outcome of this conversation: concern and mitigation, on the rfc itself. Otherwise the reasoning will be lost when it is approved.

fpacifici (approved):
I am ok with the current state.  I left two comments about writing the outcomes of the discussions in the rfc itself. Once this is merged those conversation would be basically lost. I think it would be useful to have the outcome in the rfc.

PR #73, opened 2023-02-10
rfc(decision): Usage of Transaction Types

philipphofmann:
This RFC aims to give Sentry developers insights into which types of transactions and spans our customers use.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/usage-of-transaction-types/text/0073-usage-of-transaction-types.md)


text/0073-event-origin.md
```diff
@@ -0,0 +1,100 @@
...
+### Cons <a name="option-1-cons"></a>
+
+1. Doesn't work for spans.
+2. Doesn't work for performance issues.
+
+## Option 2: Event Origin <a name="option-2"></a>
+
+Similar to option 1, but `origin` is a top level optional property directly on the event, to determine what exactly created the event. It has two fields: 
+
+- `type`: Required, type str. Identifies what created the event. At the moment it can be `sdk` or `performance-issue`.
```
AbhiPrasad, line 63: This is already available in the UI via `issue.category`: 
image

text/0073-event-origin.md
```diff
@@ -0,0 +1,100 @@
...
+- `name`: Required, type str. Contains more detailed information on what exactly created the event, such as: `swift-ui`, `http-client-errors`, `sentry-crash`, `metric-kit`, `anr`, `jetpack-compose`, `next-js`, `flask`, `django`, `log4net`, `apollo3`, `dio.http`, `file-io-on-main-thread`, `n+1-queries`, `n+1-api-calls`, `consecutive-db-calls`, etc.
```
AbhiPrasad, line 64: IMO things like `flask`, `django` here are redundant - especially since the sdk interface name will already catch this (`sentry.python.django`).

text/0073-event-origin.md
```diff
@@ -0,0 +1,100 @@
...
+For every option, Looker picks up the field, but we don't need to index it and make it searchable in Discover. Amplitude could look at this field as a property when users visit issue or transaction detail pages.
+
+- [Option 1: Event SDK Origin](#option-1)
+- [Option 2: Event Origin](#option-2)
+- [Option 3: Transaction Info Type](#option-3)
+
+
+## Option 1: Event SDK Origin <a name="option-1"></a>
+
+Add a new property to the [SDK interface of the event payload](https://develop.sentry.dev/sdk/event-payloads/sdk/) named `origin` to determine which part of the SDK created the event.
```
AbhiPrasad, line 37: I would say that errors can be mostly figured out based on
1) name in SDK interface
2) integration set
3) mechanism
4) contextual tags/message/event type

In addition - does it really matter what created an error? I fear this data will just cause anchoring bias.

I see how this is valuable for spans though - perhaps we add a new attribute to spans to track what generated them (transactions would inherit this). IMO I would like to see this proposal to be transactions/spans only, and focus the options on how we can identify them. This also can help in the UI, showing users what parts of their transaction were auto-instrumented/manually-instrumented. In addition, this difference can help the performance product build new features and performance issues..

philipphofmann: Thanks for the feedback, @AbhiPrasad. I will update the RFC accordingly.

text/0073-event-origin.md
philipphofmann: Thanks didn't know that.

text/0073-event-origin.md
```diff
@@ -0,0 +1,100 @@
...
+use the SDK integration list to determine which organizations have specific performance
+integrations enabled. The downside is that the SDK sends this list for each event, not
+giving us insights into how many events/transactions/spans stem from a specific parts
+of the SDK. 
+Furthermore, knowing what created an event, transaction, or span helps investigating issues.
+
+
+# Options Considered
+
+For every option, Looker picks up the field, but we don't need to index it and make it searchable in Discover. Amplitude could look at this field as a property when users visit issue or transaction detail pages.
```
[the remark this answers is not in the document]
philipphofmann: >  that were actually viewed in the UI by a user

Yes exactly. This option works great in combination with any other option and is not mutually exclusive.

https://github.com/getsentry/rfcs/blob/e4957dd4c1e493d61fa3301b735f6b7421ab4ca6/text/0073-usage-of-transaction-types.md?plain=1#L96

text/0073-event-origin.md
philipphofmann: @adinauer, which option would work best for your requirements?

adinauer: I presume we could use `mechanism` more thoroughly for errors, e.g. in logging integrations. I still believe a common solution for errors and transactions would be nice but I don't wanna drag this on just because of that. So we can try the `mechanism` approach first and then revisit if that doesn't work out. Created https://github.com/getsentry/sentry-java/issues/2555

markushi:
Nice one! I like the idea of extending the span interface with an `origin` field. Could also be quite useful to see which SDK features are burning a customers quota.
I'm still wondering what kind of insight an SDK developer would get out of this? If there are spans which are not part of any (performance) issues, they're likely less visible - but that's not necessarily bad right? I think we just need to be a bit careful when interpreting some aggregated numbers here.

text/0073-event-origin.md
philipphofmann: Added as a con to option 2 with https://github.com/getsentry/rfcs/commit/f27167861a4b27e645e9e39a92bdc5a05f0f5f86.

marandaneto:
I believe a single option here won't do, if the goal is to track the adoption and usage of our integrations similarly to how we do with `integrations` and `mechanism` for `errors`, we need at least 2 of the adoptions.

* We don't need to add a top-level property in the event protocol since there are already `integrations` and `mechanism` for that (errors specifically).
* We need to focus on the usage of transactions and spans, they are equally important since some integrations create transactions and some others only spans.

I believe that we should add a new property to the `trace` context and in the `span` protocol.
Example (look at `origin`):

```json
{
  "contexts": {
    "trace": {
      "op": "navigation",
      "description": "User clicked on <Link />",
      "trace_id": "743ad8bbfdd84e99bc38b4729e2864de",
      "span_id": "a0cfbde2bdff3adc",
      "status": "ok",
      "origin": "UserInteractionTracing",
    }
  }
}
```

And:

```json
{
  "spans": [
    {
      "start_timestamp": 1588601261.481961,
      "description": "GET /sockjs-node/info",
      "tags": {
        "http.status_code": "200"
      },
      "timestamp": 1588601261.488901,
      "parent_span_id": "b0e6f15b45c36b12",
      "trace_id": "1e57b752bc6e4544bbaa246cd1d05dee",
      "op": "http",
      "data": {
        "url": "http://localhost:8080/sockjs-node/info?t=1588601703755",
        "status_code": 200,
        "type": "xhr",
        "method": "GET"
      },
      "span_id": "b01b9f6349558cd1"
      "origin": "HttpTracing",
    },
    {
      "start_timestamp": 1588601261.535386,
      "description": "Vue <App>",
      "timestamp": 1588601261.544196,
      "parent_span_id": "9312d0d18bf51736",
      "trace_id": "1e57b752bc6e4544bbaa246cd1d05dee",
      "op": "update",
      "span_id": "b980d4dec78d7344"
      "origin": "VueTracing",
    }
  ]
}
```

Now I know that the transaction was created by the `UserInteractionTracing` integration.
One of the spans was created by the `HttpTracing` integration.
And the last span was created by the `VueTracing` integration.

If this is consumed by Looker/Redash (this is a MUST IMO), we are able to track usage and adoption of our integrations and prioritize accordingly with its importance for bug fixing, marketing pushes, looking into more features, etc.

Overall this is a +1 for this RFC.
Thanks for doing this @philipphofmann 
I really wanna see us taking decisions based on data, more, and more.

philipphofmann:
Thanks for your input @marandaneto. I added option 6 for the transaction context.

text/0073-usage-of-transaction-types.md
```diff
@@ -0,0 +1,229 @@
...
+```
+
+### Pros <a name="option-6-pros"></a>
+
+1. Helps users to understand which transactions where auto or manually instrumented.
+2. This addition can help the performance product to build new features and performance issues.
+
+### Cons <a name="option-6-cons"></a>
+
+1. Only works for transactions.
```
marandaneto, line 217: Nope, my suggestion was that Option 6 is actually Option 5 + 6, so we are able to track the transaction and span `origin`.

philipphofmann: Sorry about the misunderstanding. I updated the RFC.

marandaneto (approved):
Option 5 is my favorite.

jjbayer (approved):
Option 5 sounds good to me. Please don't forget to open a Relay PR that adds the field to both types.

adinauer (approved):

PR #74, opened 2023-02-15
rfc(feature): Source context via links

vaind:


mattjohnsonpint:
My vote is for option A.

It seems like with option B we'd need to think about *anywhere* that data might need to be rendered - which might not just be the UI.  You called out search, but there could be others.  Perhaps with integrations or alerts (if not now, then in the future).

mattjohnsonpint:
A third option (not my first choice though) would be to have Sentry CLI download the files from the links and use them to create a source bundle, then upload the source bundle through the existing mechanism.

That would address .NET SourceLink, but I'm not sure it would cover the other cases mentioned.

text/0074-source-context-via-links.md
```diff
@@ -0,0 +1,70 @@
...
+- `context_line` - Source code in filename at `lineno`.
+- `pre_context` - A list of source code lines before `context_line` (in order) – usually [`lineno - 5:lineno`].
+- `post_context` - A list of source code lines after `context_line` (in order) – usually [`lineno + 1:lineno + 5`].
+
+Showing a source code context and a link to it (e.g. to GitHub) currently works only if the source code itself is
+available during symbolication, thus the attributes are filled in at that time, or if they were already sent with the event by the SDK.
+
+There are, however, situations where we do have a URL where the source code resides, but not the contents (without downloading it), for example:
+
+- Portable-PDB source-link (.NET, see [this `symbolic` issue](https://github.com/getsentry/symbolic/issues/735))
```
Swatinem, line 24: other examples include:

- Debuginfod servers (we don’t support these yet)
- SourceMaps (either embedded `sourcesContent` or using individual source files)

text/0074-source-context-via-links.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Pros
+
+- Reuses existing fields (changes only needed in the `Symbolicator`).
+- Sources surrounding the frame line are then stored in the database, enabling [search-by-code](https://github.com/getsentry/sentry/issues/3755).
+- May be possible to do server-side authentication (based on project/org configuration) - this would complicate the solution & caching though, as opposed to only supporting publicly available sources.
+
+### Cons
+
+- Downloads all frame sources even though the event/frame may never be viewed by a user.
```
Swatinem, line 50: Slightly related:
We should make "we want source context" configurable / an input option in symbolicator.
For example the profiling product goes through symbolicator, but does not use source context at all (they throw it away in the profiling processor AFAIK). We could potentially make profiles cheaper to symbolicate by just not doing source context resolution at all. CC @indragiek 

It may also be worth considering making the lines of context configurable, as it is hardcoded to `5` right now.

text/0074-source-context-via-links.md
```diff
@@ -0,0 +1,70 @@
...
+- Adds potentially tens of requests for each symbolicated event that has a source link - as many as there are in the stack trace.
```
Swatinem, line 51: Agreed, it adds a ton of requests, and processing to stack frames that customers are very unlikely to ever look at (for example background threads). Though good caching should hopefully take care of most of this.

I’m not sure we have metrics for any of this right now? How many stack traces users are actually looking at?

text/0074-source-context-via-links.md
```diff
@@ -0,0 +1,70 @@
...
+
+## B | Add a source-link to the `Frame` and let the UI handle it
+
+Another option is to extend `Frame` attributes with a new field, e.g. `source_link` and let UI download the source code
+as needed or just display the link.
+
+### Pros
+
+- No overhead on the server for loading sources in situations where the event/frame wouldn't be shown.
+- Requests to fetch the source code, if any, are made by the user browser, thus avoiding potential quota limits.
```
Swatinem, line 61: Well, while *we* as a single large consumer would not run into quotas, the source server itself would get a ton more requests by individual users that are not being cached.

text/0074-source-context-via-links.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Cons
+
+- Sources not available in the database -> can't [search-by-code](https://github.com/getsentry/sentry/issues/3755).
+- Accessing private sources may be problematic. We could still show the source link, though.
+
+# Unresolved questions
+
+- Neither approach is a clear winner, any suggestions?
```
Swatinem, line 70: I believe from an implementation POV, we can even do a hybrid approach:
- Always attach a source link to stack frames.
- Only download/apply source context for the top N frames of the crashing thread.

With that, we can focus the ahead-of-time work to the context the end user is most likely to actually look at, and search for; while keeping options open for the future.

I believe this is big enough to involve more people from the product side of things, to answer the following questions:

- Which parts of the product need source context?
- How much (lines of) source context?
- How does this tie into search-by-code?
- Do we need *all* the source context at once, or is it okay to lazily load some of it at a later time (favoring a hybrid approach)

text/0074-source-context-via-links.md
vaind:
```suggestion
- Portable-PDB source-link (.NET, see [this `symbolic` issue](https://github.com/getsentry/symbolic/issues/735))
- Debuginfod servers (we don’t support these yet)
- SourceMaps (either embedded sourcesContent or using individual source files)
```

vaind:
> A third option (not my first choice though) would be to have Sentry CLI download the files from the links and use them to create a source bundle, then upload the source bundle through the existing mechanism.

Yes, that occurred to me too, originally, but I've abandoned that quickly because it doesn't translate to debug info coming from dependencies (e.g. snuget) that will only be loaded on the server during symbolication

vaind:
Who can review & merge this? As is, it just describes what is in production at the moment.

Swatinem (approved):

PR #75, opened 2023-02-15
rfc(feature): Span thread information

JonasBa:
TODO. [Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/span-thread-id/text/0075-span-thread-id.md)


text/0075-span-thread-id.md
```diff
@@ -0,0 +1,35 @@
...
+As we have started to enrich profiling and performance UIs with data from the two products, we realized that we cannot precisely show what spans were executed on each thread (ex in profiling UI) or show what function frames were executed during a span (ex in performance UI). Having span thread id data would allow us to improve the user experience between the two products.
+
+# Supporting Data
+
+In profiling, we currently show the entire span tree regardless of what thread the user is viewing and in performance we default to showing the main thread. This is not very accurate and can lead to misleading our users.
+
+Metrics:
+No usage metrics, but this was requested by customers.
+
+# Options Considered
```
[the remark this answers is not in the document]
JonasBa: Yes, we would need to make this thread safe. I'm not familiar with the SDK's enough to be able to tell how much of an investment or effort this would take

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,35 @@
...
+
+Enrich spans with thread id information.
+
+# Motivation
+
+As we have started to enrich profiling and performance UIs with data from the two products, we realized that we cannot precisely show what spans were executed on each thread (ex in profiling UI) or show what function frames were executed during a span (ex in performance UI). Having span thread id data would allow us to improve the user experience between the two products.
+
+# Supporting Data
+
+In profiling, we currently show the entire span tree regardless of what thread the user is viewing and in performance we default to showing the main thread. This is not very accurate and can lead to misleading our users.
```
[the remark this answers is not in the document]
JonasBa: @Zylphrex should I just mark PR for review and you can directly add suggestions here?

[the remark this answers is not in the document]
JonasBa: Done, please feel free to add anything you want :)

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,35 @@
...
+
+Metrics:
+No usage metrics, but this was requested by customers.
+
+# Options Considered
+
+- Add thread_id to spans
+- Add thread_id and optional ended_thread_id (in case they differ) property to spans
```
AbhiPrasad, line 24: In OpenTelemetry they refer to this as `thread.id` and `thread.name`: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-thread-attributes

could we use the same naming conventions?

We'll also have to update https://develop.sentry.dev/sdk/performance/span-data-conventions/ with whatever we use.

AbhiPrasad: I assume this data will go under `span.data` and not a new top level attribute, which I think is better suited here.

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,35 @@
...
+
+# Drawbacks
+
+Spans can span (pun not intended) different threads, so if we wanted better precision, we would probably need to track which thread a span was started and ended on.
```
AbhiPrasad, line 28: OpenTelemetry's general approach here is that the thread that started a signal (metric, span) is what they usually just track - and I think that's a reasonable approach for us to take too.

JonasBa: @AbhiPrasad I agree with that. We are also somewhat early in this discovery so besides a few edge cases, it is hard to evaluate the actual benefit of cross thread tracking aka I dont know if I could comfortably justify it. Following what otel or other tools do seems like a great first step.

k-fish: Unless we have a strong case to deviate from OTEL, I wouldn't, for compatibility purposes alone. Irrespective of that, I think the start thread is sufficient as well, it's not that we can't add end-information later if we find it to be a critical need.

k-fish (commented):
Two things from me:

- Spans adding data will increase the size of transaction payload a little bit, you should probably ballpark it's frequency / size impact based on the data type (int I'm guessing)
- w.r.t the thread id collection, is the thread id only present w/ a profile for those spans? are we always collecting it? what's the overhead of fetching the thread id on each span for non-profiled transactions (or would we even want to?)

text/0075-span-thread-id.md
JonasBa: Yes, we also need to align here - right now profiling already collects a name as well, so this maybe becomes more a question for performance cc @k-fish, but would you benefit from having thread names available as well?

JonasBa:
> Two things from me:
> 
> * Spans adding data will increase the size of transaction payload a little bit, you should probably ballpark it's frequency / size impact based on the data type (int I'm guessing)
> * w.r.t the thread id collection, is the thread id only present w/ a profile for those spans? are we always collecting it? what's the overhead of fetching the thread id on each span for non-profiled transactions (or would we even want to?)

Yes, I 100% agree. I wanted to keep this open so folks from different teams can contribute with their thoughts before I move forward

text/0075-span-thread-id.md
k-fish: We might be able to use it to annotate the transaction summary to indicate threading happened, so possibly. We don't have any current plans to do that at the moment though.

JonasBa:
@AbhiPrasad @k-fish thanks for the feedback, I updated the RFC to be more precise and I added a couple of benchmarks.

Zylphrex (commented):
One comment here is do we care about the thread name as well? Or is that too much information to be collecting?

JonasBa:
> One comment here is do we care about the thread name as well? Or is that too much information to be collecting?

@k-fish brought this up in https://github.com/getsentry/rfcs/pull/75#discussion_r1110264996, I wanted to defer that decision to each product (we definitely need it for profiling), but it seems like it can be added later on. Collecting thread ids will likely determine how we collect thread names in the future, I dont see any strong reason why we would deviate from that tbh

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,91 @@
...
+
+The common data type for a thread identifier is uint64 (also the one we currently use in profiling) and one that we likely should be using.
+
+Suggestion:
+- collect thread.id on spans regardless if profiling is enabled or not
+- collect thread.id on spans only during an active profiling session
+
+
+Where do we store the data?
+There are two places where we could store thread information - either on the span data dict and using setData or by adding new fields to the span object. The former has the benefit of a prexisting API, we would only have to ensure the data is properly set while the later seems more involved and might have infrastructure implications. Since the data dict is public, we would probably want to ensure that users do not override the thread information or discard it if they do - it sounds like acceptable risk to me.
```
philipphofmann, line 37: For file I/O on the main thread we use `setData`. We agreed on the key `blocked_main_thread`, see https://github.com/getsentry/develop/blob/1c1b528a3442405dad51eedef6e69dbbec296d24/src/docs/sdk/performance/span-data-conventions.mdx?plain=1#L24

philipphofmann (commented):
This functionality can be helpful for multiple things. I like the idea.

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,91 @@
...
+
+Marking thread id as a recommended attributes also gives us flexibility in terms of only collecting the data while there is an active profiler session - I currently do not have strong reasoning to suggest that we should be collecting it all the time, if you think it could be beneficial, please add a comment.
+
+The common data type for a thread identifier is uint64 (also the one we currently use in profiling) and one that we likely should be using.
+
+Suggestion:
+- collect thread.id on spans regardless if profiling is enabled or not
```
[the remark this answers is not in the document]
JonasBa: @philipphofmann I think what we do now in profiling is that we send both thread name and tid and rely on the naming to indicate a "main thread". I briefly touched on that in the spec, but would it make sense for you to just set tid and name then as well or would you still want to send tid and main tid?

text/0075-span-thread-id.md
AbhiPrasad: All span data documented here: https://develop.sentry.dev/sdk/performance/span-data-conventions/

AbhiPrasad (approved):
I think this change makes a lot of sense to add.

This will be harder in the other sdks since we don’t have span start hooks outside of JS atm, so if an SDK wants to add this they may have to add hooks first to implement (then an thread integration for ex. would always add a thread ID on span start).

Given this is an implementation detail I'm in favor of adding merging this RFC (and updating the develop docs) and then tackling that afterwards with the respective SDK teams.

sl0thentr0py:
The Hub is already isolated to threads, so the `hub.scope.span` reference is always thread local. How can one thread start a span and another end it? I don't know how this applies in mobile where there's only global Hub but this is the behaviour in all web SDKs.
cc @AbhiPrasad

AbhiPrasad:
I think this is mostly for mobile right? Where one thread can start and another can finish?

sl0thentr0py (approved):

text/0075-span-thread-id.md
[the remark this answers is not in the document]
philipphofmann: I guess tid and name are more flexible. If users set thread names, tid and name would work int that case. tid and main tid would not.

philipphofmann:
> I think this is mostly for mobile right? Where one thread can start and another can finish?

Yes, exactly. On mobile, it's common to have some UI event on the main thread, then do some work on a background thread as fetching data via HTTP, and then update the UI again on the main thread.

philipphofmann:
@JonasBa, what's missing for merging this RFC? We also need span thread information for [Mobile Starfish](https://www.notion.so/sentry/Tag-spans-with-if-they-ran-on-the-main-thread-or-not-5f3ebd6377514cad8962bed152db8502). It would be great to get it merged 😃.

JonasBa:
@philipphofmann we haven agreed where we should store this data on the span model - it seems like span.data could be the right place, any thoughts on that?

AbhiPrasad:
Let's please use `span.data` and the same naming scheme that opentelemetry does. I'm working on https://github.com/getsentry/rfcs/pull/116 which should hopefully make it clearer to future contributors where to place this kind of data.

JonasBa:
@AbhiPrasad Yes, lets go with future compatibility here. I would propose we standardize on the flattened `thread.id: string` and `thread.name: string` on span data.

JonasBa:
Updated the RFC with the conclusion. In case someone has any final comments/thoughts please comment them, else I'll mark this as approved and merge it.

text/0043-frame-module.md
AbhiPrasad, line 1: Can we delete this file?

JonasBa: Ugh, idk how that creeped in, yes.

text/0075-span-thread-id.md
```diff
@@ -0,0 +1,109 @@
...
+  "data": {
+    "thread.id": "unique_thread_identifier",
+    "thread.name": "thread_name"
+  }
+}
+```
+
+# Unresolved questions
+
+- Should breadcrumbs also track thread ids?
```
AbhiPrasad, line 109: the answer is yes 😄

AbhiPrasad (approved):
If you could make a PR to https://develop.sentry.dev/sdk/performance/span-data-conventions/ would be awesome!

PR #77, opened 2023-02-25
Issue Platform RFC

wedamija:
This is a draft for the issue platform rfc. Looking for feedback on what else needs to be added - happy to add diagrams, motivations for certain decisions and so on.

### DIFF
--- a/text/0073-issue-platform.md
+++ b/text/0073-issue-platform.md
@@ -1,0 +1,85 @@
+- Start Date: YYYY-MM-DD
+- RFC Type: informational
+- RFC PR: <link>
+- RFC Status: draft
+
+# Summary
+
+This rfc is intended to provide a high level via of the issue platform, it's 
+capabilities, how to interact with it and plans for the future.
+
+# Motivation
+
+The primary purpose of the issue platform is to allow different teams at Sentry to 
+detect problems relevant to our users in their data, and create new issues types from 
+that data.
+
+# High Level Concepts
+
+ - Event: A general term for data sent from SDKs to Sentry. This includes errors, transactions, profiles, replays, etc.
+ - Issue: An aggregation of events representing a unique problem we detected in the user’s system.
+ - Detector: A logical unit that analyzes data looking for problematic patterns and outputs found problems.
+ - Issue Occurrence: An event specifically associated with an issue. These are raised from detectors and sent to Sentry via the occurrences api.
+ - Fingerprint: An identifier of a unique occurrence. It’s used for grouping occurrences into issues.
+ - Issue Category: The broad category that an issue belongs to. `Error`, `Performance` and `Profile` are examples of this.
+ - Issue Type: The specific type of an issue. `N + 1 DB query` is an example of this. Type implies category.
+
+
+# Ingestion Schema
+
+Ingestion Schema
+Ingestion occurs via Kafka. The message format looks like
+
+ - id: str - A uuid representing the issue occurrence id.
+ - project_id: int.
+ - fingerprint: Sequence[str] - The fingerprint to uniquely identify a specific issue
+ - issue_title: str - The title of the occurrence. Will be used as issue title.
+ - subtitle: str - Used as the sub header on the issue.
+ - resource_id: str | None - An optional id that we can use to link to an external system
+ - evidence_data: Mapping[str, Any] - An arbitrary json blob of data for use with the frontend to render any custom components necessary for this issue type.
+ - evidence_display: A list of data to display in the ui and notifications. This will typically be displayed in a tabular format. Each entry is a dict in format
+   - name: str: Name of the value to display in the table
+   - value: str: Value to display in the table
+   - important: bool - Up to one piece of data in evidence_display should be marked as important. This entry will be used in notifications where space is constrained.
+ - type: int - The issue type that this occurrence belongs to. Must be registered as a dataclass like so: https://github.com/getsentry/sentry/blob/5fd8242aca575c84314f1b6744be2ec1d2e9e471/src/sentry/issues/grouptype.py#L143-L148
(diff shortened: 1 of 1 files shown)


text/0077-issue-platform.md
```diff
@@ -0,0 +1,85 @@
...
+Ingestion occurs via Kafka. The message format looks like
+
+ - id: str - A uuid representing the issue occurrence id.
+ - project_id: int.
+ - fingerprint: Sequence[str] - The fingerprint to uniquely identify a specific issue
+ - issue_title: str - The title of the occurrence. Will be used as issue title.
+ - subtitle: str - Used as the sub header on the issue.
+ - resource_id: str | None - An optional id that we can use to link to an external system
+ - evidence_data: Mapping[str, Any] - An arbitrary json blob of data for use with the frontend to render any custom components necessary for this issue type.
+ - evidence_display: A list of data to display in the ui and notifications. This will typically be displayed in a tabular format. Each entry is a dict in format
```
markstory, line 40:
```suggestion
 - evidence_display: A list of data to display in the ui and notifications. This will typically be displayed in a tabular format. Each entry is a dict using the format:
```

text/0077-issue-platform.md
```diff
@@ -0,0 +1,85 @@
...
+   - name: str: Name of the value to display in the table
+   - value: str: Value to display in the table
```
markstory, line 42:
```suggestion
   - name: str -  Name of the value to display in the table
   - value: str - Value to display in the table
```

text/0077-issue-platform.md
```diff
@@ -0,0 +1,85 @@
...
+   - important: bool - Up to one piece of data in evidence_display should be marked as important. This entry will be used in notifications where space is constrained.
+ - type: int - The issue type that this occurrence belongs to. Must be registered as a dataclass like so: https://github.com/getsentry/sentry/blob/5fd8242aca575c84314f1b6744be2ec1d2e9e471/src/sentry/issues/grouptype.py#L143-L148
+ - detection_time: float - Timestamp representing when the occurrence was detected
+ - level: str - The error level we want to display for the occurrence.
+
+One (but not both) of these should also be passed
+ - event_id: str - A uuid representing the event id. Pass this if you want to use an event that has already been stored into nodestore.
+ - event: dict - A dictionary that matches our event format (https://develop.sentry.dev/sdk/event-payloads/). Pass as much of this as makes sense for your issue type. Use this method if your event isn’t already in nodestore.
+
+ # General Data Flow
```
markstory, line 52:
```suggestion
# General Data Flow

```

markstory (commented):
Found a few formatting discrepancies/typos.

text/0077-issue-platform.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+# Future Plans:
+- Port performance issues to the issue platform. This gives us more consistency, and will improve query performance since we no longer need to query the transactions table, which contains mostly irrelevant data.
+- Improve the release process for new issue types. This is via a policy layer, that will automatically set up ways to release new issue types to LA, EA and GA and save adding a bunch of boilerplate code.
+
+# Things we’ve cut:
+Errors. We originally had plans to send errors through the issue platform to centralise everything. We’ve decided against this because it seems like a high risk task without much reward. The primary reason we wanted to send errors through the issue platform was so that we could split up detection from ingestion. We might still want to separate detection from ingestion in the future, and this doesn't necessarily have to involve the issue platform, or could involve a separate path through issue platform ingestion that still writes to the same errors dataset.
```
fpacifici, line 83: > The primary reason we wanted to send errors through the issue platform was so that we could split up detection from ingestion.

Does it matter much for errors? Each event is also an occurrence so there is no real difference between the two.

text/0077-issue-platform.md
```diff
@@ -0,0 +1,85 @@
...
+   - name: str: Name of the value to display in the table
+   - value: str: Value to display in the table
+   - important: bool - Up to one piece of data in evidence_display should be marked as important. This entry will be used in notifications where space is constrained.
+ - type: int - The issue type that this occurrence belongs to. Must be registered as a dataclass like so: https://github.com/getsentry/sentry/blob/5fd8242aca575c84314f1b6744be2ec1d2e9e471/src/sentry/issues/grouptype.py#L143-L148
+ - detection_time: float - Timestamp representing when the occurrence was detected
+ - level: str - The error level we want to display for the occurrence.
+
+One (but not both) of these should also be passed
+ - event_id: str - A uuid representing the event id. Pass this if you want to use an event that has already been stored into nodestore.
+ - event: dict - A dictionary that matches our event format (https://develop.sentry.dev/sdk/event-payloads/). Pass as much of this as makes sense for your issue type. Use this method if your event isn’t already in nodestore.
```
fpacifici, line 50: Do we actually store this as a dictionary in clickhouse? This would be very inefficient to query.

text/0077-issue-platform.md
mitsuhiko: The plan now however from my understanding is to move errors onto issue platform. This is something I am at least very interested in as it would allow us to test different versions of grouping algorithms without impacting existing groups.

fpacifici: Can we keep the storage where it is? that does not impact the grouping algorithm.

mitsuhiko: @fpacifici I'm not involved with the project to move errors to issue platform. There are two things we definitely need to do:

a) we need to enable that an error can exist in two issues. That the existing errors data set does not really support due to it's n:1 relationship
b) from a programming point of view we need to have a reasonable level of abstraction between all the issue types to drive future features such as detecting issue state.

The general assumption that I had is that everything moves over to this new issue world.

PR #79, opened 2023-03-07
rfc(feature): Exception Groups

mattjohnsonpint:
Design for changes required to support exception groups.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/exception-groups/text/0079-exception-groups.md)


text/0079-exception-groups.md
```diff
@@ -0,0 +1,348 @@
...
+has many disadvantages.
+
+Pros:
+- SDK changes are straightforward.
+- Already implemented in the .NET SDK.
+- Doesn't require any protocol, ingest, or UI changes.
+
+Cons:
+- Doesn't accurately represent the exception group.
+- Implies a causal relationship between sibling exception that doesn't exist.
```
jan-auer, line 175: How would you expect the error event to be grouped issues for Options 2 and 4?

text/0079-exception-groups.md
```diff
@@ -0,0 +1,348 @@
...
+# Supporting Data
+
+- [Python customer feedback](https://github.com/getsentry/sentry/issues/37716)
+- [.NET customer feedback](https://github.com/getsentry/sentry-dotnet/issues/270)
+
+# Options Considered
+
+## Option 1
+
+Do nothing.
```
jan-auer, line 149: This is an outright question: Would it make sense to add options 1b and 1c that contain:
- SDK reports each exception in a group as separate error event; no change to protocol.
- Advice to developers (customers) that they should manually inject a `capture_exception` at the places where the groups are caught and capture them separately. This is particularly to avoid cases where a group is nested.

Additionally, can it happen that _one_ of the exceptions in the group can be viewed as "causal" for a chain whereas other exceptions are just additional baggage? In this case, we would definitely need SDK-side mapping to determine which exception should be separated out.

text/0079-exception-groups.md
```diff
@@ -0,0 +1,348 @@
...
+
+Sentry needs a way to capture exception groups, and present them in a meaningful way.
+Simply raising the exception group as a single issue is insufficient.
+
+# Background
+
+## About Exception Groups
+
+An exception group is an exception unto itself.  It is caused by the exceptions in the group,
+but there is no implied causal relationship between those exceptions.
```
jan-auer, line 32: If exceptions in a group have no causal relationship, the exceptions in the group are likely to be viewed as separate issues in Sentry. In this case, is there some benefit we would gain by reporting them together and showing the connection? This could be elaborated on in the intro.

mattjohnsonpint: I experimented with sending separate events, and it's not a viable option.  I'll add details about it to the RFC though, so that it's understood.  Thanks.

text/0079-exception-groups.md
mattjohnsonpint: Yeah, that doesn't work out very well in practice.  I'll add some information and reasons why to the RFC.

mattjohnsonpint: On the second question, if only one is the cause then they're only be one exception in the group to begin with.  However, usually an exception group is raised if *any* of a list of tasks fails. Thus it can be possible that one of the exceptions is the true cause, and the others are noise created by cancelling a running task (if they're aborted). There's really no way to tell - so we have to treat all of them as potential causes.

mattjohnsonpint: Actually, I forgot one thing.  At least for .NET an `AggregateException` will never contain an `OperationCanceledException`.  So if say two tasks are being waited upon, and one of them throws some exception and the other is cancelled, only the one exception will be in the group.  And if _both_ tasks are cancelled, then an `OperationCanceledException` will be thrown _instead of_ an `AggregateException`.

So ultimately, this aligns with the idea of all exceptions in an exception group _together_ are the cause for the exception group.  There isn't a case where some are additional baggage.  Not really.

Though also, sometimes you might not care about the others.  Even if they're meaningful, they might be all be similar.  We could have an option to discard all but one of them.  I'll add a few more sub-options to the list.  Thanks.

text/0079-exception-groups.md
mattjohnsonpint: Option 2 is what .NET does today (and shouldn't, IMHO).  I don't really want to pursue this, but included it for completeness.  By default it takes one exception out of the group and only captures that.  All others are flattened below it (incorrectly).  So the issue is titled based on that exception - but it's grouped on the entire set of exceptions.  (Oops - one of the cons on grouping I accidentally put under option 1 instead of 2. Will fix.)

Also, .NET provides `SentryOptions.KeepAggregateException` which is `false` by default, but when set `true` will include the exception group.  So in that case, the title of the issue is the meaningless message about the exception group, and that is also used in issue grouping.

Both are bad.

mattjohnsonpint: For Option 4, grouping would be similar to Option 3, but would have to take nested exception groups into account somehow.  I'm really not sure how, TBH.  Maybe a strong reason not to do Option 4.

mattjohnsonpint: The RFC has been changed considerably.  The revised proposal does not require any changes to how grouping works.  Thanks.

text/0079-exception-groups.md
mattjohnsonpint: The RFC has been changed considerably. The revised proposal describes the "SDK Splitting" approach under the "Other Options Considered" section.  Thanks.

text/0079-exception-groups.md
mattjohnsonpint: The RFC has been changed considerably.  The connection between exceptions in an exception group is only important with regard to the exception group tree itself.  Linking separate issues together in Sentry is less important.  Thanks.

mattjohnsonpint:
Please note that the RFC has been revised considerably since it was first created.  Please re-read the entire RFC and leave feedback on the new proposal as you see fit.  Thanks!

text/0079-exception-groups.md
```diff
@@ -0,0 +1,558 @@
...
+- They may be representing different quantities of the _same_ issue, and the quantity should not affect issue grouping in Sentry either.
+- Sending them all in one list would imply a causal relationship between `D` and `E` (in that example), which does not exist.
+
+However, we don't want to completely lose track of these other exceptions either.  Thus they should still be sent
+with the event, but in another section instead.  We will add it to the `extras` property of the event.
+
+The format should be as follows:
+
+```json
+"extras": {
```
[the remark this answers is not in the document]
mattjohnsonpint: Yes, it should be `extra`, not `extras`.  I will fix that - thanks.  As far as I can tell, it is not deprecated.  Is there somewhere that says it has been?

`debug_meta` would not be visible in the UI presently.  We should pick some place that the data is visible even for older versions of self-hosted Sentry.  I also think it could affect processing in unexpected ways.

`contexts` would be ideal, but it has half of the [size limit](https://develop.sentry.dev/sdk/data-handling/#variable-size) of `extra`, which I can see being an issue for larger exception groups - especially if they contain stack traces.

text/0079-exception-groups.md
```diff
@@ -0,0 +1,558 @@
...
+    - To conserve space, it should be omitted for other types of exceptions. (The default is `false` when omitted.)
+  - The `items` array is _optional_.  It is an array of additional exception group objects.
+- The number of nested objects is not constrained.  However, data may be truncated if the total size exceeds 16kB.
+
+_Note: We chose `extras` over `contexts`, because items in `contexts` are limited to 8kB, whereas items in `extras` are limited to 16kB, per [documentation](https://develop.sentry.dev/sdk/data-handling/#variable-size)._
+
+### Additional SDK requirements
+
+The SDK should set a pre-formatted message on the event (via `logentry.formatted`) to the following text:
+> _"This issue was raised as part of an exception group. See the Additional Data section for details."_
```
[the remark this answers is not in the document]
mattjohnsonpint: That's an option, though it does mean adding some minor work for relay.  @jan-auer

text/0079-exception-groups.md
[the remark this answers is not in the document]
mattjohnsonpint: Update: I just spent some time discussing this with @mitsuhiko, and he agrees that we should _not_ use `extras`.

I'm going to explore yet another alternative, which is to use `exceptions` with new fields on `mechanism`.  I'll update the RFC with the results, if that works out.

Thanks.

text/0079-exception-groups.md
```diff
@@ -0,0 +1,558 @@
...
+
+The SDK should also provide an option to control whether top-level exception groups are kept with the list of exceptions, or stripped away.
```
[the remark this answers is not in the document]
mattjohnsonpint: That would stop events from being sent which contained an exception group type.  This option trims the list of exceptions, but the event with the rest of the exception still gets sent.

I'll see if there's another approach that would work, but the intent here is that the user have control over whether the issue is titled and grouped by the exception group or the first-issue.

text/0079-exception-groups.md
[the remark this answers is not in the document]
mattjohnsonpint: Major rewrite complete.  Have another read.  Thanks. 😉

mattjohnsonpint:
Yet another major rewrite of this RFC.  I believe the plan is solid now 🤞 😅

antonpirker (approved):

AbhiPrasad (approved):
This new approach looks good to me, and makes sense on the SDK side. Thanks for keeping this going @mattjohnsonpint!

lforst:
An a posteriori learning: Extending an interface, which in itself is optional but has required fields (i.e. `mechanism`) is really painful. What if the exception I want to put an `exception_id` on doesn't have a mechanism yet? I would now have to put nonsensical default values inside `mechanism` to conform to the protocol again.

Maybe we should have put all the extra data onto the exception interface directly 🤔

PR #81, opened 2023-03-21
Reliable JavaScript/SourceMap processing via `DebugId`

Swatinem:
We want to make processing / SourceMap-ing of JavaScript stack traces more reliable.
To achieve this, we want to uniquely identify a (minified / deployed) JavaScript file using a `DebugId`.
The same `DebugId` also uniquely identifies the corresponding SourceMap.
That way it should be possible to _reliably_ look up the SourceMap corresponding to
a JavaScript file.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/sourcemap-debugid/text/0081-sourcemap-debugid.md)

### DIFF
--- a/text/0081-sourcemap-debugid.md
+++ b/text/0081-sourcemap-debugid.md
@@ -48,8 +48,6 @@ a lot more reliable.
 
 # Supporting Data
 
-TODO: please fill in the gaps here!
-
 Sentry has used the `release (+ dist) + abs_path` solution for quite some time and found it inadequate.
 A lot of events are not being resolved correctly due to these mismatches, and problems with source-mapping are very
 common in customer-support interactions. Up to X% of customer-support questions are related to SourceMaps.
@@ -93,34 +91,23 @@ Different approaches and examples are shown below, but in summary, we would like
 deterministic for a build. Building the exact same project and original code should always result in the same assets.
 This means the same minified JS files (with embedded `DebugId`), and the same SourceMap file.
 
+## Deterministic
+
 As long as the `DebugId` is deterministic, it does not matter how it is being derived. Different tools will use different
 methods. And a `DebugId` may differ from one tool to the next, just as the exact output files may differ between tools,
 and even tool versions.
 
-### Based on JavaScript Content-hash
+We do not want content hashes to be only based on the minified JavaScript file to also pick up on the original source
+cahnges.  For a source hash only approach the solutions outlined in [RFC 85](https://github.com/getsentry/rfcs/pull/85)
+apply.
 
-This creates a new `DebugId` by hashing the contents of the (minified) JavaScript file.
-
-**pros**
-
-- Is fully reproducible. The same JavaScript file will always have the same `DebugId`.
-- Works well with existing caching solutions.
-
-**cons**
-
-- Increases overhead in server-side SourceMap processing, as one file can potentially be included in multiple _bundles_.
-  See [_What is an `ArtifactBundle`_](#what-is-an-artifactbundle) below.
-- A difference in a source file might not be reflected in the JavaScript file. An example of this might be changes to
-  whitespace, comments, or code that was dead-code-eliminated by bundlers.
-
-### Based on SourceMap Content-hash
-
-This creates a new `DebugId` by hashing the contents of the SourceMap file. The reasoning for this is primarily
-motivated by a drawback of the previous option.
-It is possible that non-essential changes to the source file, such as whitespace or comments, will be completely
-removed by a minifier, and will result in an identical minified output file.
-These changes would however create different mappings in the SourceMap file, making it possible to resolve to the
-correct original source.
+Instead we recommend to generate a new `DebugId` by hashing the contents of the
+SourceMap file. The reasoning for this is primarily motivated by a drawback of
+the previous option.  It is possible that non-essential changes to the source
+file, such as whitespace or comments, will be completely removed by a minifier,
+and will result in an identical minified output file.  These changes would
+however create different mappings in the SourceMap file, making it possible to
+resolve to the correct original source.
 
 It is also possible to use a combined content-hash of both the (minified) JavaScript file, and its corresponding SourceMap.
 
@@ -133,7 +120,7 @@ It is also possible to use a combined content-hash of both the (minified) JavaSc
 
 - Does lead to slightly more cache invalidation.
 
-### Random `DebugId`
+### Random
 
 This option would create a new random `DebugId` for each file, on each build.
 
@@ -386,12 +373,7 @@ to call this at the appropriate time depending on their unique build pipeline.
 
 In this scenario, injection happens at the time of `sentry-cli upload`, and will also modify the files at that time.
 
-**pros**
-
-- Makes sure that assets uploaded to Sentry have a `DebugId`.
-- No additional command and invocation needed.
-
-**cons**
+This however is rejected due to the many downsides:
 
 - Violates expectations of file immutability and integrity.
 - Does not work with bundlers that integrate fingerprinting.
@@ -417,8 +399,6 @@ access and modification of a chunks `code` and `map` (SourceMap) output.
 This hook runs before the `augmentChunkHash` and `generateBundle` hooks which are responsible for fingerprinting and
 generating the _final_ output for each chunk.
 
-TODO: further investigation and experimentation for this is needed
-
 #### Webpack
 
 Webpack documentation for plugin hooks is not as extensive, and there is no broad overview of the internal pipeline and
@@ -429,10 +409,6 @@ It might be possible to use the [`processAssets`](https://webpack.js.org/api/com
 for this purpose. Documentation mentions the `PROCESS_ASSETS_STAGE_DEV_TOOLING` phase which is responsible for
 extracting SourceMaps, or the `PROCESS_ASSETS_STAGE_OPTIMIZE_HASH` which looks to be responsible for generating the
 final fingerprint of an asset.
-
-TODO: further investigation and experimentation for this is needed
-
-#### TODO: other popular bundlers and build tools
 
 #### Implementation in bundlers directly
 
@@ -456,6 +432,26 @@ their `"sourcesContent"`.
 Additionally, to simplify `DebugId` generation, and to reduce implementation complexity, we strongly encourage to use
 "flat" SourceMaps, as opposed to "indexed" SourceMaps, which themselves contain other nested SourceMaps.
 
+## Protocol Changes
+
+The Sentry protocol gains a new debug image type called `sourcemap` which can be used to refer to source maps
+by `DebugId` in the following format:
+
+```json
+{
+  "images": [
+    {
+      "type": "sourcemap",
+      "code_file": "http://example.com/file.min.js",
+      "debug_id": "dd6a14d1-fd08-48a5-b827-b87d8e9856bc"
+    }
+  ]
+}
+```
+
+The two keys defined are `code_file` which needs to match the frame's `abs_path` and `debug_id` which needs to
+match the `DebugId` of the file.
+
 # Drawbacks
 
 The main drawback is that this might feel like an invasive change to the JavaScript ecosystem. It is a huge implementation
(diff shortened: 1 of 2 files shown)


mitsuhiko:
Ad unresolved questions: I think the `debug_id` reference in the sourcemap is something that makes quite a bit of sense to change.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+Current JavaScript stack traces include the absolute path (called `abs_path`) of each stack frame. It should be possible
+to load and inspect that file at runtime whenever an error happens.
+
+**pros**
+
+- Does not require injecting any _code_ into the JavaScript files.
+
+**cons**
+
+- Might incur some async fetching / IO when capturing an Error. Though any `abs_path` in the stack trace should be cached already.
```
lforst, line 156: This doesn't sound like a good idea to me. Mostly gut feeling but some other things come to mind:

- Resource could be large
- There might be delay in fetching the resource causing racing situations with navigations and so on
- (I'll add additional reasons here when they come to mind)

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+
+In a shiny utopian future, Browsers would directly expose builtin APIs to programmatically access each frame of an `Error`s stack.
+This might include the absolute path, the line and column number, and the `DebugId`.
+Though the reality of today is that each browser has its own text-based `Error.stack` format, which might even give
+completely different line and column numbers across the different browsers.
+No programmatic API exists today, and might never exist. At the very least, widespread support for this is years away.
+
+It is therefore necessary to extract this `DebugId` through other means.
+
+#### Reading the `//# debugId` comment when capturing Errors
```
lforst, line 145: Can you add some example (pseudo code) of how you would create the mapping of `Frame -> Debug ID` when an error is captured? I struggle to understand how the pieces would fit together.

My current suspicion is that a comment is not enough for runtime resolution but I will be glad if I am proven wrong.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+**pros**
+
+- Does not require any async fetching at time of capturing an `Error`.
+
+**cons**
+
+- It does however require parsing of the `Error.stack` at time of capturing the `Error`.
+
+An alternative implementation might use the [`import.meta.url`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import.meta)
+property. This would avoid capturing and post-processing an `Error.stack`, but does require usage of ECMAScript Modules.
```
lforst, line 181: > but does require usage of ECMAScript Modules

IMO this is a deal breaker. At least while IE is still somewhat relevant.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+This snippet adds a complete `Error.stack` to a global called `_sentryDebugIds`.
+Further post-processing at time of capturing an `Error` is required to extract the `abs_path` from that captured stack.
+
+**pros**
+
+- Does not require any async fetching at time of capturing an `Error`.
+
+**cons**
+
+- It does however require parsing of the `Error.stack` at time of capturing the `Error`.
```
lforst, line 178: This is a low-impact con, as we are already parsing the stack trace when it is captured.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+    bundle --> minify[Minify bundled chunk]
+    minify --> fingerprint[Fingerprint minified chunks]
+    minify --> sentry[Upload release to Sentry]
+    fingerprint --> upload[Upload assets to CDN]
+    upload --> propagate[Wait for CDS assets to propagate]
+    fingerprint --> deploy[Deploy updated asset references]
+    propagate --> deploy
+```
+
+In this example, assets are _fingerprinted_, and after being fully propagated
```
lforst, line 213: What does "fingerprinting" mean here?

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+### Injection at `sentry-cli upload` time
+
+In this scenario, injection happens at the time of `sentry-cli upload`, and will also modify the files at that time.
+
+**pros**
+
+- Makes sure that assets uploaded to Sentry have a `DebugId`.
+- No additional command and invocation needed.
+
+**cons**
```
lforst, line 273: We should add as a con here that this may easily break the integrity of files (CSP and subresource integrity), as there is the possibility that users calculate and store the files' digest before running `sentry-cli upload`.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+
+- Does not work with bundlers that integrate fingerprinting.
+- Does not work in build pipelines where `sentry-cli upload` is not in the main deployment path.
+
+### Injection via bundler plugins
+
+Here, we would build `DebugId` injection right into the various JavaScript bundlers. This can happen with a third-party
+plugin at first, and might move into the core bundler packages once there is enough community buy-in for `DebugId`s.
```
lforst, line 281: > This can happen with a third-party
plugin at first, and might move into the core bundler packages once there is enough community buy-in for `DebugId`s.

I think this can go in right away (even if it is an experimental option). I think we should remove this sentence. It's unnecessary.

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+burden, and might not be received positively by neither customers nor the wider JS tools ecosystem.
+
+Especially injecting a piece of JavaScript into every production asset might alienate some users.
+
+The effectiveness and success of this initiative needs to be proved out first, and is not guaranteed.
+
+# Unresolved questions
+
+- ~~Why do we call the new SourceMap field `debug_id` and not `debugId`?
+  All existing fields in SourceMaps are camelCase, and so is the general convention in the JS ecosystem.~~
```
lforst, line 333: +1

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+There is two options of choosing a `DebugId`: Making it completely random, or make it reproducible by deriving it from
+a content hash.
+
+### Based on JavaScript Content-hash
+
+This creates a new `DebugId` by hashing the contents of the JavaScript file.
+
+**pros**
+
+- Is fully reproducible. The same JavaScript file will always have the same `DebugId`.
```
marandaneto, line 84: Is the minification always deterministic? Do 2 builds of the same code always generate the same source maps file so that the same hash can be computed?

text/0081-sourcemap-debugid.md
```diff
@@ -0,0 +1,362 @@
...
+**pros**
+
+- Simpler server-side SourceMap processing, as one `DebugId` is only included in a single _bundle_, and that one bundle
+  can serve multiple stack frames for multiple files of the same build.
+
+**cons**
+
+- Completely breaks the concept of _caching_, as every file is unique for every build.
+
+## How to inject the `DebugId` into the JavaScript file?
```
marandaneto, line 119: Do we actually need to do this? for example, the `proguard` cmd takes a `uuid` if required and makes the association.
https://github.com/getsentry/sentry-android-gradle-plugin/blob/a8fc82a9d45fdc0208b276fa18eb377bcaea440f/plugin-build/src/main/kotlin/io/sentry/android/gradle/tasks/SentryUploadProguardMappingsTask.kt#L85-L86

marandaneto:
Have we considered languages/frameworks that are transpiled to JS? Such as Flutter Web (Dart -> JS), you don't have much control over the transpilation and modification steps, nor you do build hooks, not sure how the `inject` would work here.

text/0081-sourcemap-debugid.md
Swatinem: Well, the UUID of the proguard mapping is embedded into the shipped app though, so the SDK can transmit that with every event.
This is the same concept, just that doing it for JS makes it a ton more complex.

Swatinem:
@marandaneto 
I have not considered Dart / Flutter at all so far, and I do not know anything about it either. Can you give more details here? How does the compilation pipeline look like? Does Dart use SourceMaps? If so, how? Do developers usually combine that pipeline with other JS-native build tools/bundlers?

marandaneto:
> @marandaneto I have not considered Dart / Flutter at all so far, and I do not know anything about it either. Can you give more details here? How does the compilation pipeline look like? Does Dart use SourceMaps? If so, how? Do developers usually combine that pipeline with other JS-native build tools/bundlers?

Yeah Flutter web spits out source maps, and our symbolication works except for a few optimizations on the Dart side (as of now not a problem at all).
I guess it's possible to combine with other build tools/bundlers but it's not a common case at all, nor I have seen that, nor is documented, but maybe technically possible, I guess we don't need to support this at all at least as of now.
People just do `flutter build web --source-maps` and the magic happens.
Not sure which tooling flutter uses for source maps generation, would need to dig into the [source code](https://github.com/flutter/flutter/blob/4a1002e2e25543c370dc227c4f6162cf908b789c/packages/flutter_tools/lib/src/web/compile.dart#L29).

text/0081-sourcemap-debugid.md
mitsuhiko: We also defined a process by which the proguard file comes to a deterministic debug id through generating a SHA1 hash from the content hash.

text/0081-sourcemap-debugid.md
marandaneto: I can confirm it is deterministic.
`flutter build ios --split-debug-info=symbols --obfuscate --extra-gen-snapshot-options=--save-obfuscation-map=mapping.json` spits out the very same `mapping.json`.

text/0081-sourcemap-debugid.md
marandaneto: We just need to be sure that the proguard file (mapping) is also deterministic, the proguard tooling changed multiple times (now R8), Java to Kotlin, etc, better to confirm, I asked Android folks to test it out with the recent tooling at least.

text/0081-sourcemap-debugid.md
the-spyke: I'd rather build for the future, as when this spec become widespread the IE will be no more.

the-spyke:
Could be also `hash(minifiedContent + sourceMap)`

text/0081-sourcemap-debugid.md
lforst: This is an implementation detail anyhow so instead of building for the future I'd rather build for compatibility right now.

text/0081-sourcemap-debugid.md
markushi: I did a quick test for R8 + Android, rebuilding the same source from scratch multiple times and consistenly received the same mapping file:
````
% rm -rf build app/build
% ./gradlew :app:assembleRelease --no-build-cache --no-daemon
% md5 app/build/outputs/mapping/release/mapping.txt 
MD5 (app/build/outputs/mapping/release/mapping.txt) = e2ee64df9b51983445ffac36a4da455c
````
But this should be taken with a grain of salt, e.g. f-droid reports that in some cases R8 is not deterministic: https://f-droid.org/docs/Reproducible_Builds/#r8-optimizer

Comment (markushi) [reply text/0081-sourcemap-debugid.md:125]: Side note: I don't think we're using this right now but current versions of R8 generate a SHA-256 hash of the mapping payload and put it into the header of the mapping.txt file
````
head -n 10 app/build/outputs/mapping/release/mapping.txt
# compiler: R8
# compiler_version: 3.3.83
# min_api: 24
# common_typos_disable
# {"id":"com.android.tools.r8.mapping","version":"2.0"}
# pg_map_id: 0f71627
# pg_map_hash: SHA-256 0f71627e8e144f6a30bf2c32bef474bd7ab4f3173094061774e4c0722ebc8772
android.app.AppComponentFactory -> android.app.AppComponentFactory:
# {"id":"com.android.tools.r8.synthesized"}
    void <clinit>() -> <clinit>
````

File: text/0081-sourcemap-debugid.md (hunk 3 above)
Comment (Swatinem) [reply text/0081-sourcemap-debugid.md:272]: Once this is being implemented as bundler plugins, its possible to also make this configurable, with appropriate defaults. For example, if a bundler is configured to output ESM, it might inject the more compact ESM snippet, falling back to the other one in case of CommonJS/UMD output.

File: text/0081-sourcemap-debugid.md (hunk 10 above)
Comment (Swatinem) [reply text/0081-sourcemap-debugid.md:190]: Ah nice, we might even use that if it is also being exposed anywhere inside of apps.

The concept of file identity is definitely also a concern for proguard / java and other ecosystems, however I would like to focus discussions in this RFC to JS/SourceMaps :-)

**Comment by lforst** [comment]
> Could be also `hash(minifiedContent + sourceMap)`

Don't know about this, since in some cases we don't have either `minifiedContent` or `sourceMap` available. (e.g. when injecting during bundler builds)

**Comment by iambriccardo** [comment]
Considering that our goal is to capture change as defined by any change in the non-minified source, the hash of the source map is our best bet. The only caveat I am thinking of is if there is the possibility that `hash(sm1) != hash(sm2)` if `sc1 == sc2` where `sm` is the source map and `sc` is the non-minified source. An example of this is if the bundler omits/adds some information in the source map for some reason.

**Comment by mitsuhiko** [comment]
Auxiliary options with source hashes: [Original Spec from Edge](https://github.com/MicrosoftEdge/DevTools/blob/main/explainers/SourceHash/explainer.md)

## Runtime Support

The global `Error.prepareStackTrace` can be used on modern Chrome versions
to look up the script hashes from the stack frames.  This can be used to
create a script URL to SHA256 hash.

```javascript
const originalPrepareStackTrace = Error.prepareStackTrace;
Error.prepareStackTrace = (err, callsites) => {
  let scriptHashes = {};
  for (var callsite of callsites) {
    if (callsite.getScriptHash) {
      scriptHashes[callsite.getFileName()] = callsite.getScriptHash();
    }
  }
  err.scriptHashes = scriptHashes;
  if (originalPrepareStackTrace) {
    return originalPrepareStackTrace(err, callsites);
  } else {
    return err.stack;
  }
};
```

## Source Map Changes

In source maps a new key would have to be added that holds that hash (eg: `fileHash`):

```json
{
  "file": "myfile.min.js",
  "fileHash": "2bb80d537b1da3e38bd30361aa855686bde0eacd7162fef6a25fe97bf527a25b"
}
```

Before bundling up the script files, the bundler can also store the script
hashes with the minified files as well as the source maps that go along
with them:

```json
{
  "files": {
    "files/_/_/foo.min.js": {
      "url": "~/foo.min.js",
      "type": "minified_source",
      "headers": {
        "script-hash": "2bb80d537b1da3e38bd30361aa855686bde0eacd7162fef6a25fe97bf527a25b",
        "sourcemap": "foo.sourcemap.json"
      }
    },
    "files/_/_/foo.sourcemap.json": {
      "url": "~/foo.sourcemap.json",
      "type": "sourcemap",
      "headers": {
        "script-hash": "2bb80d537b1da3e38bd30361aa855686bde0eacd7162fef6a25fe97bf527a25b"
      }
    }
  }
}
```

This system can work alongside debug-ids.

Swatinem:
Do you know if the `SourceHash` proposal still takes suggestions? I would recommend to be able to explicitly state the hash algorithm, similar to https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#pdb-checksum-debug-directory-entry-type-19 to make it forward compatible to changing the hash algorithm.
In your example `SourceBundle` manifest, you have prefixed the `digest` with the algorithm as well. And that example snippet is missing the corresponding hash for the SourceMap entry.

mitsuhiko:
@Swatinem as the source hash proposal is already implemented in Chrome I'm not sure how many changes will happen there.

I moved the source hash proposal from my comment here: https://github.com/getsentry/rfcs/pull/85

PR #82, opened 2023-03-24
rfc(informational): Combined Replay Envelope Item

JoshFerge:
[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/combined-replay-envelope-item/text/0082-combined-replay-envelope-item.md)

### DIFF
--- a/text/0082-combined-replay-envelope-item.md
+++ b/text/0082-combined-replay-envelope-item.md
@@ -1,0 +1,35 @@
+- Start Date: 2023-03-24
+- RFC Type: informational
+- RFC PR: https://github.com/getsentry/rfcs/pull/82
+- RFC Status: draft
+
+# Summary
+
+One paragraph explanation of the feature or document purpose.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Background
+
+The reason this decision or document is required. This section might not always exist.
+
+# Supporting Data
+
+[Metrics to help support your decision (if applicable).]
+
+# Options Considered
+
+If an RFC does not know yet what the options are, it can propose multiple options. The
+preferred model is to propose one option and to provide alternatives.
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+# Unresolved questions
+
+- What parts of the design do you expect to resolve through this RFC?
+- What issues are out of scope for this RFC but are known?
--- a/text/0082-combined-replay-envelope-item.md
+++ b/text/0082-combined-replay-envelope-item.md
@@ -25,7 +25,7 @@ We'd then modify the sentry-sdk to send combined item types going forward
 graph
     A([SDK Generated Event])-->|HTTP Request| Z{Relay}
     Z-->|Published to Kafka topic|B[Replays Kafka Cluster]
-    B-->|ingest-replay|J[Replay Consumer]
+    B-->|ingest-replay-events|J[Snuba Consumer]
     J-->|Buffered Bulk Commit|K[(Clickhouse)]
     B-->|ingest-replay-recordings|E[Recording Consumer]
     E-->C[Recording Chunk Processor]
--- a/text/0082-combined-replay-envelope-item.md
+++ b/text/0082-combined-replay-envelope-item.md
@@ -39,10 +39,10 @@ graph
 
 We'd like to combine these envelope payloads into one for the following reasons:
 
-- Now that we're decompressing, parsing, and indexing data in the recording, the dileniation between the two events no longer makes sense.
+- Now that we're decompressing, parsing, and indexing data in the recording, the delineation between the two events no longer makes sense.
 - Right now there exists a race condition between ReplayEvent and ReplayRecording -- if a ReplayEvent makes it to clickhouse and is stored before the ReplayRecording, it can result in a bad user experience as a user can navigate to the replay, but the replay's blobs may not be stored yet. We'd like to change it so the snuba writes happen _downstream_ from the recording consumer, as we don't want to make a replay available for search until it's corresponding recording blob has been stored.
 - Right now our rate limits are separated per Itemtype. It would be less confusing if the rate limit applied to a single event type.
-- It is very hard to do telemetry on the recordings consumer now as we do not have SDK version. combining the envelopes allows us to have metadata in our recordings consumer in an easily accesible way.
+- It is very hard to do telemetry on the recordings consumer now as we do not have SDK version. combining the envelopes allows us to have metadata in our recordings consumer in an easily accessible way.
 
 # Options Considered
 
@@ -74,7 +74,7 @@ graph
 ReplayEventJSON\nCompressedReplayRecording
 ```
 
-That the downstream consumer can easily parse by splittin on newline.
+That the downstream consumer can easily parse by splitting on newline.
 
 ### SDK Changes
 
@@ -98,5 +98,5 @@ This is a decent chunk of engineering work.
 
 # Unresolved questions
 
-- Is there a better format for the combined envelope item type? If we split on \n, this means the replay_event should never have a newline in it. I belive this is acceptable, but is there a better format for sending a combined JSON / binary piece of data?
+- Is there a better format for the combined envelope item type? If we split on \n, this means the replay_event should never have a newline in it. I believe this is acceptable, but is there a better format for sending a combined JSON / binary piece of data?
 - If rate limits are applied before processing, it seems like we'll need to add a new rate limit for this combined item. This should be okay, anything else to think about here?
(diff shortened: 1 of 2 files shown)


text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+
+# Motivation
+
+## The Current Flow:
+
+```mermaid
+graph
+    A([SDK Generated Event])-->|HTTP Request| Z{Relay}
+    Z-->|Published to Kafka topic|B[Replays Kafka Cluster]
+    B-->|ingest-replay|J[Replay Consumer]
```
bmckerry, line 28:
```suggestion
    B-->|ingest-replay-events|J[Snuba Consumer]
```
matches the labels used in the other diagram below

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+
+```
+ReplayEventJSON\nCompressedReplayRecording
+```
+
+replacing the previous transport calls.
+
+### Replay Recording Consumer Changes
+
+1. Create a new ReplayRecording consumer that can run along-side the existing consumer, as there will be a change-over period
```
cmanallen, line 91: I don't think we need a new consumer.  We can switch on the `type` field in the kafka payload.  Right now the struct name is converted to snake-case and applied as the "type" value of the serialized object.  We would update our consumer to accept the new type and then process accordingly.

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+2. This consumer will do all the same things as the previous recordings consumer with the addition of:
+   - emitting a JSON event to the snuba-replay-events kafka topic
+
+# Drawbacks
+
+This is a decent chunk of engineering work.
+
+# Unresolved questions
+
+- Is there a better format for the combined envelope item type? If we split on \n, this means the replay_event should never have a newline in it. I believe this is acceptable, but is there a better format for sending a combined JSON / binary piece of data?
```
cmanallen, line 101: Caught this at the very end.  I'm leaving my previous feedback in place :P

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+- If rate limits are applied before processing, it seems like we'll need to add a new rate limit for this combined item. This should be okay, anything else to think about here?
```
cmanallen, line 102: I'm not sure I understand.  If rate-limits are applied before processing then this merged type doesn't exist yet.

text/0082-combined-replay-envelope-item.md
cmanallen: Tags could have a new-line character.

text/0082-combined-replay-envelope-item.md
JoshFerge: yeah, this would only be the case when people are on new SDKs that send the proposed new combined envelope format

text/0082-combined-replay-envelope-item.md
JoshFerge: got it. 👍🏼

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+### Relay Changes:
+
+1. Create a new ItemType `CombinedReplayRecordingEvent`
+2. in `processor.rs` https://github.com/getsentry/relay/blob/606166fca57a84ca1b9240253013871d13827de3/relay-server/src/actors/processor.rs#L1134, replace the `process_replays` function that will combine the envelope items if its the old format using `take_item_by` and `add_item` Envelope functions, or leave them if its the new format. This function would then process the two items and set the item payload.
+3. We can begin rejecting replay segments over 10 megabytes, which will allow us to not need chunking in our recording consumer.
+
+- The combined Payload will have the format of
+
+```
+ReplayEventJSON\nCompressedReplayRecording
```
mydea, line 74: Instead of making this a newline-separated string, could we also send this as a JSON like this:
```
{
  event: ReplayEventJSON,
  recording: CompressedReplayRecording
}
```

Not sure about implications for parsing this, but from an SDK perspective this seems a bit nicer/safer than string-concatting this. But not a hard feeling, just a thought.

mydea: a wait, I guess we can't put the compressed stuff into JSON 🤔 but now that I think about this, not sure if we can concat the UInt8Array that is the compressed payload onto a string as well 🤔

cmanallen: If this merged type is generated on the client then you will need to serialize with a binary format like `msgpack`.

mydea: hmm, I see. I guess we can do that, convert the payload into UInt8Array if it isn't compressed, then combine them together into a single UInt8Array.

One downside to note here is that it may make debugging harder - I currently tend to disable compression when debugging issues, in order to see what data is sent. This would not be possible anymore (as the payload would then _always_ be UInt8Array). Not a blocker, just to be aware of this downside!

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+That the downstream consumer can easily parse by splitting on newline.
+
+### SDK Changes
```
jjbayer, line 79: What is the motivation for introducing the format in SDKs? If Relay needs to handle legacy envelopes anyway, would it be acceptable to leave SDKs as-is? Or would using a single item simplify SDK code as well?

JoshFerge:
update: in the RFC I proposed a combined format using a newline. I'd like to propose an alternative format of which we transmit it on the backend.

The Replays Recording Consumer expects MessagePack bytes in the `payload` field. I propose that in our [processor](https://github.com/getsentry/relay/blob/082b8b4e323850eceb44e611002d0f93188d63dd/relay-server/src/actors/processor.rs#L1238), we combine the ReplayEvent and ReplayRecording into a single messagepack payload, and change the [content type ](https://github.com/getsentry/relay/blob/082b8b4e323850eceb44e611002d0f93188d63dd/relay-server/src/actors/processor.rs#L1290)to `ContentType::MessagePack`.

The messagepack payload will be a key value object like:
```
replay_event: {...}
replay_recording: _binary_data_ or uncompressed data
replay_recording_headers {}
```
In our recordings consumer, we'll parse the message pack which will then give us a dict with these values.


Depending on whether or not we can specify the entire kafka message be compressed or not, we can set the replay recording to the uncompressed value, or as a byte array.

We can also take this opportunity to put the parsed recording headers in the item in relay. This will remove another complication downstream in the consumer.


We could either emit these events onto a new topic with a new consumer, or have the existing consumer use a field to know whether or not the payload is a combined one or not. I think it would be easy to add a field onto the kafka message, like 

```
"combined_event_and_recording": True
```

I will re-evaulate whether or not its worth it to make SDK changes to combine the envelope client side, but server side I feel good about this solution.

cmanallen:
@JoshFerge 

> I think it would be easy to add a field onto the kafka message, like

Agree.  Though I think a `version` field is better.  Since we don't have one currently we can perform a condition check if the version field exists and if it does do something.  In the future, we can increment the version number to handle new payload types.

JoshFerge:
@cmanallen great, i'll get started on this today.

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,100 @@
...
+In our recordings consumer, we'll parse the message pack which will then give us a dict with these values.
+
+We can also take this opportunity to put the parsed recording headers in the item in relay. This will remove another complication downstream in the consumer.
+
+We will emit these combined messages onto the existing replay recorings kafka topic, with a new field `version` added which the consumer will use to know that the ReplayEvent exists on the message as well.
+
+
+### Replay Recording Consumer Changes
+
+1. Look at the version of the payload to determine if its the new event, and if so, handle additional work for the ReplayEvent, and load the SDK contextual data into the ReplayRecording for events emitted from it.
```
bruno-garcia, line 90:
```suggestion
1. Look at the version of the payload to determine if it's the new event, and if so, handle additional work for the ReplayEvent, and load the SDK contextual data into the ReplayRecording for events emitted from it.
```

text/0082-combined-replay-envelope-item.md
```diff
@@ -0,0 +1,100 @@
...
+
+ReplayEvent and ReplayRecording.
+
+- ReplayEvent contains indexable metadata that is stored in our clickhouse table.
+- ReplayRecording contains a large, usually compressed blob that is written to our Filestore (GCS as of now)
+
+These are always sent in the same envelope by the sdk. We'd like to combine these two envelope item types into one.
+
+We'd first write a compatibility layer in relay that takes the old format and converts them into a single item type.
+We'd then modify the sentry-sdk to send combined item types going forward
```
bruno-garcia, line 18: We need to consider compatibility with Self Hosted (and Single Tenant) before we change the SDK

bruno-garcia (approved):
Makes sense. Let's do this! Left a note on future considerations regarding SDK changes but doesn't affect anything agreed here

cmanallen (approved):

PR #83, opened 2023-03-29
Starfish Tracing Model

mitsuhiko:
This RFC proposes a new tracing model for Sentry's performance product to better
address future product improvements.  The goal of this model is to allow
**storing entire traces witohut gaps**, support **dynamic sampling**, **indexing
of spans** and to **extract metrics from spans pre-sampling**.  This is an
evolution of the current transaction based approach.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/starfish-tracing-model/text/XXXX-starfish-tracing-model.md)

### DIFF
--- a/text/0000-starfish-tracing-model.md
+++ b/text/0000-starfish-tracing-model.md
@@ -1,0 +1,151 @@
+- Start Date: 2023-03-29
+- RFC Type: feature
+- RFC PR: <link>
+- RFC Status: draft
+
+# Summary
+
+This RFC proposes a new tracing model for Sentry's performance product to better
+address future product improvements.  The goal of this model is to allow
+**storing entire traces witohut gaps**, support **dynamic sampling**, **indexing
+of spans** and to **extract metrics from spans pre-sampling**.  This is an
+evolution of the current transaction based approach.
+
+# Motivation
+
+Today Sentry has a strong concept of a "transaction" which appears in multiple parts of the
+product.  It is both the transport of span data, the billable entity and the only indexed
+part of the product experience.  This means that spans that exist outside of the transaction
+cannot be represented and it also means that spans within a transaction are not indexed itself.
+
+The existing model has worked well for us to get started with evolving the Sentry errors
+product to capture performance traces, but it has restricted out ability to evolve the product
+forward.  It has created some restrictions on the SDK technology side (from high
+memory pressure, payload size limits) and also has promoted a separate of transaction to
+span on the API layer which is untypical for tracing products.  It also has meant that Sentry
+has challenges with accepting traces coming directly from an OpenTelemetry exporter as the
+transaction concept is not a concept that OpenTelemetry has.
+
+We want to set a future direction that enables more flexible product choices and that we can
+move towards from our existing tracing model.  The goals are:
+
+* Support capturing entire traces
+* Have a data model story that allows us higher compatibility with Open Telemetry.  Specifically
+  we want a model that would permit us to ingest Open Telemetry data right from an exporter
+* Have a clear story for indexing and extracing metrics on a per-span level
+* Unified spans and transactions from an SDK perspective
+* Enable a path that allows clients to send 100% of spans outwards at least to a local aggregator,
+  preferrably a remote relay
+* To better and directly support dynamic sampling in the core tracing model
+
+We want to lay out a better path forward that 
+
+* capture entire traces
+* browser tabs = a trace
+* index and extract metrics on a span level
+* clients send 100% of metrics
+* dynamic sampling narrows in on traces
+
+# Terms
+
+The new tracing model is an extension to our existing tracing model.  As such we try go adhere
+to some of the existing terms.  Note that this document is intentionally glossing over some of
+the details to better describe the desired end result.  Individual RFCs will have to be written
+to narrow down on specific schema definitions.
+
+## Session
+
+A session is an optional concept when talking about user actors on the system.  A
+session outlives one or more traces and is exclusively used when talking about human
+interactions with a system.
+
+## Trace
+
+A trace has no end.  It bundles spans together, some of those spans are organized
+into segments (marked in red in the graph).  The user experience does not center
+around a trace, which really is an internal way to bundle things together but it
+narrows down on segments within the trace.
+
+```mermaid
+gantt
+    title Example Starfish Trace
+    dateFormat x
+    axisFormat %S.%L
+
+    section Frontend
+    /checkout                                        :crit, 0, 1500ms
+    GET /api/session                                 :150, 170ms
+    POST /api/analytics                              :190, 70ms
+    GET /api/checkout/state                          :200, 500ms
+    GET /api/checkout/cart                           :1100, 140ms
+    <App/>                                           :1300, 180ms
+    POST /api/analytics                              :done, 1450, 70ms
+    GET /assistent/poll                              :done, 1450, 120ms
+    POST /api/analytics                              :done, 1580, 70ms
+
+    section API Service
+    /api/checkout/state                              :crit, 240, 440ms
+    cache.get session#58;[redacted]                  :360, 10ms
+    db.query select from session                     :370, 20ms
+    db.query select from user                        :390, 20ms
+    db.query select from checkout                    :410, 20ms
+    http.request GET http#58;//payments/poll  :450, 210ms
+    thread.spawn refresh-checkout-cache              :done, 670, 220ms
+
+    section Payment Service
+    /poll                                            :crit, 470, 180ms
+    db.query select from payment                     :490, 30ms
+    db.query update payment                          :530, 60ms
+```
+
+## Span
+
+Spans are very similar to how the function today, but they get elevated to a more significant
+level.  They largely follow the general semantics in the wider tracing eco system.  To drive
+our product ideas we are going to ensure that the quality of the spans is high and that they
+provide at least the following pieces of information:
+
+* `op`: defines the core operation that the span represents (eg: `db.query`)
+* `description`: the most significant description of what this operation is (eg: the database query).
+  The description also gets processed and cleaned in the process.
+* `trace_id`: a span relates to one trace by ID
+* `parent_id`: a span optionally points back to a parent trace which could be from a different
+  service
+* `segment_id`: a span that is part of a segment, always refers back to it.
+* `tags`: a key/value pair of arbitrary tags that are set per-span.  Conceptionally however spans
+  also inherit the tags of the segments they are contained in.
+
+Spans always belong to a trace, but not all spans belong to a segment. A span not belonging to a
+segment are referred to as "detached" spans, spans that belong to a segment are "attached" spans.
+
+## Transaction
+
+Previously transaction referred to the container that held spans.  In the future a "transaction"
+refers to the name of a specific type of segment level span that describes a meaningful activity
+that starts with a request, and results in some meaningful response.  By definition a span that
+holds a transaction tag becomes a "segment".  Other than that, it's just a regular span.
+
+## Segment
+
+A segment is a special type of span that is the "logical" activity in a service.  For instance a
+segment can be the endpoint implementation of an API request, it could be a task that is processed
+by a task worker, it could be the navigation a user performs in a UI or a screen transition.
+Conceptionally segments fall into two categories: "transactions" which are quite mechanical and
+clearly defined operations and "interactions" which are user triggered operations.  The difference
+is that an "interaction" has a human user as an actor in it, that might influence it, whereas a
+"transaction" is unlikely to be interrupted once started.  A user for instance is quite likely to
+navigate again even before the previous interaction finished, whereas a task is more likely than no
+to conclude, even if what triggered the task is no longer interested in it's result.
+
+The primary user experience in the product can narrow down on certain segments and make the trace
+explorable via that segment.
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+# Unresolved questions
+
+- What parts of the design do you expect to resolve through this RFC?
+- What issues are out of scope for this RFC but are known?


text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+**storing entire traces without gaps**, support **dynamic sampling**, **indexing
+of spans** and to **extract metrics from spans pre-sampling**.  This is an
+evolution of the current transaction based approach.
+
+# Motivation
+
+Today Sentry has a strong concept of a "transaction" which appears in multiple parts of the
+product.  It is both the transport of span data, the billable entity and the only indexed
+part of the product experience.  This means that spans that exist outside of the transaction
+cannot be represented and it also means that spans within a transaction are not indexed itself.
```
markstory, line 19:
```suggestion
cannot be represented and it also means that spans within a transaction are not indexed.
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+
+The existing model has worked well for us to get started with evolving the Sentry errors
+product to capture performance traces, but it has restricted out ability to evolve the product
```
markstory, line 22:
```suggestion
product to capture performance traces, but it has restricted our ability to evolve the product
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+forward.  It has created some restrictions on the SDK technology side (from high
+memory pressure, payload size limits) and also has promoted a separate of transaction to
+span on the API layer which is untypical for tracing products.  It also has meant that Sentry
```
markstory, line 25:
```suggestion
memory pressure, payload size limits) and also has promoted a separation of transactions and
spans on the API layer which is untypical for tracing products.  It also has meant that Sentry
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+
+* capture entire traces
+* browser tabs = a trace
+* index and extract metrics on a span level
+* clients send 100% of metrics
+* dynamic sampling narrows in on traces
+
+# Terms
+
+The new tracing model is an extension to our existing tracing model.  As such we try go adhere
```
markstory, line 51:
```suggestion
The new tracing model is an extension to our existing tracing model.  As such we try to adhere
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+
+    section Payment Service
+    /poll                                            :crit, 470, 180ms
+    db.query select from payment                     :490, 30ms
+    db.query update payment                          :530, 60ms
+```
+
+## Span
+
+Spans are very similar to how the function today, but they get elevated to a more significant
```
markstory, line 103:
```suggestion
Spans behave very similar to how the function today, but they get elevated to a more significant
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+* `op`: defines the core operation that the span represents (eg: `db.query`)
+* `description`: the most significant description of what this operation is (eg: the database query).
+  The description also gets processed and cleaned in the process.
+* `trace_id`: a span relates to one trace by ID
+* `parent_id`: a span optionally points back to a parent trace which could be from a different
+  service
+* `segment_id`: a span that is part of a segment, always refers back to it.
+* `is_segment`: when set to `true` this span is a segment.
+* `start_time` and `end_time` to give the span time information.
+* `tags`: a key/value pair of arbitrary tags that are set per-span.  Conceptionally however spans
```
markstory, line 117:
```suggestion
* `tags`: a key/value pair of arbitrary tags that are set per-span.  Conceptually however spans
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+## Segment
+
+A segment is a special type of span that is the "logical" activity in a service.  For instance a
+segment can be the endpoint implementation of an API request, it could be a task that is processed
+by a task worker, it could be the navigation a user performs in a UI or a screen transition.
+Conceptionally segments fall into two categories: "transactions" which are quite mechanical and
+clearly defined operations and "interactions" which are user triggered operations.  The difference
+is that an "interaction" has a human user as an actor in it, that might influence it, whereas a
+"transaction" is unlikely to be interrupted once started.  A user for instance is quite likely to
+navigate again even before the previous interaction finished, whereas a task is more likely than no
```
markstory, line 144:
```suggestion
navigate again even before the previous interaction finished, whereas a task is more likely than not
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+to conclude, even if what triggered the task is no longer interested in it's result.
```
markstory, line 145:
```suggestion
to conclude, even if what triggered the task is no longer interested in the result of the task.
```

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Span
+
+Spans are very similar to how the function today, but they get elevated to a more significant
+level.  They largely follow the general semantics in the wider tracing eco system.  To drive
+our product ideas we are going to ensure that the quality of the spans is high and that they
+provide at least the following pieces of information:
+
+* `op`: defines the core operation that the span represents (eg: `db.query`)
+* `description`: the most significant description of what this operation is (eg: the database query).
```
AbhiPrasad, line 109: I would heavily be in favor of re-naming this to `name`.

For those unfamiliar we currently have an inconsistency in the sentry schema where transactions have a `name` attribute and spans have a `description` (functionality identical).

I think most users are familiar with name as an attribute (closer to otel and other vendors as well then), and it would require less delta for us to approach.

We even alias span descriptions to names in the UI

image

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+  The description also gets processed and cleaned in the process.
+* `trace_id`: a span relates to one trace by ID
+* `parent_id`: a span optionally points back to a parent trace which could be from a different
```
AbhiPrasad, line 112:
```suggestion
* `parent_span_id`: a span optionally points back to a parent trace which could be from a different
```

text/0083-starfish-tracing-model.md
AbhiPrasad, line 117: Currently there are a couple ways to store arbitrary data on the existing schema:

on transactions you have tags, contexts and extras (inherting from error events)

on spans you have tags and data. 

As per our spec, tags are restricted to string keys -> string values, and have strict size limits., while data is restricted to string keys, but can have arbitrary values (with more liberal size limits).

This has lead to most of the SDKs storing [arbitrary values in span data](https://develop.sentry.dev/sdk/performance/span-data-conventions/), and the OpenTelemetry conversion setting span attributes (otels arbitrary span data) in the data field as well.

Span data is now also used for semantic conventions, similar to how opentelemetry semantic conventions use otel span attributes. I would like to keep this behaviour. For example, the `span.data` with the key [`blocked_main_thread`](https://develop.sentry.dev/sdk/performance/span-data-conventions/#mobile) is used in the file io performance issue detector.

 I would like an arbitrary place to continue to store data to still exist, so that means we either move to saying this is `data`, or we update the limitations on `tags`.

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+    http.request GET http#58;//payments/poll  :450, 210ms
+    thread.spawn refresh-checkout-cache              :done, 670, 220ms
+
+    section Payment Service
+    /poll                                            :crit, 470, 180ms
+    db.query select from payment                     :490, 30ms
+    db.query update payment                          :530, 60ms
+```
+
+## Span
```
AbhiPrasad, line 101: I noticed a distinct lack of `span.status` (let's goooo).

Would it be prudent to introduce a basic status field though, similar to OpenTelemetry's? This will have 3 values:

- `ok` (successfully finished)
- `error` (error occurred during span)
- `unknown` (default value)

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+
+Spans are very similar to how the function today, but they get elevated to a more significant
+level.  They largely follow the general semantics in the wider tracing eco system.  To drive
+our product ideas we are going to ensure that the quality of the spans is high and that they
+provide at least the following pieces of information:
+
+* `op`: defines the core operation that the span represents (eg: `db.query`)
+* `description`: the most significant description of what this operation is (eg: the database query).
+  The description also gets processed and cleaned in the process.
+* `trace_id`: a span relates to one trace by ID
```
AbhiPrasad, line 111: Don't we also need a `span_id`? Otherwise how can we form a proper parent child relationship?

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+* `parent_id`: a span optionally points back to a parent trace which could be from a different
+  service
+* `segment_id`: a span that is part of a segment, always refers back to it.
```
AbhiPrasad, line 114: What is the type of a `segment_id`? Will it be 32 hex chars like a `trace_id`, 16 hex chars like a `span_id`, or something else entirely?

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+* `is_segment`: when set to `true` this span is a segment.
```
AbhiPrasad, line 115: If segment is meant to be either a transaction or an interaction, should we build it into the spec here?

so `segment: "transaction" | "interaction" | null`, where null says that the span is not a segment.

AbhiPrasad (commented):
Exciting!

We should probably add some notes about cardinality (in span name, span tags, extracted metrics, etc.), I feel like that's necessary information to make the best decisions around this.

text/0083-starfish-tracing-model.md
mitsuhiko: FWIW I did not spec out the entire model here but yes, `status` should be on there.

text/0083-starfish-tracing-model.md
mitsuhiko: I think if we want to get rid of `description` as the primary field then we need to rethink how we do the UX around some of those pieces. We currently stash into `description` what others for instance stash into `http.route` etc.

text/0083-starfish-tracing-model.md
mitsuhiko: It's the `span_id` of the segment it points to. Will clarify.

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,181 @@
...
+* `tags`: a key/value pair of arbitrary tags that are set per-span.  Conceptionally however spans
+  also inherit the tags of the segments they are contained in.
+* `measurements`: are span specific metrics that are stored with the span.  There is a natural
+  measurement of a span which is the `duration` which is automatically calculated from the difference
+  of the end to the start timestamp.
+
+Spans always belong to a trace, but not all spans belong to a segment. A span not belonging to a
+segment are referred to as "detached" spans, spans that belong to a segment are "attached" spans.
+A span must only be attached to a segment if it belongs to the same process and service.  Remote
+spans must never be attached to a segment.
```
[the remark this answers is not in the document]
mitsuhiko: Again, that schema is not fully defined yet. However remote means from another process or service.

text/0083-starfish-tracing-model.md
AbhiPrasad: I don't think we need to get rid of description, simply just call it `name` instead of `description`. It can work the same exact way.

text/0083-starfish-tracing-model.md
mitsuhiko: I agree that we should be keeping this. I won't have time yet to focus on the span schema so if you want, maybe you can sketch this out directly on the RFC with what the schema should look like?

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,214 @@
...
+We want to set a future direction that enables more flexible product choices and that we can
+move towards from our existing tracing model.  The goals are:
+
+* Support capturing entire traces
+* Have a data model story that allows us higher compatibility with Open Telemetry.  Specifically
+  we want a model that would permit us to ingest Open Telemetry data right from an exporter
+* Have a clear story for indexing and extracing metrics on a per-span level
+* Unified spans and transactions from an SDK perspective
+* Enable a path that allows clients to send 100% of spans outwards at least to a local aggregator,
+  preferrably a remote relay
```
fpacifici, line 38: Which types of aggregation do you have in mind here ?
Is that the usual metrics extraction or something else?

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,214 @@
...
+## Span
+
+Spans behave very similar to how the function today, but they get elevated to a more significant
+level.  They largely follow the general semantics in the wider tracing eco system.  To drive
+our product ideas we are going to ensure that the quality of the spans is high and that they
+provide at least the following pieces of information:
+
+* `op`: defines the core operation that the span represents (eg: `db.query`)
+* `description`: the most significant description of what this operation is (eg: the database query).
+  The description also gets processed and cleaned in the process.
```
fpacifici, line 110: How do we expect to use this? 
When we experimented with span search and when we built suspect spans one of the major issues was that all the useful information was stored unstructured in this field and all queries would have been regexes. It did not end well. 

@nikhars , if this is going to be a searchable field you may want to look into ngram indexes.

dcramer:
Ive not been able to review this in depth but I wanted to escalate the convo we're having in Slack.

I do not believe a session should be a trace, or vice versa. To me these are different constructs. I, as a user, expect a trace to represent a request cycle. That is, when I take an action, a trace occurs, and it ends when that action fully completes. That said, there are problems in all models.

To illustrate what I mean, I want to call out several cases that begin root traces:

- I curl an API endpoint, it starts when the edge receives the request
- I load a SPA, it starts when the browser initiaties the request
- I load a Slack, on my Desktop, it starts when I load it

When do these traces end?

- API endpoint is straight forward. Commonly it'll end when the response is returned (unless some async work happens)
- SPA could end when the rendering completes, however, there may still be things going on. What happens with that instrumentation?
- Slack, it could end when the app has fully loaded, but what about all the other tasks going on. Should background tasks be their own trace? Then they're sessions again. Should whoevers pushing data to Slack initiate the trace? You lose context.

My pov is all models are trash. You lose data if you make traces behave as expected, which is what I originally describe here. Traces were not designed for long lived applications and frankly speaking they do not work for it.

Should we throw it all away, call it something else, and just go with it? e.g. we could just call it Session Tracing, explicitly say a trace generally speaking will live for an entire session? This has its own problems.

I load a SPA, or Slack, and keep it open for days. All those requests are a single trace.

- How do I sample? Sampling an entire session is prohibitively expensive, thus a session still needs bounds.
- How do I make sense of the trace data? Theres infinite segments/transactions inside of it. Segments solve for this but its still an immense amount of complexity.
- How do I make sense of interactions?

I'm sure there are more things I've forgotten to write up here, but the problem does not seem to have a right answer.

mattjohnsonpint:
If we're really considering a new model, IMHO this would be a good time to tackle the time synchronization problem.  Timestamps aren't enough  IMHO, because each clock that generates them is synchronizationes independently. It's all to easy for a "distributed trace" to show nonsensical timestamps when viewed as a whole, such as server responses that complete before the request has been sent.  I have some ideas here we could experiment with.  Would this be welcome as part of this new model?

text/0083-starfish-tracing-model.md
```diff
@@ -0,0 +1,214 @@
...
+* Enable a path that allows clients to send 100% of spans outwards at least to a local aggregator,
+  preferrably a remote relay
+* To better and directly support dynamic sampling in the core tracing model
+
+We want to lay out a better path forward that 
+
+* capture entire traces
+* browser tabs = a trace
+* index and extract metrics on a span level
+* clients send 100% of metrics
```
JoshFerge, line 46: maybe this is in the TODO, but how does this proposal help with this "100% sending" or "memory-pressure/payload size" issue listed above? what about the current CPU overhead?

It is helpful for me to remember our own sampling limits, even after dynamic sampling has been implemented, are often set extremely low for performance reasons https://github.com/getsentry/sentry/blob/cc47128efd5a00e66b2475acc318ff57860fdf94/src/sentry/utils/sdk.py#L35

JoshFerge: even with a local relay I don't understand how we could ever have metrics sampled enough to be useful for SLAs / Alerts where performance of the process is not severely degraded in medium throughput and above endpoints. It seems to me the only way to accomplish this is for metrics collection directly in the process.

JoshFerge: ah maybe this is what 
>**extract metrics from spans pre-sampling**.

is going to address?

maggiepint:
Sorry for rather drive-by comments but since I have a vested interest in better UI tracing a couple thoughts.

I want to note from experience that most SLO initiatives fail to meaningfully improve reliability in organizations. This is generally because SLOs are measured over individual services and endpoints, and that aggregation doesn't connect meaningfully to user experience. That's really the same problem @dcramer is calling out here.

There's a lot of opportunity to build a better model, but I think you'd have to think somewhat platform specific. If I live just in React in my head (though I think conceptually this fits the whole web platform), the meaningful measurements I would want for my users are:
- Route/url change
- Interactions that load new stuff inline (infinite scroll, drawers, modals, etc)
- Events to server (like form posts)
- Events from server (server sent events)

When you think about it - nearly all of these events will have some 'main' component/DOM element associated with them. Figuring out what that main component is is is hard because they nest, but logically:
- router loads
- New suspense or error boundary loads (well designed modals and drawers will have these, less well designed stuff this is tough)
- Button or link clicks/form posts

All mark the start of a new transaction and most likely the end of the previous one. They also likely represent a logical 'feature' in terms an average product manager would understand.

Server sent events are a wild-card because they represent an almost parallel thread to your main UI thread - but I do feel they could almost always be looked at separately as they are async and at least shouldn't represent real-time interaction (the chat scenario being the weird counterexample that must always exist - but then each chat exchange can likely be thought of as it's own transaction as well.)

This might be possible through auto-instrumentation if you are willing to go into the frameworks (historically y'all are). I think you end up with sort of a 'transaction type' and then a 'transaction boundary' following this schema - where the 'transaction type' is 'route | suspense | button/form' and the boundary is the name of the associated thing (dom ID/name, component name, etc).

text/0083-starfish-tracing-model.md
JoshFerge: Screenshot 2023-04-03 at 5 57 56 PM
from a sentry staff engineer. I feel like for the type of customer that wants to use this product, _all_ of their endpoints are defined as "high perf backends".  have we talked to prospective customers about this?

weyert:
You might find the OpenTelemetry Client Instrumentation SIG group interesting:
https://docs.google.com/document/d/16Vsdh-DM72AfMg_FIt9yT9ExEWF4A_vRbQ3jRNBe09w/edit#heading=h.yplevr950565

Quick scan I didn't see Sentry mentioned in it yet. I think it would be great to consider
join the group and solve it together.

PR #84, opened 2023-04-06
feat(rfc): Move get started docs to sentry repo

priscilawebdev:
This RFC aims to brainstorm a more practical, dynamic, and scalable solution to create and render onboarding documentation for each platform we support.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/ecd6b62bbcd917f78d879053613ec932c579cf8f/text/0084-move-docs-to-sentry-repository.md)


text/0084-move-docs-to-sentry-repository.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Options considered
+
+1. Gradually move the get started docs for the onboarding from the sentry-docs repository to sentry, where we'll write everything in MDX.
+
+   That way, we can write everything in React and possibly reuse components from our component library, like alerts. This will allow us to be more consistent with styles used in our application, remove several functions we have in place for he current mechanism, and use best practices.
+
+   We can still ask the docs team for reviews on our Pull Requests when needed.
+
+2. Do you have other ideas? Please share with us!
```
AbhiPrasad, line 46: To propose an idea:

Could we change the docs repo to a monorepo and publish the onboarding instructions as a javascript library? Then we can write first in docs - and use dependabot to automatically update this dependency whenever needed. We could even hard pin to latest.

Source of truth only written once, and in a single place!

This helps with a few things:

a) Makes it much easier to add docs changes, we just need to `npm link` to test out onboarding instruction changes locally
b) We can use whatever libraries we want for mdx, and even write tests for them without needing to burden the sentry frontend repo.


In addition this opens us up to expanding the docs monorepo to publish other packages that other consumers could use.

priscilawebdev: yes, that is a good idea too! but do we use currently some tools like https://turbo.build/repo to manage our mono repo?

AbhiPrasad: Nope, we don't use anything like that in the docs repo, so we would have to add it. This does have non-trivial engineering effort.

text/0084-move-docs-to-sentry-repository.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Drawbacks
+
+1. If we go with option one proposed above, we will have to live with 2x different mechanisms for a while until we port all the documentation over to Sentry and we will have to install a few `@mdx-js/*` dependencies
```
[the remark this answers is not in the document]
AbhiPrasad: To clarify this.

After this change is completed we will have two sets of docs:

1. short form onboarding snippets that aim
2. long form sdk docs that cover all functionality (also includes short form onboarding docs)

This means that if short form onboarding snippet changes (for example we update what to provide `Sentry.init`), we'll first have to update sentry and then update the sentry-docs. Since this is two different repos, these can get out of sync (even when the migration is all over)

Although this is a drawback, it's not a huge one, since the onboarding docs should not change in massive ways (probably only during a major bump of the SDK).

PR #85, opened 2023-04-07
Source Hash Extension for Source Maps

mitsuhiko:
This proposes and specifices support for source hashes in Sentry.

Relates to https://github.com/getsentry/rfcs/pull/81

[Rendered RFC](https://github.com/getsentry/rfcs/blob/source-hash-sourcemap-keying/text/XXXX-source-hash-sourcemap-keying.md)

### DIFF
--- a/text/xxxx-source-hash-sourcemap-keying.md
+++ b/text/xxxx-source-hash-sourcemap-keying.md
@@ -1,0 +1,160 @@
+- Start Date: 2023-04-06
+- RFC Type: feature
+- RFC PR: <link>
+- RFC Status: draft
+
+# Summary
+
+This specification proposes the keying / discovering of source maps via
+the source hashes of the generated JavaScript files.  This is an alternative
+lookup method to the proposed debug ID lookup.  It's there to support the
+`CallSite.getScriptHash` browser API that Chrome and Edge already ship.
+
+# Motivation
+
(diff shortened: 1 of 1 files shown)


text/0085-source-hash-sourcemap-keying.md
```diff
@@ -0,0 +1,162 @@
...
+}
+```
+
+Note that they use of the embedded key for now is likely discouraged for the
+made up header as we do not plan on standardizing the file hashes.
+
+To avoid re-calculating the SHA 256 hashes of the minified files constantly we
+also want to add them to the manifest.  For this we could leverage the digest
+header, with the small suboptimal situation that the HTTP digest spec uses
+a base64 hash:
```
Swatinem, line 109: We should just embrace the reality that the `headers` have nothing to do with HTTP anymore (even though they originally were inspired by that).

This is all a bit bikesheddy, but we should strive for consistency here between the protocol, the sourcemap JSON, and these `headers`. We use a mix of different names and string/objects all over the place.
I believe we are currently limited by `headers` being a `String->String` mapping. But we could relax that as well if needed.

mitsuhiko: I would honestly just move all this stuff out of `headers` and keep toplevel, but as it stands these are actually still http headers in the current implementation. It really just overlays the HTTP fetch.

PR #86, opened 2023-04-13
rfc(feature): Sentry Bundler Plugins API

lforst:
We wanna update our bundler plugins API with the next major.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/sentry-bundler-plugins-api/text/0086-sentry-bundler-plugins-api.md)

### DIFF
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -9,7 +9,7 @@ This RFC intends to discuss the future API of the [Sentry JavaScript Bundler Plu
 
 # Motivation
 
-With the major bump to v2 of the bunlder plugins, we have the opportunity to rethink the existing API that was orignially built on top of the [Sentry Webpack Plugin](https://github.com/getsentry/sentry-webpack-plugin).
+With the major bump to v2 of the bundler plugins, we have the opportunity to rethink the existing API that was orignially built on top of the [Sentry Webpack Plugin](https://github.com/getsentry/sentry-webpack-plugin).
 
 Facts to consider when deciding on an API:
 
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -1,7 +1,9 @@
 - Start Date: 2023-04-13
 - RFC Type: feature
 - RFC PR: https://github.com/getsentry/rfcs/pull/86
-- RFC Status: draft
+- RFC Status: approved
+- RFC Driver: [Luca Forstner](https://github.com/lforst)
+- RFC Approver: JavaScript SDK Team
 
 # Summary
 
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -49,7 +49,7 @@ interface Options {
   sourcemaps?: {
     assets: string | string[]; // globs pointing to the javascript files that should be uploaded to Sentry with their source maps - these are the built assets and not the source files
     ignore?: string | string[]; // globs to exclude javascript files from being uploaded (will also not upload their source maps)
-    sourcesRoot?: string | null; // default: process.cwd() - We rewrite the `sources` field in source maps to point to sources relative to this value. Set to `null` to disable this rewriting.
+    rewriteSources?: (source: string) => string; // Hook to rewrite individual entries in the sourcemaps's `sources` field. By default, if not defined, the plugin will try to rewrite the entries to be relative to `process.cwd()`.
   };
 
   // General configuration for release creation and release injection - we decouple this from the debug ID source maps upload because they're technically not related anymore.
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -50,6 +50,7 @@ interface Options {
     assets: string | string[]; // globs pointing to the javascript files that should be uploaded to Sentry with their source maps - these are the built assets and not the source files
     ignore?: string | string[]; // globs to exclude javascript files from being uploaded (will also not upload their source maps)
     rewriteSources?: (source: string) => string; // Hook to rewrite individual entries in the sourcemaps's `sources` field. By default, if not defined, the plugin will try to rewrite the entries to be relative to `process.cwd()`.
+    deleteAfterUpload?: string | string[]; // globs pointing to any assets that should be deleted after assets upload - example: "./dist/**/*.js.map"
   };
 
   // General configuration for release creation and release injection - we decouple this from the debug ID source maps upload because they're technically not related anymore.
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -1,0 +1,35 @@
+- Start Date: 2023-04-13
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/86
+- RFC Status: draft
+
+# Summary
+
+One paragraph explanation of the feature or document purpose.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Background
+
+The reason this decision or document is required. This section might not always exist.
+
+# Supporting Data
+
+[Metrics to help support your decision (if applicable).]
+
+# Options Considered
+
+If an RFC does not know yet what the options are, it can propose multiple options. The
+preferred model is to propose one option and to provide alternatives.
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+# Unresolved questions
+
+- What parts of the design do you expect to resolve through this RFC?
+- What issues are out of scope for this RFC but are known?
--- a/text/0086-sentry-bundler-plugins-api.md
+++ b/text/0086-sentry-bundler-plugins-api.md
@@ -5,31 +5,99 @@
 
 # Summary
 
-One paragraph explanation of the feature or document purpose.
+This RFC intends to discuss the future API of the [Sentry JavaScript Bundler Plugins](https://github.com/getsentry/sentry-javascript-bundler-plugins).
 
 # Motivation
 
-Why are we doing this? What use cases does it support? What is the expected outcome?
+With the major bump to v2 of the bunlder plugins, we have the opportunity to rethink the existing API that was orignially built on top of the [Sentry Webpack Plugin](https://github.com/getsentry/sentry-webpack-plugin).
 
-# Background
+Facts to consider when deciding on an API:
 
-The reason this decision or document is required. This section might not always exist.
+- The bump to v2 of unplugin will imply a major bump of the Sentry Webpack plugin and its switch to our unplugin based architecture.
+- We recently introduced debug IDs as a source map resolving mechanism. This results in different needed options needed when running the plugins. Some options must be added, some removed.
+- Releases are not a required parameter anymore for source maps upload.
+
+As an additional thought: The new API should be seen as a "clean-slate" API - it's less about what has been and more about what the optimal API should look like.
 
 # Supporting Data
 
-[Metrics to help support your decision (if applicable).]
+- Debug ID RFC: https://github.com/getsentry/rfcs/blob/ab0f75567d95e247b5cc3f09c466b7254844e654/text/0081-sourcemap-debugid.md
+- Debug ID changes related to the unplugin based Sentry bundler plugins: https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/191
 
 # Options Considered
 
-If an RFC does not know yet what the options are, it can propose multiple options. The
-preferred model is to propose one option and to provide alternatives.
+The API generally only defines the options passed to the individual plugins. The options are the same for all bundlers.
+
+For the sake of simplicity of this RFC, we will not decide on the JSdoc. All JSdoc below is just to illustrate what the options are used for.
+
+Proposed API (TypeScript typedef):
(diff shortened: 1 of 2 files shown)


text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+// Identical to status-quo. See https://github.com/getsentry/sentry-webpack-plugin/blob/2b7d274a7355f0d27a431b2c20c37c9786bbe4cb/README.md for more information.
+interface SetCommitsOptions {
+  // ...
+}
+```
+
+Some removed options:
+
+- `dryRun` - Instead, just not use the plugin.
```
lforst, line 88: Hm, maybe we should keep it. I can see how it is useful. I just want it to be clean - I guess by making every side-effect a no-op and emitting a message instead.

The thing is that some chains of execution don't make sense in dry run mode anymore: e.g. If we don't inject debug IDs into a bundle due to dry run, the debug ID source map upload step will emit a warning because a file doesn't have a debug id injected.

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,103 @@
...
+  url?: string; // A way to configure self-hosted instances: unchanged
+  headers?: Record<string, string>; // Headers added to every outgoing network request.
+  debug?: boolean; // default: false - Whether to print debug information.
+  silent?: boolean; // default: false - Whether to print anything at all.
+  errorHandler?: (err: Error) => void; // when provided will swallow errors unless the provided function throws itself - By default any error will stop compilation to abort
+  telemetry?: boolean; // default: true - whether to send telemetry to Sentry
+
+  // General configuration for source maps uploading - if not provided, source map uploading will be disabled
+  sourcemaps?: {
+    jsAssets: string | string[]; // globs pointing to the javascript files that should be uploaded to Sentry with their source maps - these are the built assets and not the source files
```
AbhiPrasad, line 49: Let's just call this `assets`? We could upload more than just js in the future!

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+The API generally only defines the options passed to the individual plugins. The options are the same for all bundlers.
+
+For the sake of simplicity of this RFC, we will not decide on the JSdoc. All JSdoc below is just to illustrate what the options are used for.
+
+Proposed API (TypeScript typedef):
+
+```ts
+interface Options {
+  org: string; // Previously this was optional and could have been set with `env.SENTRY_ORG` or `.sentryclirc` - for the sake of explicity, we remove these configuration methods
```
AbhiPrasad, line 37: I understand how this simplifies a lot, but a little concerned with removing the environmental variables config.

I'd like to keep it because I feel that it reduces the changes needed for users CI/CD setups, and also we don't need to update things like the Vercel integration.

I'm all for `sentryclirc` - bundler plugin users shouldn't have to be burdened with the fact that we use `sentry-cli` under the hood.

text/0086-sentry-bundler-plugins-api.md
AbhiPrasad: It might make sense to introduce just a top level flag to the plugin to enable/disable it, so that users don't have to conditionally add the plugin, DX feels a little better with that.

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Users will have to jump through a few hoops upgrading to the new Webpack plugin major just based on changing the options type.
+- Users may have to adjust their code AND build system if we drop environment variables as a way of configuring the plugin.
+- We are maintaining two systems of uploading source maps.
+
+# Unresolved questions
+
+- Should the options we intend to remove really be removed or do we see use cases for them?
+- Should we remove environment variables and `.sentryclirc` as a means to configure the plugin?
+- Should we deprecate the legacy source maps upload system entirely?
```
AbhiPrasad, line 103: This is too large of a breaking change IMO - let's just keep it around, like we do with `extras`.

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,103 @@
...
+Some removed options:
+
+- `dryRun` - Instead, just not use the plugin.
+- `configFile` - The config file is very intransparent in combination with the plugin and generally not well documented. Let's remove this feature to lower maintenance burden.
+- `injectReleasesMap` - Was used to support some micro frontends setups. This is not needed anymore with the introduction of debug IDs.
+- `releaseInjectionTargets` - This was used to disable/enable release injection for a particular set of files. This is not really useful and cannot be used in all bundlers (esbuild). Additionally, for technical reasons we need to inject the release into every processed module anyhow (it will be deduped) so this option simply doesn't make a lot of sense anymore.
+
+# Drawbacks
+
+- Users will have to jump through a few hoops upgrading to the new Webpack plugin major just based on changing the options type.
```
AbhiPrasad, line 95: Good migration guide will solve most of this.

text/0086-sentry-bundler-plugins-api.md
lforst: > I understand how this simplifies a lot, but a little concerned with removing the environmental variables config.
> 
> I'd like to keep it because I feel that it reduces the changes needed for users' CI/CD setups, and also we don't need to update things like the Vercel integration.

The way I thought about solving this particular issue would just be to have a very explicit `org: process.env.ORG_SLUG` in the docs/readmes/wizard snippets.

Something bout having env vars feed into bundler plugins feels odd... I get the convenience aspect though.

Do you think it would be an option for us to just unship env vars for now, and reintroduce them later if people see good use for them?

text/0086-sentry-bundler-plugins-api.md
lforst: Yup, that makes sense to me!

text/0086-sentry-bundler-plugins-api.md
lforst: > IIRC the dry run would for example print the chosen release, right?

Yes, but for printing debug information we already have the `debug` option.

> Maybe as a user, I want to know what release value the plugin/cli chose w/o uploading anything yet?

My view on this is that uploading source maps and bundling your project is **very** low stakes. You can delete releases and source maps with ease.

I see a use for `dryRun` options in tooling where you have a chance of majorly messing something up (e.g. deployment CLIs like terraform). Here it feels overkill.

Right now, I am still leaning towards deleting it for now, since it comes with maintenance effort that scales linearly with new features. I agree with the `disabled` flag though.

text/0086-sentry-bundler-plugins-api.md
AbhiPrasad: > Do you think it would be an option for us to just unship env vars for now, and reintroduce them later if people see good use for them?

My opinion is just to keep environmental variables, it's an acceptable amount of complexity in my eyes, esp for folks who have their CI pipelines set up that way. Just my opinion though, so I'm fine to go with what majority wants.

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,110 @@
...
+  debug?: boolean; // default: false - Whether to print debug information.
+  silent?: boolean; // default: false - Whether to print anything at all.
+  errorHandler?: (err: Error) => void; // when provided will swallow errors unless the provided function throws itself - By default any error will stop compilation to abort
+  telemetry?: boolean; // default: true - whether to send telemetry to Sentry
+  disable?: boolean; // default: false - just a quick way to entirely disable the plugin - purely for convenience
+
+  // General configuration for source maps uploading - if not provided, source map uploading will be disabled
+  sourcemaps?: {
+    assets: string | string[]; // globs pointing to the javascript files that should be uploaded to Sentry with their source maps - these are the built assets and not the source files
+    ignore?: string | string[]; // globs to exclude javascript files from being uploaded (will also not upload their source maps)
```
AbhiPrasad, line 51: Should assets/ignore always be an array?

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Users will have to jump through a few hoops upgrading to the new Webpack plugin major just based on changing the options type. We have the theory that a good migration guide will solve most of this though.
+- Users may have to adjust their code AND build system if we drop environment variables as a way of configuring the plugin.
+- We are maintaining two systems of uploading source maps.
+
+# Unresolved questions
+
+- Should the options we intend to remove really be removed or do we see use cases for them?
+- Should we remove environment variables and `.sentryclirc` as a means to configure the plugin?
+- ~~Should we deprecate the legacy source maps upload system entirely?~~ No. We think it still has its use cases so we keep it.
```
AbhiPrasad, line 110: Also keeping this around helps self-hosted people!

AbhiPrasad (commented):
Do we need to add notes about Node versions? I thought unplugin is Node 14+, which may cause issues.

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,110 @@
...
+    finalize?: boolean; // default: true
+    dist?: string; // Can be used to segment releases further
+    vcsRemote?: string; // default: 'origin' - Version control system remote name
+    setCommits?: SetCommitsOptions;
+    deploy?: DeployOptions;
+    legacySourcemaps?: {
+      // Pre-debugid way of uploading source maps - for users that for whatever reason cannot inject code into their bundles
+      include: string | IncludeEntry | Array<string | IncludeEntry>;
+      cleanArtifacts?: boolean; // default: false - Wether to delete artifacts previously uploaded to the release
+    };
```
Lms24, line 70: Looks like a clean fallback to me 👍 Big +1 for removing the top-level `IncludeEntry` options!

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Some removed options:
+
+- `dryRun` - Current assumption is that this option is unnecessary:
+  - Debug information can be printed with the `debug` option instead.
+  - Executing the plugin is **very** low stakes. Releases and artifact bundles can be deleted with ease and without any lasting side-effects.
+  - This option comes with maintenance effort that scales linearly with new features and there is a high chance implementing is overlooked for new features.
+  - We introduce a `disabled` option instead that still is a convenient way of disabling the plugin without any continuous maintenance effort.
+- `configFile`
```
Lms24, line 94: +1! Based on my understanding, removing this option would anyways still allow the CLI to access a `.sentryclirc` file if one is present.

text/0086-sentry-bundler-plugins-api.md
lforst: Yeah let's keep this for convenience.

text/0086-sentry-bundler-plugins-api.md
lforst: I guess so. In my "perfect world" it wouldn't even pick up that file but whatever. If users discover the `.sentryclirc` feature they have gone so deep they deserve to use it.

text/0086-sentry-bundler-plugins-api.md
lforst: Ok let's keep them then!

text/0086-sentry-bundler-plugins-api.md
```diff
@@ -0,0 +1,110 @@
...
+  silent?: boolean; // default: false - Whether to print anything at all.
+  errorHandler?: (err: Error) => void; // when provided will swallow errors unless the provided function throws itself - By default any error will stop compilation to abort
+  telemetry?: boolean; // default: true - whether to send telemetry to Sentry
+  disable?: boolean; // default: false - just a quick way to entirely disable the plugin - purely for convenience
+
+  // General configuration for source maps uploading - if not provided, source map uploading will be disabled
+  sourcemaps?: {
+    assets: string | string[]; // globs pointing to the javascript files that should be uploaded to Sentry with their source maps - these are the built assets and not the source files
+    ignore?: string | string[]; // globs to exclude javascript files from being uploaded (will also not upload their source maps)
+    rewriteSources?: (source: string) => string; // Hook to rewrite individual entries in the sourcemaps's `sources` field. By default, if not defined, the plugin will try to rewrite the entries to be relative to `process.cwd()`.
```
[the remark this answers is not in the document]
lforst: Hm, I don't wanna overengineer this right now. I think probably nobody will use this option except for us in SDKs that have automatic source map uploading capabilities like in nextjs and sveltekit. There, the function is enough.

I like the suggestion to pass the entire map!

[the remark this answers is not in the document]
lforst: Adding more options here is backwards compatible.

lforst:
> Do we need to add notes about Node versions? I thought unplugin is Node 14+, which may cause issues.

Not in this RFC. We will add a note in the changelog + migration guide.

lforst:
Ok, I think everything has been addressed. @Lms24 There is one open point about the rewriting - lmk if you think this is blocking, otherwise can anybody slap approve? cc @AbhiPrasad

text/0086-sentry-bundler-plugins-api.md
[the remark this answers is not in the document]
Lms24: Yeah, fair enough and agreed, we can add this later, good point.

Lms24:
>. @Lms24 There is one open point about the rewriting - lmk if you think this is blocking

Nope, not blocking.

Lms24 (approved):

AbhiPrasad (approved):
📝

PR #88, opened 2023-04-24
rfc(feature): Fix Memory Limitiations in Session Replay's Access Pattern

cmanallen:
TODO. [Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/fix-memory-limitiations-in-session-replays-access-pattern/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md)

### DIFF
--- a/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
+++ b/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
@@ -65,7 +65,7 @@ Our current subquery solution works very well. However, there are escape hatches
 
 **Proposal**
 
-The number of rows aggregated can be reduced by restricting the maximum time range we query over. We should validate the timestamp range such that it does not exceed a 24-hour period. This would satify every organization which ingests fewer than 1 billion replay-segments every 90 days.
+The number of rows aggregated can be reduced by restricting the maximum time range we query over. We should validate the timestamp range such that it does not exceed a 24-hour period. This would satisfy every organization which ingests fewer than 1 billion replay-segments every 90 days.
 
 **Drawbacks**
 
--- a/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
+++ b/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
@@ -32,10 +32,6 @@ The non-aggregated query does not allow **exclusive** filter conditions against 
 
 The non-aggregated query does not allow multiple, **inclusive** filter conditions against **non-static** columns. For example, we can not say "find a replay where this url exists and this other url exists". It will find rows which have both urls but not replays which have both urls. Transforming the `AND` operator to an `OR` operator does not satisfy the condition because it will match replays which contain one of the urls - not both.
 
-# How To Read Each Proposal
-
-Each proposal is prefixed with an assumption and a solution that is perceived to logically follow from it. The assumption does not exist to state a fact. It only exists to say that the world must be so in order for the following solution to be applicable. If the assumption is incorrect then the solution is discarded. If the assumption is correct then the solution must be evaluated on its merits before acceptance or rejection.
-
 # Options Considered
 
 Any option may be accepted in whole or in part. Multiple options can be accepted to achieve the desired outcome. The options are ordered from perceived ease to perceived difficulty.
@@ -44,7 +40,7 @@ Any option may be accepted in whole or in part. Multiple options can be accepted
 
 **Proposal**
 
-The current subquery solution is the appropriate way to scan our dataset. However, there are escape hatches which require us to issue an aggregation query over the whole dataset. Those escape hatches should be closed.
+Our current subquery solution works very well. However, there are escape hatches which require us to issue an aggregation query over the whole dataset. Those escape hatches should be closed.
 
 - Remove ability to filter and sort by aggregated values (e.g. count_errors, activity_score).
 - Remove ability to filter by negation values which change (e.g. urls).
@@ -67,7 +63,7 @@ The current subquery solution is the appropriate way to scan our dataset. Howeve
 
 **Proposal**
 
-The current aggregation query is the appropriate way to model the data. However, the range we allow it to scan is too large. We should validate the timestamp range such that it does not exceed a 24-hour period. This would satify every organization which ingests fewer than 1 billion replay-segments every 90 days.
+The number of rows aggregated can be reduced by restricting the maximum time range we query over. We should validate the timestamp range such that it does not exceed a 24-hour period. This would satify every organization which ingests fewer than 1 billion replay-segments every 90 days.
 
 **Drawbacks**
 
@@ -82,7 +78,7 @@ The current aggregation query is the appropriate way to model the data. However,
 
 **Proposal**
 
-The current aggregation query is the appropriate way to model the data. However, the range we allow it to scan is too large. For select queries the backend can issue multiple queries on subsets of the range. For example, if we assume that no sort value was provided or that the sort value was applied to the timestamp column then the back end can transparently query a subset of the window attempting to populate the result set without querying the entire range.
+The number of rows aggregated can be reduced by restricting the maximum time range we query over. For select queries the backend can issue multiple queries on subsets of the range. For example, if we assume that no sort value was provided or that the sort value was applied to the timestamp column then the back end can transparently query a subset of the window attempting to populate the result set without querying the entire range.
 
 **Drawbacks**
 
@@ -95,7 +91,12 @@ The current aggregation query is the appropriate way to model the data. However,
 
 **Proposal**
 
+- Normalize schema.
+- Use joins.
+
 **Drawbacks**
+
+- Performance.
 
 **Questions**
 
@@ -103,7 +104,7 @@ The current aggregation query is the appropriate way to model the data. However,
 
 **Proposal**
 
-The current aggregation query is not the appropriate way to model the data. Aggregating in Clickhouse has proven to be too expensive to be scalable. Instead of aggregating in ClickHouse we should maintain a stateful representation of the Replay in an alternative service (such as a key, value store). Additionally, we will migrate to the "VersionedCollapsingMergeTree" table engine and write our progress incrementally.
+Instead of aggregating in ClickHouse we should maintain a stateful representation of the Replay in an alternative service (such as a key, value store). Additionally, we will migrate to the "VersionedCollapsingMergeTree" table engine and write our progress incrementally.
 
 The versioned engine works as follows:
 
@@ -142,7 +143,7 @@ The ingestion process can be described as follows:
 
 **Proposal**
 
-The current aggregation query is not the appropriate way to model the data. Aggregating in Clickhouse has proven to be too expensive to be scalable. Instead we should aggregate our replays in a stateful service such as Apache Spark before writing the final result to ClickHouse.
+Instead of aggregating in ClickHouse we should aggregate our replays in a stateful service such as Apache Spark before writing the final result to ClickHouse.
 
 TODO: Josh to fill in specifics.
 
@@ -160,7 +161,7 @@ TODO: Josh to fill in specifics.
 
 **Proposal**
 
-Older versions of ClickHouse are not sufficiently sophisticated for our use case. Upgrading to a newer version of ClickHouse will enable us to use experimental features such as "Live View" and "Window View".
+Upgrading to a newer version of ClickHouse will enable us to use experimental features such as "Live View" and "Window View".
 
 **Drawbacks**
 
@@ -174,7 +175,7 @@ Older versions of ClickHouse are not sufficiently sophisticated for our use case
 
 **Proposal**
 
-ClickHouse is not a sufficiently sophisticated database for our use case. OLAP Databases such as Apache Pinot support upserts which appear to be a key requirement for the Replays product. An aggregation state schema can be defined for merging columns in real-time https://docs.pinot.apache.org/basics/data-import/upsert.
+OLAP Databases such as Apache Pinot support upserts which appear to be a key requirement for the Replays product. An aggregation state schema can be defined for merging columns in real-time https://docs.pinot.apache.org/basics/data-import/upsert.
 
 **Drawbacks**
 
@@ -194,7 +195,7 @@ ClickHouse is not a sufficiently sophisticated database for our use case. OLAP D
 
 **Proposal**
 
-OLAP databases are not suitable for our use case. OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be a key requirement for the Replays product.
+OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be a key requirement for the Replays product.
 
 **Drawbacks**
 
--- a/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
+++ b/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
@@ -199,11 +199,11 @@ OLAP Databases such as Apache Pinot support upserts which appear to be a key req
 
 **Proposal**
 
-OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be a key requirement for the Replays product. Our scale is small enough that a shared PostgreSQL database could handle it. Read volume is low relative to write volume. We could optimize our database for this use case.
+OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be a key requirement for the Replays product. Our scale is small enough that a sharded PostgreSQL database could handle it. Read volume is low relative to write volume. We could optimize our database for this use case.
 
 **Drawbacks**
 
-- Race conditions will require single-threaded processing of replay events.
+- Race conditions will require single-threaded processing of like replay-id events.
 - Duplicate messages will necessitate ordering requirements.
 - Always a possibility for dropped and duplicated data regardless of safe guards.
 - AlloyDB is still in developer preview on Google Cloud.
--- a/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
+++ b/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
@@ -218,14 +218,33 @@ OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be
 
 **Proposal**
 
-Use a cron job, which runs at some interval (e.g. 1 hour), that would select the finished replays from the last hour, aggregate them, and write them to a materialized view or some destination table. We then alter the index page to reflect two different dataset. The "live" dataset and the "video-on-demand" dataset. A "live" page would fetch replays from the last hour. A "video-on-demand" dataset would function similarly to the current replays index page however it would only contain data that has been aggregated by the cron job.
+Use a cron job, which runs at some `interval` (e.g. 1 hour), that would select the finished replays from the last `interval`, aggregate them, and write them to a materialized view or some destination table. We then alter the index page to reflect two different dataset. The "live" dataset and the "video-on-demand" dataset. A "live" page would fetch replays from the last `interval`. A "video-on-demand" dataset would function similarly to the current replays index page however it would only contain data that has been aggregated by the cron job.
 
 **Drawbacks**
 
-- The cron job is another process which can fail. This would not lead to data loss per say but depending on the implementation of the query we may skip rows when recovering from an outage.
-  - A record could be kept of the time-windows the cron job has processed. This record could exist in ClickHouse or another datastore. By using this dataset the cron job could recover from an outage without skipping rows.
+- Requires product changes.
+- Requires manual management of a secondary dataset.
+- Introduces fragility with a secondary post-processing step.
 
 **Questions**
+
+- Colton Allen asks: What happens if the cron job runs once, a replay is aggregated and stored, the cron runs a second time after its interval but finds new rows for the previously aggregated replay?
+  - Insert a new aggregation row. Do not group by or merge. On the read side we still WHERE query but accept that some replays that _should_ match a given condition are not returned.
+  - We will need to validate our success rate to make sure we're writing as close to one row per replay-id as possible.
+- Colton Allen asks: What happens if a new row is written to the table with an old timestamp? The aggregation process could have already run for that timestamp range.
+  - An old timestamp does not necessarily indicate an old replay. The client clock could be incorrect.
+    - Implementing a server-generated timestamp column could help with this.
+  - An old timestamp does not necessarily indicate an incorrect clock. The message could have been backlogged on a consumer.
+    - Write a new aggregation state row. We should tolerate multiple aggregation states for a replay.
+- Colton Allen asks: What happens if the cron job is down for an extended period of time?
+  - We will need to write a log of each successful run.
+    - PostgreSQL could store run_id, from_ts, to_ts columns.
+    - We query PostgreSQL for the max(to_ts) value and then bootstrap our process from there.
+  - If a cron fails mid-run then no log is written.
+    - Duplicate aggregations are possible and should be tolerated.
+    - Duplicate aggregations can be merged asynchronously and should not impact the user-experience.
+  - The cron should attempt to catch up sequentially. If the cron was down for `m` \* `n` hours and the interval is `n` hours then we will need to call the aggregation function `m` times.
+    - If the process fails then it will restart from the previously stored max(to_ts).
 
 # Selected Outcome
 
--- a/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
+++ b/text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
@@ -28,6 +28,8 @@ We have introduced a second, non-aggregated query which precedes our main aggreg
 
 **Drawbacks**
 
+There are aggregated values which users want to sort and filter on. Columns such as count_errors and activity which are prominently featured on our index page. Sorting and filtering by these values disables the preflight query and only runs the aggregated query.
+
 The non-aggregated query does not allow **exclusive** filter conditions against **non-static** columns. For example, we can not say "find a replay where this url does not exist". The query will find _rows_ where that condition is true but it will not find _replays_ where that is condition true.
 
 The non-aggregated query does not allow multiple, **inclusive** filter conditions against **non-static** columns. For example, we can not say "find a replay where this url exists and this other url exists". It will find **ROWS** which have both urls but not **REPLAYS** which have both urls. Transforming the `AND` operator to an `OR` operator does not satisfy the condition because it will match replays which contain one of the urls - not both.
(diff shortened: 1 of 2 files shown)


text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,216 @@
...
+**Proposal**
+
+Instead of aggregating in ClickHouse we should aggregate our replays in a stateful service such as Apache Spark before writing the final result to ClickHouse.
+
+TODO: Josh to fill in specifics.
+
+**Drawbacks**
+
+- Replays are not available until they have finished.
+- There are no existing Apache Spark installations within the Sentry org.
```
JoshFerge, line 156: the datateam manages several spark jobs! cc @trillville

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,216 @@
...
+- The Replays team is not large enough or experienced enough to manage a Spark installation.
+  - This would require another team assuming the burden for us.
+  - Otherwise, additional budget would need to be allocated to the Replays team to hire outside experts.
```
JoshFerge, line 159: google's dataproc makes this pretty straightforward, hopefully we wouldn't need too much outside help other than ops https://cloud.google.com/dataproc?utm_source=google&utm_medium=cpc&utm_campaign=na-US-all-en-dr-bkws-all-all-trial-e-dr-1605212&utm_content=text-ad-none-any-DEV_c-CRE_547545257220-ADGP_Desk%20%7C%20BKWS%20-%20EXA%20%7C%20Txt%20_%20Data%20Analytics%20_%20Dataproc_Data-KWID_43700065464355775-kwd-313714756996&utm_term=KW_google%20cloud%20dataproc-ST_google%20cloud%20dataproc&gclid=CjwKCAjw9J2iBhBPEiwAErwpeX4oajs8rvfjlwky-3tzjGeaLYmzfcu8LmSNZs15L8B5NckSAkMExhoCT-AQAvD_BwE&gclsrc=aw.ds

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,216 @@
...
+- To eliminate the need for grouping queries we would need to supply the `FINAL` keyword.
+- Our column types are not ideal for this use case.
+
+**Questions**
+
+### 6. Stateful Streaming using Apache Spark or Similar Service
+
+**Proposal**
+
+Instead of aggregating in ClickHouse we should aggregate our replays in a stateful service such as Apache Spark before writing the final result to ClickHouse.
```
JoshFerge, line 149: With spark, we can view our Kafka topic as a streaming DataFrame. Spark has a mechanism for stateful streaming based on windows, https://spark.apache.org/docs/latest/structured-streaming-programming-guide.html#window-operations-on-event-time. We would use a "Session Window", with the timeout being the same as the one we define on the client to aggregate data per replay-id. we'd then take advantage of watermarking to ensure that late data still gets grouped accordingly. As you mentioned, the downsides are:

1. we'd still need to handle in progress replays somehow, as with "update mode", it only emits the row once finished.
2. once stateful streaming is introduced, the probability of data loss from an outage goes up, as if there are problems upstream we have to take action to pause the job, etc.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,216 @@
...
+
+- Race conditions will require single-threaded processing of replay events.
+- Duplicate messages will necessitate ordering requirements.
+- Always a possibility for dropped and duplicated data regardless of safe guards.
+- AlloyDB is still in developer preview on Google Cloud.
+- We need to re-write our application logic for querying the datastore.
+- Migration pains.
+
+**Questions**
+
```
JoshFerge, line 213: I think one other idea would be to do our own managing of materialized views. how this would work:

1. we have a cron that runs every 24 hours, and selects all finished replays that occured in the last 24 hours, aggregates them, and writes them to a materialized view table. when we query replays, we query both the "last 24 hours table" and "beyond 1 day pre-aggregated table" and join the results.

This could be extended to run every hour or quicker, to try and pre-aggregate all completed replays. someone with more clickhouse knowledge could perhaps expand upon this.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: That's good to know!  Definitely makes Spark more appealing.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,216 @@
...
+- Replays are not available until they have finished.
+- There are no existing Apache Spark installations within the Sentry org.
+- The Replays team is not large enough or experienced enough to manage a Spark installation.
+  - This would require another team assuming the burden for us.
+  - Otherwise, additional budget would need to be allocated to the Replays team to hire outside experts.
+
+**Questions**
+
+### 7. Upgrade ClickHouse Version to Utilize Experimental Features
```
bruno-garcia, line 163: I hope we can kick this problem down the road until this is a feasible solution :)

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
bruno-garcia: > google's dataproc makes this pretty straightforward
what's the alternative for self-hosted? Can we have an adapter so folks can plug spark?

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Thanks!  Updated that section to reflect this.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Added.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,232 @@
...
+
+The Session Replay index page will run out of memory when processing queries for our largest customers. This document contains proposed solutions for addressing this short-coming.
+
+# Motivation
+
+Our goal is to make the Session Replay product highly scalable. We expect changes to the product will eliminate our OOM issues and increase performance for customers with large amounts of ingested data.
+
+# Background
+
+The Session Replay data model is different from most at Sentry. Replays are received in parts referred to as "segments". One row does not represent one replay. Instead many rows are aggregated together to represent a single replay. When we ask a question such as "which replays did not visit this url?", we have to aggregate every row in the database (minus whatever rows were reduced by conditions in the WHERE clause). As it turns out the number of rows is very large and the amount of data is significant enough that we run out of memory on large customers.
```
evanh, line 16: It's not actually that different. One issue is made up of many events, and one transaction has many events as well.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,232 @@
...
+### 2. Reduce the Scan Range on the Front End
+
+**Proposal**
+
+The number of rows aggregated can be reduced by restricting the maximum time range we query over. We should validate the timestamp range such that it does not exceed a 24-hour period. This would satify every organization which ingests fewer than 1 billion replay-segments every 90 days.
+
+**Drawbacks**
+
+- To search for a unique value a user would need to issue a query for each day of the retention period or until the value was found.
+- Small customers would not see any replays on their index page due to the limited window size.
```
evanh, line 73: I think you could probably issue multiple queries in this case.

1) Fetch a small initial window (past 24 hours).
2) If you get more than N results, stop.
3) If you get less than N results, query again for another time window (last week - last 24 hours).
4) Repeat 3 until you get N results.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,232 @@
...
+- Requires O(retention_days) queries to satisfy the result set in the worst case.
+- Adds an additional layer of complexity and does not solve our OOM issue.
+
+**Questions**
+
+### 4. Normalize Schema and Remove Snuba Join Restriction
+
+**Proposal**
+
+- Normalize schema.
```
evanh, line 96: Can you elaborate on this?

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
evanh: Don't hold your breath.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,232 @@
...
+  - To solve this we will need to partition our Kafka messages by replay-id and process sequentially.
+  - This has scalability limitations but those limitations are likely to be less than existing limitations.
+  - Aggregation states can be updated atomicly with some databases. See "Use Alternative OLAP Database" proposal.
+  - Alternatively we can tolerate losing aggregation states and continue using parallel consumers.
+- Kafka will sometimes produce duplicate messages. If we assume order we can set a requirement on segment_id > previous_segment_id.
+  - Order can not be assumed. Valid aggregation states could be lost.
+  - A delay in scheduling in Relay could cause segment_id 1 to be processed prior to segment_id 0.
+  - This is expected to be uncommon but not impossible.
+  - Alternatively, we could tolerate duplicates and accept that segment_count and other values have some margin of error.
+- To eliminate the need for grouping queries we would need to supply the `FINAL` keyword.
```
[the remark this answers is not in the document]
cmanallen: Ah I see.  Then this likely won't work if we're filtering/sorting aggregated, in-memory views.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: 🤣

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+- RFC PR: https://github.com/getsentry/rfcs/pull/88
+- RFC Status: draft
+
+# Summary
+
+The Session Replay index page will run out of memory when processing queries for our largest customers. This document contains proposed solutions for addressing this short-coming.
+
+# Motivation
+
+Our goal is to make the Session Replay product highly scalable. We expect changes to the product will eliminate our OOM issues and increase performance for customers with large amounts of ingested data.
```
jas-kas, line 12: I didn't see this in the solution options considered if I understand them correctly: But would limiting the number of replays shown on the Index page materially improve the OOM issues or probability of them occurring?

Right now we display 50 replays upon load of this page which is * a lot * and seems unnecessary. I assume the number of replays we display on this view, will result in fetching more info on sort/filter actions and retention period changes as well.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Questions**
+
+- Colton Allen asks: Is it possible to write to tables partitioned by the hour? When the hour expires the table is indexed for fast read performance. Replays on the most recent hour partition would be unavailable while we are writing to it. Does PostgreSQL expose behavior that allows us to query over multiple partitions transparently? Is this even necessary given our scale? Currently processing 200 messages per second.
+
+### 10. Manually Manage An Aggregated Materialized View
+
+**Proposal**
+
+Use a cron job, which runs at some `interval` (e.g. 1 hour), that would select the finished replays from the last `interval`, aggregate them, and write them to a materialized view or some destination table. We then alter the index page to reflect two different dataset. The "live" dataset and the "video-on-demand" dataset. A "live" page would fetch replays from the last `interval`. A "video-on-demand" dataset would function similarly to the current replays index page however it would only contain data that has been aggregated by the cron job.
```
jas-kas, line 221: I can't speak to the details of the cron job described here in this solution, but I would be in favour of a solution proposal that does not display "in progress" replays on the Index page by default -- as this leads to many short (and 'unfinished') replays for customers that have a high volume.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: > I didn't see this in the solution options considered if I understand them correctly: But would limiting the number of replays shown on the Index page materially improve the OOM issues or probability of them occurring?

Unfortunately no.  Whats displayed and what was actually aggregated in the database are not the same in this case.  The "scan range" (the stats period value: 14 days, 24 hours, 90 days, etc) is what creates these OOM issues.

> Right now we display 50 replays upon load of this page which is * a lot * and seems unnecessary.

If it makes the product better we should do it!  Reducing the row set will have a positive impact on performance.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: > I would be in favour of a solution proposal that does not display "in progress" replays on the Index page by default

This is good to know.  This gives us a lot of flexibility to solve the problem.

From a product perspective I agree as well.  Some of our large customers have low utility index pages because of the number of new replays.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+  - Simplify with cron job: If the cron job has an outage it can be resumed from a save point. However, if the Kafka consumer is down then we need to pause the cron so that it does not write incomplete aggregation rows. Both use cases require a piece of infrastructure to be paused in the event of an outage.
+    - If our Kafka consumer encounters a RabbitMQ failure it can pause processing, publish the message to a DLQ, or publish the message back to itself to retry automatically at a later time.
+
+### 11. Configure ClickHouse to Optimize Aggregations
+
+**Proposal**
+
+Use `optimize_aggregation_in_order` to limit the number of rows we need to aggregate. In testing this eliminates OOM errors. Queries complete somewhat speedily.
+
+**Drawbacks**
```
bmckerry, line 282: Would be good to mention here how this may show incorrect data for replays with a timestamp around the partition cutoff time

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+- Replays can be "idle" for up to 15 minutes before sending a new segment when user activity resumes.
+- Replays can be up to 1 hour in duration.
+
+# Existing Solution
+
+We have introduced a second, non-aggregated query which precedes our main aggregation query. This query will trigger if we are exclusively querying by static values which are consistent throughout the lifespan of a Replay. It will return a list of replay IDs which we can then pass to our main aggregation query. The aggregation query is then only responsible for aggregating a subset of replays (the pagination limit). This optimization allows us to query very efficiently while still providing aggregated results to the end user.
+
+**Drawbacks**
+
+There are aggregated values which users want to sort and filter on. Columns such as count_errors and activity which are prominently featured on our index page. Sorting and filtering by these values disables the preflight query and only runs the aggregated query.
```
volokluev, line 31: Examples of these queries would be very helpful to designing a performant solution. as someone who doesn't use replay, it's hard for me to map these abstract concepts onto a material reality.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+
+The non-aggregated query does not allow **exclusive** filter conditions against **non-static** columns. For example, we can not say "find a replay where this url does not exist". The query will find _rows_ where that condition is true but it will not find _replays_ where that is condition true.
```
volokluev, line 33: What makes a column `static` or not?

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
+- Start Date: 2023-04-24
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/88
+- RFC Status: draft
+
+# Summary
+
+The Session Replay index page will run out of memory when processing queries for our largest customers. This document contains proposed solutions for addressing this short-coming.
+
+# Motivation
```
volokluev, line 10: What is the expected timeline of this? Those constraints would be good to know when evaluating solutions

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Proposal**
+
+OLTP databases such as PostgreSQL and AlloyDB support updates which appear to be a key requirement for the Replays product. Our scale is small enough that a sharded PostgreSQL database could handle it. Read volume is low relative to write volume. We could optimize our database for this use case.
+
+**Drawbacks**
+
+- Race conditions will require single-threaded processing of like replay-id events.
+- Duplicate messages will necessitate ordering requirements.
+- Always a possibility for dropped and duplicated data regardless of safe guards.
```
volokluev, line 208: what is the tolerance for dropped and duplicated data?

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,288 @@
...
+  - Outages: If Relay is down new messages are rejected and the queues are drained. If the Kafka consumer is down then the messages backlog until the consumer resumes processing. If RabbitMQ is down the Kafka consumer must pause.
+  - Simplify with cron job: If the cron job has an outage it can be resumed from a save point. However, if the Kafka consumer is down then we need to pause the cron so that it does not write incomplete aggregation rows. Both use cases require a piece of infrastructure to be paused in the event of an outage.
+    - If our Kafka consumer encounters a RabbitMQ failure it can pause processing, publish the message to a DLQ, or publish the message back to itself to retry automatically at a later time.
+
+### 11. Configure ClickHouse to Optimize Aggregations
+
+**Proposal**
+
+Use `optimize_aggregation_in_order` to limit the number of rows we need to aggregate. In testing this eliminates OOM errors. Queries complete somewhat speedily.
```
volokluev, line 280: have you tested this in production?

evanh (commented):
One thing that is missing from this is any solution involving the client side SDK. I think we've talked about this on Slack but documenting that discussion would be useful.

For example, why doesn't the client SDK set a "final" flag on the last segment of a replay? The SDK could also update each segment with the data from all the previous segments, so that the "final" segment would effectively be a fully formed replay sent all at once. We could then store that directly in a replay table and/or filter on the "final" flag.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,324 @@
...
+The SDK can maintain a buffer of the replay metadata. Once the replay has finished the SDK will flush that buffered data on the final segment. The segment marked as "final" can then be fetched without aggregation.
+
+**Drawbacks**
+
+- Requies SDK upgrade.
+- API will need to fallback to aggregating behavior if no final segments can be found (or otherwise detect old SDK usage prior to querying).
+
+**Questions**
+
+- Is it possible for the SDK to know when its done and mark the request object with a final attribute?
```
cmanallen, line 299: @billyvg Can we assess this?  Is it possible for the SDK to _know_ when the replay has finished?  If so do we have any idea how reliable it is?

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Yes.  Didn't work with custom ordering which makes it non-viable.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Discussed on call.  Static means the value of the column on segment 0 can be presumed to be the value for every segment.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -283,6 +283,42 @@ Use `optimize_aggregation_in_order` to limit the number of rows we need to aggre
...
+### 12. Aggregate Replay Event Metadata in the SDK and Set Final Flag
+
+**Proposal**
+
+The SDK can maintain a buffer of the replay metadata. Once the replay has finished the SDK will flush that buffered data on the final segment. The segment marked as "final" can then be fetched without aggregation.
+
+**Drawbacks**
+
+- Requies SDK upgrade.
+- API will need to fallback to aggregating behavior if no final segments can be found (or otherwise detect old SDK usage prior to querying).
```
evanh, line 297: This is only true if you want to show "in progress" replays. If we exclude those then this isn't a problem.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -283,6 +283,42 @@ Use `optimize_aggregation_in_order` to limit the number of rows we need to aggre
...
+**Drawbacks**
+
+- Likely requires the use of `FINAL` to fetch the most recent row.
+- Not backwards compatible. Requires SDK upgrade.
+- Requires table migration.
+  - Either double writer or materialized view.
+
+**Questions**
+
+- Will using `FINAL` be a problem? Is it a big deal relative to the problems we're experiencing currently?
```
evanh, line 320: In this context, the FINAL keyword will have the same effect as the `GROUP BY replay_id`. Under the hood CH is doing the same thing, collapsing all the parts in memory to collect the correct result.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Old SDKs won't set the `final` attribute and will require us to aggregate.  Eventually we could drop this requirement but only after some portion of customers upgraded.

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
cmanallen: Are we sure its exactly the same?  My reading of the documentation suggests that it would return the highest version row whereas `GROUP BY` will return an aggregation of the non-collapsed rows.

cmanallen: Specifically here: https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree#ver

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+LIMIT 10
+OFFSET 0
+GRANULARITY 3600
+```
+
+# Options Considered
+
+Any option may be accepted in whole or in part. Multiple options can be accepted to achieve the desired outcome.
+
+### 1. Change the Product's Query Access Pattern
```
volokluev, line 129: Possibility: High
Customer Impact: High (bad)
Maintainability: High

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+    - Filtering by error_id OR url would be allowed.
+
+**Drawbacks**
+
+- Significant reduction in index page quality.
+- Removes the most important features of our index page. Count-errors sort, activity sort, duration sort and filter.
+
+**Questions**
+
+### 2. Reduce the Scan Range on the Front End
```
volokluev, line 152: Infra Burden: Low
Customer Impact: Medium
Maintainability: Low

Notes: most people will query over 14 day window

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+**Drawbacks**
+
+- To search for a unique value a user would need to issue a query for each day of the retention period or until the value was found.
+- Small customers would not see any replays on their index page due to the limited window size.
+  - Necessitates a special flag for large customers to enable this optimization.
+  - We may not know who needs this flag in advance and we may present a degraded customer experience without realizing.
+
+**Questions**
+
+### 3. Reduce the Scan Range on the Back End
```
volokluev, line 167: Scratch this, it probably won't solve the problem

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+The number of rows aggregated can be reduced by restricting the maximum time range we query over. For select queries the backend can issue multiple queries on subsets of the range. For example, if we assume that no sort value was provided or that the sort value was applied to the timestamp column then the back end can transparently query a subset of the window attempting to populate the result set without querying the entire range.
+
+**Drawbacks**
+
+- Requires O(retention_days) queries to satisfy the result set in the worst case.
+- Adds an additional layer of complexity and does not solve our OOM issue.
+
+**Questions**
+
+### 4. Normalize Schema and Remove Snuba Join Restriction
```
volokluev, line 180: 🔪  Remove

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+- Use joins.
+
+**Drawbacks**
+
+- Performance.
+- Requires Snuba work.
+
+**Questions**
+
+### 5. Migrate to CollapsingMergeTree Table Engine and Pre-Aggregate Data in Key-Value Store
```
volokluev, line 194: Infra Burden: Medium
```suggestion
### 5. Make a replays table as opposed to a replay_segments table
```
Infra burden: High
Customer Impact: Low
Maintainability: High

Notes: If we have a `replays` table instead of a `replay_segments` table we can filter better even if we aggregate the same amount of replays

text/0088-fix-memory-limitiations-in-session-replays-access-pattern.md
```diff
@@ -0,0 +1,414 @@
...
+- Always a possibility for dropped and duplicated data regardless of safe guards.
+- AlloyDB is still in developer preview on Google Cloud.
+- We need to re-write our application logic for querying the datastore.
+- Migration pains.
+
+**Questions**
+
+- Colton Allen asks: Is it possible to write to tables partitioned by the hour? When the hour expires the table is indexed for fast read performance. Replays on the most recent hour partition would be unavailable while we are writing to it. Does PostgreSQL expose behavior that allows us to query over multiple partitions transparently? Is this even necessary given our scale? Currently processing 200 messages per second.
+
+### 10. Manually Manage An Aggregated Materialized View
```
volokluev, line 305: Infra burden: medium (cron job generated inserts is a new mechanism)
Maintainability: Medium
Customer Impact: Low/medium (depending on if the product changes)

Questions:
Do we need to merge the last hour segments with aggregated?

PR #91, opened 2023-05-15
Simplified Upload Tokens for CI Usage

mitsuhiko:
This RFC proposes adding org level tokens that contain meta information which simplifies user flows and lets clients forgo of necessary other information.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/feature/upload-tokens/text/0091-ci-upload-tokens.md)

Refs the following issues:
* https://github.com/getsentry/sentry/issues/49084
* https://github.com/getsentry/sentry/issues/48690

### DIFF
--- a/text/0091-ci-upload-tokens.md
+++ b/text/0091-ci-upload-tokens.md
@@ -79,9 +79,13 @@ We want to encode certain information into the tokens.  The following attributes
 * `sentry_org`: a token is uniquely bound to an org, so the slug of that org is also always
   contained.  Note that the slug is used rather than an org ID as the clients typically
   need these slugs to create API requests.
-* `sentry_projects`: a token can be valid for more than one project.  For operations such as
-  source map uploads it's beneficial to issue tokens bound to a single project in which
-  case the upload experience does not require providing the project slugs.
+
+Potential fields:
+
+* `sentry_projects`: normally a token is valid for the entire org, but it could potentially
+  be restricted.  For operations such as source map uploads it might be interesting to issue
+  tokens bound to a single project in which case the upload experience does not require
+  providing the project slugs.  However we currently do not want to start with this.
 
 For JWT the facts could be encoded as custom claims:
 
@@ -90,8 +94,7 @@ For JWT the facts could be encoded as custom claims:
     "iss": "sentry.io",
     "iat": 1684154626,
     "sentry_site": "https://myorg.sentry.io/",
-    "sentry_org": "myorg",
-    "sentry_projects": ["myproject"]
+    "sentry_org": "myorg"
 }
 ```
 
@@ -115,9 +118,6 @@ to use this information to route requests.  For the keys the following rules app
 * `org`: a token is uniquely bound to an org, so the slug of that org is also always
   contained.  Note that the slug is used rather than an org ID as the clients typically
   need these slugs to create API requests.
-* `projects`: a token can be valid for more than one project.  For operations such as
-  source map uploads it's benefitial to issue tokens bound to a single project in which
-  case the upload experience does not require providing the project slugs.
 
 An example of this with a JWT token:
 
@@ -130,7 +130,7 @@ An example of this with a JWT token:
   'iat': 1684154626,
   'sentry_site': 'https://myorg.sentry.io/',
   'sentry_org': 'myorg',
-  'sentry_projects': ['myproject']
+  'sentry_projects': ['myproject']  # not used currently
 }
 ```
 
@@ -160,6 +160,7 @@ module.exports = {
   devtool: "source-map",
   plugins: [
     new SentryWebpackPlugin({
+      url: "https://sentry.io/",  // defaults to sentry.io
       org: "demo-org",
       project: "demo-project",
       include: "./dist",
@@ -182,22 +183,21 @@ module.exports = {
   plugins: [
     new SentryWebpackPlugin({
       authToken: "AUTO GENERATED TOKEN HERE",
+      project: "demo-project",
       include: "./dist",
     }),
   ],
 };
 ```
 
-There are however some cases where manual configuration would still be necessary:
+Some manual configuration remains as we still want ask users to provide the slug
+of the project explicitly to allow cross-org token issuance by default.
 
-* **Multi project tokens:** if a token contains more than one project, it's unclear if
-  tools can handle this transparently.  In that case sentry-cli in particular is encouraged
-  to fail with an error and ask the user to explicitly configure the slug of the project
-  to use.
-* **Legacy tools:** for tools not using sentry-cli but using the API directly, there might
-  be a transitionary phase until the tool supports structural tokens.  In that case the
-  documentation would need to point out the correct way to configure this.  The same applies
-  to old installations of sentry-cli.
+Additionally **legacy tools** will require more configuration.  For tools not
+using sentry-cli but using the API directly, there might be a transitionary
+phase until the tool supports structural tokens.  In that case the documentation
+would need to point out the correct way to configure this.  The same applies to
+old installations of sentry-cli.
 
 # Order of Execution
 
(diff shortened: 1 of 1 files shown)


text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,273 @@
...
+interested in extracting data from the token.
+
+## Token Facts
+
+We want to encode certain information into the tokens.  In Biscuit terms these are called
+_facts_.  The following facts exist:
+
+* `site`: references the target API URL that should be used.  A token will always have a
+  site in it and clients are not supposed to provide a fallback.  For instance this
+  would be `https://myorg.sentry.io/`.
```
markstory, line 81: For hybrid cloud we won't be able to have API endpoints use `orgslug.sentry.io` without proxying through control silo. However, we could have structured tokens contain regional domains eg. `us.sentry.io`

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,273 @@
...
+* `org`: a token is uniquely bound to an org, so the slug of that org is also always
+  contained.  Note that the slug is used rather than an org ID as the clients typically
+  need these slugs to create API requests.
+* `projects`: a token can be valid for more than one project.  For operations such as
+  source map uploads it's benefitial to issue tokens bound to a single project in which
```
markstory, line 86:
```suggestion
  source map uploads it's beneficial to issue tokens bound to a single project in which
```

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,273 @@
...
+  'sentry_site': 'https://myorg.sentry.io/',
+  'sentry_org': 'myorg',
+  'sentry_projects': ['myproject']
+}
+```
+
+## Token Issuance
+
+The purpose of this change is to allow any organization member to issue tokens with little
+overhead. As users can already issue tokens which shocking levels of access to any of the
```
markstory, line 158:
```suggestion
overhead. As users can already issue tokens with shocking levels of access to any of the
```

mitsuhiko:
For future readers: it's unlikely we will go down the path of using Biscuits. They are too complex, so don't be put off by the part of that in the RFC.

scefali:
@mitsuhiko Might be a dumb question, but can we expose these tokens in the docs UI? It would be amazing if we could just automatically generate a token and put it in the docs link like we do for the user/org

Screen Shot 2023-05-16 at 11 26 56 AM

text/0091-ci-upload-tokens.md
dcramer: whys this different than org? why do we need this site component?

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,273 @@
...
+We want to encode certain information into the tokens.  In Biscuit terms these are called
+_facts_.  The following facts exist:
+
+* `site`: references the target API URL that should be used.  A token will always have a
+  site in it and clients are not supposed to provide a fallback.  For instance this
+  would be `https://myorg.sentry.io/`.
+* `org`: a token is uniquely bound to an org, so the slug of that org is also always
+  contained.  Note that the slug is used rather than an org ID as the clients typically
+  need these slugs to create API requests.
+* `projects`: a token can be valid for more than one project.  For operations such as
```
dcramer, line 85: we dont need project for the use case we have - so do we really need that as part of this spec?

dcramer:
A more broad question: why dont we just make an org level token that can be provisioned? why do we need an entire new token format/etc to solve such a simplistic problem?

```
OrgToken { token: string; scope: ["projects:symbols"] }
```

philipphofmann:
> A more broad question: why dont we just make an org level token that can be provisioned? why do we need an entire new token format/etc to solve such a simplistic problem?
> 
> ```
> OrgToken { token: string; scope: ["projects:symbols"] }
> ```

I had similar thoughts. If I'm not mistaken, the underlying problem is that users have a hard time understanding which token to create or use for uploading symbols. We could either solve this by what @dcramer suggested and/or improve the UI.

There are several ways how we could solve this. We could add a new some kind of symbols integration. If the user doesn't have the right permissions we currently show the warning `These settings can only be edited by users with the organization owner or manager role.` We could add a request token button, that notifies the people with the proper permissions to create such a token or an integration.

text/0091-ci-upload-tokens.md
mitsuhiko: For hybrid cloud the benefit is that you can (without a preflight) hit the right instance. Particularly for debug file uploads we do not want them to go via the central proxy at all times. For single tenant they all require different URLs anyways and today the docs are quite blind on what you have to reconfigure to point it to the right single tenant instance.

text/0091-ci-upload-tokens.md
mitsuhiko: Having the project as part of the token would make it possible that _just_ the token is all that is needed to upload debug files or source maps and automatically associate it with the right project without further configuration needed.  Otherwise it's yet again an extra piece of information that needs to be configured separately.

mitsuhiko:
> A more broad question: why dont we just make an org level token that can be provisioned? why do we need an entire new token format/etc to solve such a simplistic problem?

To upload correctly the following pieces of information are needed:

* target site: `sentry.io` vs `foo.my.sentry.io` (and whatever hybrid cloud needs)
* org slug
* project slug
* auth token

By encoding most of this information into the token it means that most configuration is no longer necessary.

mitsuhiko:
@scefali the goal is for sure that the docs can directly get such a token.

HazAT:
@scefali think of an experience similar to this
image

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,247 @@
...
+correctly handles Single Tenant:
+
+```javascript
+const SentryWebpackPlugin = require("@sentry/webpack-plugin");
+
+module.exports = {
+  devtool: "source-map",
+  plugins: [
+    new SentryWebpackPlugin({
+      authToken: "AUTO GENERATED TOKEN HERE",
```
evanpurkhiser, line 184: Should this still be something like `process.env.SNTRY_STRUCTURED_TOKEN` for this example? This seems to imply that it is not a secret

mitsuhiko: @evanpurkhiser the idea is that the docs directly inject the token same as they do for the slugs.

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,271 @@
...
+it possible to detect it by security scrapers.  Anyone handling such a token is
+required to check for the `sntrys_` prefix and disregard it before parsing it.  This
+can also be used by the client side to detect a structural token if the client is
+interested in extracting data from the token.
+
+## Token Facts
+
+We want to encode certain information into the tokens.  The following attributes are defined:
+
+* `sentry_site`: references the target API URL that should be used.  A token will always have a
```
mydea, line 76: Maybe specify here that this is `options['system.url-prefix']`, for reference?

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,271 @@
...
+  'iat': 1684154626,
+  'sentry_site': 'https://myorg.sentry.io/',
+  'sentry_org': 'myorg',
+  'sentry_projects': ['myproject']  # not used currently
+}
+```
+
+## Token Issuance
+
+The purpose of this change is to allow any organization member to issue tokens with little
```
mydea, line 139: I wonder if we should restrict _revocation_ to managers/owners? 🤔 this has the downside that a user cannot revoke a token they just created, but may be a reasonable safety net (even to disallow accidental deletion)?

text/0091-ci-upload-tokens.md
```diff
@@ -0,0 +1,271 @@
...
+
+For JWT the facts could be encoded as custom claims:
+
+```json
+{
+    "iss": "sentry.io",
+    "iat": 1684154626,
+    "sentry_site": "https://myorg.sentry.io/",
+    "sentry_org": "myorg"
+}
```
mdtro, line 98: We should add a `nonce` field here so these are less predictable/guessable. With this current format, someone would just need to brute force the `iat` to find a value token.

---

We should also store the tokens as hashed values in the database.

Then when it comes to verifying an inbound request, it's a matter of:
1. hash the provided token
2. check the database for a hash of equal value
3. if one is found, decode the provided token
5. perform requested operation (pending authorization)

mdtro: Are the last characters to help identify the token? If so, Could we instead expand the model to have a `name` or `description` or both? Then the user can provide a meaningful value to them.

mdtro: Yeah, that is okay. 👍

text/0091-ci-upload-tokens.md
mydea: referencing this comment from @markstory here: https://github.com/getsentry/sentry/pull/50409#discussion_r1224758745

> In the future, organizations will have a couple URLs:

> sentryUrl which is our root domain and is the same as system.url-prefix. ex. sentry.io
organizationUrl which is domain for the organization's UI. ex acme.sentry.io
regionUrl Which is the domain that the organization's API endpoints are available on. ex us.sentry.io
Any API requests sent to sentryUrl will be routed to the correct region, but will suffer a latency penalty (because we're proxying the request to the region) and the request will go to the US (this can matter for EU customers).

> For this scenario, I think you'll want both the sentryUrl and regionUrl available in the JWT. There are a small number of API endpoints that are only available on sentryUrl (users, integrations, sentry apps).

So based on this, I would say we should add `sentry_url` and `sentry_region_url` fields instead of `sentry_site`?

dcramer: why is anything other than the api url needed?

PR #92, opened 2023-05-15
rfc(decision): Replay Issue Creation

cmanallen:
This RFC proposes two possible paths for creating Replay issues.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/replay-issue-creation/text/0092-replay-issue-creation.md)

### DIFF
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -15,7 +15,7 @@ We want to detect certain categories of ssues that can greatly benefit from vide
 
 # Options Considered
 
-### Option 1: SDK
+### Option 1: SDK Generates Issues Through a Generic Interface
 
 When the SDK encounters a "replay issue" it will make an HTTP request to a generic interface which will handle the issue creation process.
 
@@ -32,7 +32,7 @@ When the SDK encounters a "replay issue" it will make an HTTP request to a gener
    - E.g. "Experienced 10 occurences in the past hour".
 3. The replay containing the issue could not be sampled.
 
-### Option 2: Ingest
+### Option 2: SDK Generates Issues Through a Session-Replay Specific Interface
 
 The SDK publishes a "replay issue" to the Replay back-end. The back-end will decide to process the issue or not.
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -28,7 +28,7 @@ When the SDK encounters a "replay issue" it will make an HTTP request to a gener
 1. Uses quota.
 2. Unclear if we're able to use dynamic thresholds.
    - E.g. "Experienced 10 occurences in the past hour".
-3. The replay containing the issue could be sampled.
+3. The replay containing the issue could not be sampled.
 
 ### Option 2: Ingest
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -5,7 +5,9 @@
 
 # Summary
 
-We want to detect certain categories of issues only available through the Session Replay product. The first of these issues is the "Slow Click" detection issue.
+We want to detect certain categories of issues only available through the Session Replay product. These issues can only be detected on the SDK. The Replay back-end will never have enough data to find these issues. For that reason this is primarily an SDK driven workload with the Replay back-end acting as a middleman between the SDK and the Issues platform.
+
+Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface?
 
 # Motivation
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -21,7 +21,9 @@ When the SDK encounters a "replay issue" it will make an HTTP request to a gener
 
 **Pros:**
 
-1. Can be tested in isolation without impacting existing production services.
+1. We can be tested in isolation without impacting existing production services.
+2. We can expose these new features in a more product agnostic way.
+   - This would allow us to reach much larger customer segments when compared to Replay customers.
 
 **Cons:**
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -5,7 +5,9 @@
 
 # Summary
 
-We want to detect certain categories of ssues that can greatly benefit from videos and context in the Session Replay product. These issues can only be detected on the SDK. The Replay back-end will never have enough data to find these issues. For that reason this is primarily an SDK driven workload. The question is: what role should the Replay back-end have in Replay issue creation? Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface? Each option would have significantly different product implications that will be discussed below.
+The Session Replay team wants to detect certain categories of issues that can greatly benefit from videos and context in the Session Replay product.
+
+These issues can only be detected on the SDK. The Replay back-end will never have enough data to find these issues. For that reason this is primarily an SDK driven workload. What role should the Replay back-end have in Replay issue creation? Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface? Each option would have significantly different product implications that will be discussed below.
 
 # Motivation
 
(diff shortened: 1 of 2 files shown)


text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+
+We want to detect certain categories of issues only available through the Session Replay product. The first of these issues is the "Slow Click" detection issue.
+
+# Motivation
+
+1. Inform the long-term strategy for detecting and ingesting Session Replay generated issues.
+2. Provide actionable feedback to developers that could not otherwise be captured in a performance span or error event.
+3. Increase product awareness in free-tier customers and encourage adoption.
+
+# Options Considered
```
ryan953, line 16: I'm trying to think of a way to make sure we're looking at an exhaustive list of options, even if some are bad. 

here's where i'm starting from:

| SDK | Network data | Relay & Ingest | Result |
| --- | --- | --- | --- |
| @sentry/javascript | error payload (call `captureException()` in browser)| normal processing | error quota risk |
| @sentry/javascript | click detection (same payload as in experiment) | server-side detector needed | - | 
| @sentry/replay | existing breadcrumbs + rrweb events | normal relay processing | server-side detector needed | replay must be sampled |
| @sentry/replay | click detection breadcrumb (same payload as in experiment) | server-side detector needed | replay must be sampled |

jas-kas: @ryan953 @cmanallen Could we add a column here that is "Product Experience" or "Product Implications" so we can visualize what this means to our users?

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
+- Start Date: 2023-05-15
+- RFC Type: decision
+- RFC PR: https://github.com/getsentry/rfcs/pull/92
+- RFC Status: draft
+
+# Summary
+
+We want to detect certain categories of issues only available through the Session Replay product. These issues can only be detected on the SDK. The Replay back-end will never have enough data to find these issues. For that reason this is primarily an SDK driven workload. The question is: what role should the Replay back-end have in Replay issue creation? Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface?
```
jas-kas, line 8: > We want to detect certain categories of issues only available through the Session Replay product.

Since we _can_ detected these categories of issues without the help of the Replay SDK all together, we should probably reword the "issues only available through the Session Replay product." language. How about, "issues that can greatly benefit from videos and context in the Session Replay product." or similar?

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Motivation
+
+1. Inform the long-term strategy for detecting and ingesting Session Replay generated issues.
+2. Provide actionable feedback to developers that could not otherwise be captured in a performance span or error event.
+3. Increase product awareness in free-tier customers and encourage adoption.
```
jas-kas, line 14:
```suggestion
3. Increase product awareness in platform customers and encourage adoption.
```

text/0092-replay-issue-creation.md
jas-kas, line 8:
```suggestion
We want to detect certain categories of issues only available through the Session Replay product. These issues can only be detected on the SDK. The Replay back-end will never have enough data to find these issues. For that reason this is primarily an SDK driven workload. The question is: what role should the Replay back-end have in Replay issue creation? Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface? Each option would have significantly different product implications that will be discussed below.
```

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+
+# Options Considered
+
+### Option 1: SDK
```
jas-kas, line 18: Can we have the option names match the wording you used in the Summary section? Or vice versa. It would make it a little easier to follow~

"Should the Replay SDK use the Replay back-end to generate new issues or should the SDK generate those issues through a generic, non-replay-specific interface?"

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+
+When the SDK encounters a "replay issue" it will make an HTTP request to a generic interface which will handle the issue creation process.
+
+**Pros:**
+
+1. Can be tested in isolation without impacting existing production services.
+
+**Cons:**
+
+1. Uses quota.
```
jas-kas, line 28: I assume you mean errors quota here. It would be good to state that specifically for the revised RFC content.

As an aside, is it possible to do Option 1 and exclude it from consuming errors quota?

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+2. Unclear if we're able to use dynamic thresholds.
+   - E.g. "Experienced 10 occurences in the past hour".
+3. The replay containing the issue could be sampled.
```
jas-kas, line 31:
```suggestion
3. The replay containing the issue could not be sampled.
```

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+
+### Option 2: Ingest
+
+The SDK publishes a "replay issue" to the Replay back-end. The back-end will decide to process the issue or not.
+
+**Pros:**
+
+1. Does not use quota.
```
jas-kas, line 39:
```suggestion
1. Does not use errors quota.
```

text/0092-replay-issue-creation.md
jas-kas: ^ If the answer is 'yes' to the above question, we'll have to modify the pros/cons list

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,55 @@
...
+2. Provide actionable feedback to developers that could not otherwise be captured in a performance span or error event.
+3. Increase product awareness in free-tier customers and encourage adoption.
+
+# Options Considered
+
+### Option 1: SDK
+
+When the SDK encounters a "replay issue" it will make an HTTP request to a generic interface which will handle the issue creation process.
+
+**Pros:**
```
jas-kas, line 22: For each option, I would split out the Pros/Cons from engineering POV vs. product POV. (I can fill in Product ✨ ) 

A clear "Pro" to this solution is we'd be exposing Dead Click Issues to much much larger customer segment when compared to only Replay customers, for example. 😃

text/0092-replay-issue-creation.md
wedamija, line 8: When we first worked on performance issues the initial spike was doing them via sdk. But there was a lot of pushback to move things to the backend, since implementing this logic across N sdks isn't really a scalable solution.

Maybe this isn't such a big deal if this only applies to the JS world though? Either way it might be worth checking that upper management isn't going to veto this approach later on.

text/0092-replay-issue-creation.md
```diff
@@ -0,0 +1,58 @@
...
+2. Overhead is low.
+   - Publishing to a Kafka consumer can happen asynchronously.
+3. We can use dynamic thresholds.
+   - E.g. "Experienced 10 occurences in the past hour".
+4. Replay is guaranteed to be sampled.
+
+**Cons:**
+
+1. Poor rollout could impact service availability during testing period.
+2. Requires coordination between the SDK and Ingest to create new issue types.
```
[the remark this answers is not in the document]
cmanallen: Profiling does it in the back end.  Performance I'm not sure exactly.  I know they have a lot of back end detectors so at least some portion is handled there.

text/0092-replay-issue-creation.md
cmanallen: Added.  I left it as a single pros/cons table for now.  If we start to accumulate a lot of items we can split it.

text/0092-replay-issue-creation.md
cmanallen: > Maybe this isn't such a big deal if this only applies to the JS world though? 

Yeah I'm hoping thats the case.

Unfortunately this work has to be done on the SDK due to the large sizes of replays and the large time differences between segments (segments can refer back to any previous segment).  This is probably something that should be documented in the RFC :P

bruno-garcia:
The team decided to reopen this RFC in order to expand on the options further. Some of the goals:

> Should an HTTP interface for creating issue events be created this RFC can be re-addressed.

* Give more details on what this interface can look like.
* Talk about a simple SDK-only approach with `captureEvent`

PR #93, opened 2023-05-17
Request and Response bodies for GraphQL errors

marandaneto:

[Rendered RFC](https://github.com/getsentry/rfcs/blob/choregraphql-errors/text/0087-graphql-errors.md)

### DIFF
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -6,7 +6,7 @@
 
 # Summary
 
-Add `Request` and `Response` dies to GraphQL errors.
+Add `Request` and `Response` bodies to GraphQL errors.
 
 This feature is opt-in by default due to PII concerns.
 
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -1,6 +1,6 @@
 * Start Date: 2023-05-17
 * RFC Type: feature
-* RFC PR: [#](https://github.com/getsentry/rfcs/pull/)
+* RFC PR: [#93](https://github.com/getsentry/rfcs/pull/93)
 * RFC Status: draft
 * RFC Driver: [Manoel Aranda Neto](https://github.com/marandaneto)
 
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -24,7 +24,7 @@ This is necessary to implement the [GraphQL Client Errors](https://www.notion.so
 
 The `Request` and `Response` bodies could contain PII.
 
-Request body example:
+Request body example (query language only):
 
 ```
 query { 
@@ -46,7 +46,7 @@ Response body example:
 }
 ```
 
-A Request example with an error:
+A Request example with an error (query language only):
 
 ```
 query { 
@@ -85,13 +85,42 @@ Response body example:
 }
 ```
 
-Using the `locations` and the request body, we can highlight the error in the request body.
+Using the `locations` and the request body (query language only), we can highlight the error in the request body.
 
 ```
 query { 
   viewer { 
     # note the removed `n` at the end
     -> logi
+  }
+}
+```
+
+Request body example, full body (not only the query language):
+
+```json
+{
+  "query": "{\n  viewer {\n    login\n  }\n}",
+  "variables": {}
+}
+```
+
+Request body example, full body (not only the query language):
+
+```json
+{
+  "query": "{\n  viewer {\n    login\n  }\n}",
+  "variables": {}
+}
+```
+
+The Request body can also contain `variables`.
+
+```json
+{
+  "query": "{\n  viewer {\n    login\n  }\n}",
+  "variables": {
+    "login": "marandaneto"
   }
 }
 ```
@@ -143,10 +172,11 @@ Since GraphQL is a well defined spec, we can also scrub the GraphQL fields.
 
 Request example:
 
-```
-query { 
-  viewer { 
-    login
+```json
+{
+  "query": "{\n  viewer {\n    login\n  }\n}",
+  "variables": {
+    "login": "marandaneto"
   }
 }
 ```
@@ -162,6 +192,8 @@ Response example:
   }
 }
 ```
+
+In this case, we only need to use the Request `variables` and its keys to scrub the Response `data`.
 
 # Drawbacks
 
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -107,15 +107,6 @@ Request body example, full body (not only the query language):
 }
 ```
 
-Request body example, full body (not only the query language):
-
-```json
-{
-  "query": "{\n  viewer {\n    login\n  }\n}",
-  "variables": {}
-}
-```
-
 The Request body can also contain `variables`.
 
 ```json
@@ -154,7 +145,9 @@ By doing this, we can keep the `Request` interface as it is, we can copy the `da
       "type": "response",
       "status_code": 500,
       "body_size": 1000,
-      "data": "foo",
+      "data": {
+        "foo": "bar"
+      }
     }
   }
 }
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -28,7 +28,7 @@ The `Request` and `Response` bodies could contain PII.
 
 Request body example (query language only):
 
-```
+```graphql
 query { 
   viewer { 
     login
@@ -50,7 +50,7 @@ Response body example:
 
 A Request example with an error (query language only):
 
-```
+```graphql
 query { 
   viewer { 
     # note the removed `n` at the end
@@ -89,7 +89,7 @@ Response body example:
 
 Using the `locations` and the request body (query language only), we can highlight the error in the request body.
 
-```
+```graphql
 query { 
   viewer { 
     # note the removed `n` at the end
--- a/text/0087-graphql-errors.md
+++ b/text/0087-graphql-errors.md
@@ -1,0 +1,226 @@
+* Start Date: 2023-05-17
+* RFC Type: feature
+* RFC PR: [#](https://github.com/getsentry/rfcs/pull/)
+* RFC Status: draft
+* RFC Driver: [Manoel Aranda Neto](https://github.com/marandaneto)
+
+# Summary
+
+Add `Request` and `Response` body to events.
+
+# Motivation
+
+The [Request](https://develop.sentry.dev/sdk/event-payloads/types/#request) interface contains information on a HTTP request related to the event.
+
+The [Request] contains the `data` field which is the request body. Can be given as string or structural data of any format.
+
+The [Response](https://develop.sentry.dev/sdk/event-payloads/types/#responsecontext) interface contains information on a HTTP response related to the event.
+
+However, the `Response` interface does not contain the `data` field which is the response body.
+
+# Background
+
+This is necessary to implement the [GraphQL Client Errors](https://www.notion.so/sentry/GraphQL-proposal-d6e5846f30434770903cf3af20bc2568) with syntax highlight.
+
+The Request and Response body could contain PII.
+
+Request body example:
+
+```
+query { 
+  viewer { 
+    login
+  }
+}
+```
+
+Response body example:
+
+```json
+{
+  "data": {
+    "viewer": {
+      "login": "marandaneto"
+    }
+  }
+}
+```
+
+Now a Request example with an error:
+
+```
+query { 
+  viewer { 
+    # note the removed `n` at the end
+    logi
+  }
+}
+```
+
+Response body example:
+
+```json
+{
+  "errors": [
+    {
+      "path": [
+        "query",
+        "viewer",
+        "logi"
+      ],
+      "extensions": {
+        "code": "undefinedField",
+        "typeName": "User",
+        "fieldName": "logi"
+      },
+      "locations": [
+        {
+          "line": 4,
+          "column": 5
+        }
+      ],
+      "message": "Field 'logi' doesn't exist on type 'User'"
+    }
+  ]
+}
+```
+
+Using the `locations` and the request body, we can highlight the error in the request body.
+
+```
+query { 
+  viewer { 
+    # note the removed `n` at the end
+    -> logi
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


text/0087-graphql-errors.md
```diff
@@ -0,0 +1,228 @@
...
+
+# Background
+
+This is necessary to implement the [GraphQL Client Errors](https://www.notion.so/sentry/GraphQL-proposal-d6e5846f30434770903cf3af20bc2568) with syntax highlight.
+
+The `Request` and `Response` bodies could contain PII.
+
+Request body example:
+
+```
```
marandaneto, line 29: Is there a ```graphql for markdown?

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,228 @@
...
+
+Attachments have to be special cased in Sentry, seems hacky, we do that with screenshots already.
+
+# Unresolved questions
+
+- Do we need to send the GraphQL scheme to Sentry in order to do data scrubbing properly?
+- Should this feature be opt-in due to the size of the data and PII risk?
+- Should we send the Request and Response as different envelope items? (avoid the size limit)
+- Should PII be scrubbed in the SDK instead?
```
marandaneto, line 228: We'll need to duplicate quite some work in all SDKs.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,228 @@
...
+bar
+```
+
+Attachments have to be special cased in Sentry, seems hacky, we do that with screenshots already.
+
+# Unresolved questions
+
+- Do we need to send the GraphQL scheme to Sentry in order to do data scrubbing properly?
+- Should this feature be opt-in due to the size of the data and PII risk?
+- Should we send the Request and Response as different envelope items? (avoid the size limit)
```
marandaneto, line 227: See option 3 and 4.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,228 @@
...
+foo
+{"type":"attachment","length":10,"content_type":"application/json","filename":"response.txt"}\n
+bar
+```
+
+Attachments have to be special cased in Sentry, seems hacky, we do that with screenshots already.
+
+# Unresolved questions
+
+- Do we need to send the GraphQL scheme to Sentry in order to do data scrubbing properly?
```
marandaneto, line 225: It'd require changes on sentry-cli, data scrubbing, etc.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,228 @@
...
+}
+```
+
+# Drawbacks
+
+[Envelopes](https://develop.sentry.dev/sdk/envelopes) (Events) contain way lower [size limits](https://develop.sentry.dev/sdk/envelopes/#size-limits). The `data` fields could be large and it could be a problem.
+
+SDKs should discard large and binary bodies by default, using the [maxRequestBodySize](https://docs.sentry.io/platforms/android/configuration/options/#max-request-body-size) and `maxResponseBodySize` (it'll be added) options.
+
+The difference is that for GraphQL errors, this should be enabled by default.
```
marandaneto, line 172: If we don't send the payload, the user won't have context about what has to be fixed.

marandaneto:
@scefali @malwilley adding you to the loop since this change affects the FE for syntax highlight.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,260 @@
...
+}
+```
+
+* `data`: Can be given as string or structural data of any format..
+
+The `Response` interface keeps arbitrary fields, it's back compatible with the current implementation.
+
+## Must have
+
+The fields `Request#data` and `Response#data` could contain PII and they should run data scrubbing agressively.
```
cleptric, line 167: We should consider allowing people to opt out of scrubbing. Often, the actual payload is the information needed to reproduce bugs.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,260 @@
...
+{"type":"attachment","length":10,"content_type":"application/json","filename":"response.txt"}\n
+bar
+```
+
+Attachments have to be special cased in Sentry, seems hacky, we do that with screenshots already.
+
+# Unresolved questions
+
+- Do we need to send the GraphQL scheme to Sentry in order to do data scrubbing properly?
+- Should this feature be opt-in due to the size of the data and PII risk?
```
cleptric, line 258: I would strongly advise making this an opt-in feature only.

text/0087-graphql-errors.md
marandaneto: Headers and URLs will follow this rule already, this is already implemented on some if not all.
Will add a note.

text/0087-graphql-errors.md
marandaneto: Added a note.

text/0087-graphql-errors.md
cleptric: Which project config? Are you talking about server-side scrubbing?

text/0087-graphql-errors.md
cleptric: Wrong RFC...
I meant https://github.com/getsentry/rfcs/blob/main/text/0062-controlling-pii-and-credentials-in-sd-ks.md

marandaneto:
The event will likely need an identification either via `mechanism.type` or something else to trigger Relay scrubbing specific to GraphQL spec.
WDYT @getsentry/ingest ?

text/0087-graphql-errors.md
marandaneto: Yes, eg https://sentry-sdks.sentry.io/settings/projects/sentry-cocoa/security-and-privacy/
Scrubbing will be server-side, so if there's an option to enable/disable, it should be there.
People can also opt-in for SDK scrubbing using `beforeSend`.

text/0087-graphql-errors.md
marandaneto: The other option is to manipulate the request and payload sent, for example, sending only the `errors` of the response and the body without the `variables` from the request, which requires serializing/deserializing the request and response of the client, this might add performance overhead as just sending it as it is, but it's an option.

krystofwoldrich: How will we handle situations where partial data are returned with errors? Will we only send the errors to Sentry or the filtered data too?

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,262 @@
...
+    }
+  }
+}
+```
+
+In this case, we only need to use the Request `variables` and its keys to scrub the Response `data`.
+
+# Drawbacks
+
+[Envelopes](https://develop.sentry.dev/sdk/envelopes) (Events) contain way lower [size limits](https://develop.sentry.dev/sdk/envelopes/#size-limits). The `data` fields could be large and it could be a problem.
```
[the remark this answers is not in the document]
marandaneto: Yes.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,262 @@
...
+```
+
+Request body example, full body (not only the query language):
+
+```json
+{
+  "query": "{\n  viewer {\n    login\n  }\n}",
+  "variables": {}
+}
+```
```
philipphofmann, line 117: `l`: This section is duplicated if I'm not mistaken.
```suggestion
```

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,262 @@
...
+By doing this, we can keep the `Request` interface as it is, we can copy the `data` scrubbing rules from the `Request` interface.
+
+```json
+{
+  "contexts": {
+    "response": {
+      "type": "response",
+      "status_code": 500,
+      "body_size": 1000,
+      "data": "foo",
```
philipphofmann, line 157: `l`: I guess most of the time it will be something similar to the following. Would make the RFC easier to understand, IMO.
```suggestion
    "data": {
      "foo": "bar"
    },
```

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,262 @@
...
+
+The `Response` interface keeps arbitrary fields, it is backwards compatible with the current implementation.
+
+## Must have
+
+The fields `Request#data` and `Response#data` could contain PII and they should run data scrubbing agressively.
+
+[Session Replay](https://docs.sentry.io/platforms/javascript/guides/remix/session-replay/configuration/) already sends the request and response bodies, so we can use the same data scrubbing rules.
+
+Since GraphQL is a well defined spec, we can also scrub the GraphQL fields.
```
philipphofmann, line 173: What about only sending the fields we are sure about not containing any PII? Then we could turn this feature on by default and no scrubbing would be needed.

philipphofmann (commented):
I'm a bit confused by this RFC 🤔. The title is `Request and Response bodies for GraphQL errors`, the summary describes

> Add `Request` and `Response` body to events.

and the option 1 is about 

> adding a `data` field in the [Response]

I'm totally on board with option 1 to solve the problem for GraphQL errors. What I'm unsure about is if we also want to decide in this RFC 

1. if we should add request and response bodies to events (that's unclear to me)
2. how to send GraphQL request and response data (I would assume not)

Could you please help me 😄, @marandaneto?

text/0087-graphql-errors.md
marandaneto: It's still an error, https://www.notion.so/sentry/GraphQL-proposal-d6e5846f30434770903cf3af20bc2568?pvs=4#65c51cb28c1c421e8b3edd4f5fe2f427
I'll check during the implementation if there's a performance impact of serializing the error, extracting the non-required fields, etc.

text/0087-graphql-errors.md
marandaneto: Good catch, I guess it was a merging or copypasta error.

text/0087-graphql-errors.md
marandaneto: The downside is the performance impact of serializing on the client and duplicating all the data scrubbing logic in every SDK, hence focusing on Relay for PII.
It's also easier to roll out bug fixing on Relay.
It is a valid option though.

marandaneto: I'll add it as a note.

text/0087-graphql-errors.md
```diff
@@ -0,0 +1,257 @@
...
+
+[GraphQL Support](https://github.com/getsentry/sentry/discussions/38913)
+
+[Support for GraphQL errors](https://github.com/getsentry/sentry/issues/33723)
+
+[Send request body for http client integrations and similar](https://github.com/getsentry/team-mobile/issues/41)
+
+# Proposal (Option 1)
+
+The proposal is adding a `data` field in the [Response](https://develop.sentry.dev/sdk/event-payloads/types/#responsecontext) interface.
```
jjbayer, line 137: Dart seems to use the key `body` instead of `data`: https://github.com/getsentry/sentry-dart/blob/7011abe27ac69bd160bdc6ecf3314974b8340b97/dart/lib/src/protocol/sentry_response.dart#L59

I think `data` is better because it is consistent with `Request`, but I wanted to point out the inconsistency.

jjbayer:
@marandaneto 

> The event will likely need an identification either via mechanism.type or something else to trigger Relay scrubbing specific to GraphQL spec.

In the RFC, you wrote

> By doing this, we can keep the `Request` interface as it is, we can copy the `data` scrubbing rules from the `Request` interface.

I am confused now, do we want to apply the same data scrubbing rules as for `request.data`, or do we need to introduce new scrubbing rules? In the latter case, we would need to define a strict protocol for graphql payloads.

text/0087-graphql-errors.md
philipphofmann: Doing it on every SDK is a downside, yes, but the logic for only selecting well-known fields shouldn't be that complicated. We could also move it to Relay if it is too complicated.

text/0087-graphql-errors.md
marandaneto: Good point, I can deprecate `body` and add `data`, both will serialize as `data` anyway.

marandaneto:
> @marandaneto
> 
> > The event will likely need an identification either via mechanism.type or something else to trigger Relay scrubbing specific to GraphQL spec.
> 
> In the RFC, you wrote
> 
> > By doing this, we can keep the `Request` interface as it is, we can copy the `data` scrubbing rules from the `Request` interface.
> 
> I am confused now, do we want to apply the same data scrubbing rules as for `request.data`, or do we need to introduce new scrubbing rules? In the latter case, we would need to define a strict protocol for graphql payloads.

The goal is that if the event isn't a GraphQL error, it should run data scrubbing for `response.data` anyway similarly to `request.data`.
If the Event is a GraphQL error, then yes, we'll need to run data scrubbing based on the graphql protocol, so technically both things.
See the question https://github.com/getsentry/rfcs/pull/93#issuecomment-1551282085
We need a marker to make the difference between the normal scrubbing and the graphql protocol scrubbing.

marandaneto:
@jjbayer and @jan-auer is there something you'd like to see changed in this RFC or are we good to go?

jjbayer:
> If the Event is a GraphQL error, then yes, we'll need to run data scrubbing based on the graphql protocol

@marandaneto so you want to parse the graphql and iterate over its keys / values, and apply existing data scrubbing rules to those parsed items? If so, we definitely need a discriminator. I would put it inside the request / response, maybe an optional `type` field in both interfaces?

marandaneto:
> > If the Event is a GraphQL error, then yes, we'll need to run data scrubbing based on the graphql protocol
> 
> @marandaneto so you want to parse the graphql and iterate over its keys / values, and apply existing data scrubbing rules to those parsed items? If so, we definitely need a discriminator. I would put it inside the request / response, maybe an optional `type` field in both interfaces?

yes, that can be the mechanism as suggested here https://github.com/getsentry/rfcs/pull/93#issuecomment-1551282085
Or it could be somewhere else as well, no strong feelings about that, where it'd make more sense, not sure if it makes sense to be part of the request/response tho, though probably something under the event makes more sense.
The FE bits will need that too for the special casing of the issue view and syntax highlight.

jjbayer (approved):

philipphofmann (approved):
LGTM

PR #95, opened 2023-05-23
rfc(decision): Escalating Forecasts Merged Issues

NisanthanNanthakumar:
TODO. [Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/escalatingforecastsmergedissues/text/0095-escalating-forecasts-merged-issues.md)

### DIFF
--- a/text/0095-escalating-forecasts-merged-issues.md
+++ b/text/0095-escalating-forecasts-merged-issues.md
@@ -5,31 +5,64 @@
 
 # Summary
 
-One paragraph explanation of the feature or document purpose.
+We need to decide how to handle the Issue States and Escalating Forecasts for Merged issues.
+
+## What is Merged Issues
+
+If you have similar-looking Error Issues that have not been grouped together automatically and you want to reduce noise, you can do so by merging them. A user can also un-merge the issues. 
+
+Sentry does not infer any new grouping rules from how you merge issues. Future events will be added to the merged set of issues by the same criteria as they would've been added to the individual issues that are now in the merged set.
+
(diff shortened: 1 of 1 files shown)


text/0095-escalating-forecasts-merged-issues.md
```diff
@@ -0,0 +1,68 @@
...
+### Un-Merge
+
+On un-merge, the GroupStatus of the separated Issues will be of the merged Issue. We should start a new timer for the newly split Issues, but maintain the timer for the primary Issue (the Issue that accumulates all the other Issues).
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+We need to determine how often events are getting merged and un-merged by users. A high volume could lead to high load to Snuba when we generate new Escalating Forecasts for the newly merged/un-merged Issues.
```
wedamija, line 66: I think we should have stats on this somewhere in datadog already. I don't think it'll be a huge issue to recalculate on merge, because even if we merge 1000 issues into a single issue, it's still just generating a single forecast.

I suppose that on unmerge is the larger risk, since we might trigger a large number of new forecasts from a single issue. Maybe the way to keep things safe here is to just keep the number of workers on the forecast queue low enough that it won't hurt snuba too much.

One other potential issue is that I think it takes a while for a merge to go through to clickhouse and to rewrite the group ids. So we'd need to delay the task, but I don't know that we have  any specific confirmation on when the merge has completed (maybe sns can advise here)

wedamija (approved):
This seems well thought out to me 👍

PR #97, opened 2023-05-24
RFC0071 Shorter ENV var names

antonpirker:
Just a minor update to have shorter env var names. Rest stays the same. As this has not implemented in any SDK, this should be save to merge.


antonpirker:
Ok, will not change the env var names, because a unified prefix makes more sense than having short env var names.

PR #98, opened 2023-05-24
rfc(decision): Store Multiple Replay Segments in a Single Blob

cmanallen:


text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,91 @@
...
+
+Google Cloud Storage lists the costs for writing and storing data as two separate categories. Writing a file costs $0.005 per 1000 files. Storing that file costs $0.02 per gigabyte. For the average file this works out to: $0.000000012 for storage and $0.00000005 for the write.
+
+In practical terms, this means 75% of our spend is allocated to writing new files.
+
+# Proposal
+
+First, a new table called "recording_byte_range" with the following structure is created:
+
+| replay_id | segment_id | path     | start | stop  |
```
JoshFerge, line 30: i wonder if we could store this on the corresponding replay event for the segment?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,91 @@
...
+    B --> C[Push to Buffer];
+    C --> D{Buffer Full?};
+    D -- No --> A;
+    D -- Yes --> E[Write Single File to GCS];
+    E --> F[Bulk Insert Byte Ranges];
+    F --> G[Clear Buffer];
+    G --> A;
+```
+
+Third, when a client requests recording data we will look it up in the "recording_byte_range" table. From it's response, we will issue as many fetch requests as there are rows in the response. These requests may target a single file or many files. The files will be fetched with a special header that instructs the service provider to only respond with a subset of the bytes. Specifically, the bytes that related to our replay.
```
JoshFerge, line 56: could you link to information about the GCS API and how it uses byte ranges? i'm wondering about the mechanics of that.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,91 @@
...
+| B         | 0          | file.txt | 6242  | 8214  |
+| A         | 1          | file.txt | 8214  | 12457 |
+
+Replay and segment ID should be self-explanatory.  Path is the location of the blob in our bucket.  Start and stop are integers which represent the index positions in an exclusive range. This range is a contiguous sequence of related bytes. In other words, the segment's compressed data is contained within the range.
+
+Notice each row in the example above points to the same file but with different start and stop locations. This is implies that multiple segments and replays can be present in  the same file. A single file can be shared by hundreds of different segments.
+
+This table will need to support, at a minimum, one write per segment.
+
+Second, the Session Replay recording consumer will not _commit_ blob data to Google Cloud Storage for each segment. Instead it will buffer many segments and flush them together as a single blob to GCS. Next it will make a bulk insertion into the database for tracking.
```
JoshFerge, line 42: need to think about: possibilities for incidents & data loss in case of a consumer crash or increased load.

would it be better to use redis for the buffer?

what about an SSD based solution so we have some persistence potential? 

I'm also wondering if the flushing to GCS should be in a separate consumer / process altogether

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,91 @@
...
+    G --> A;
+```
+
+Third, when a client requests recording data we will look it up in the "recording_byte_range" table. From it's response, we will issue as many fetch requests as there are rows in the response. These requests may target a single file or many files. The files will be fetched with a special header that instructs the service provider to only respond with a subset of the bytes. Specifically, the bytes that related to our replay.
+
+The response bytes will be decompressed, merged into a single payload, and returned to the user as they are now.
+
+# Drawbacks
+
+1. Deleting data becomes tricky. See "Unresolved Questions".
```
JoshFerge, line 62: i'm also a bit concerned about the security implications -- if byte math gets messed up we could potentially return data from the wrong project

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: Updated the RFC with links and examples.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: Added a section on security.  If you still have concerns or feel like a point needs to be addressed better please re-raise the question.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: We could.  We would need to merge the replay-event and recording-event payloads first.  But this means that we wouldn't be able to store segment parts separately according to the `Extensions` section.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+     - We would remove all capability to access it making it functionally deleted.
+   - GDPR deletes:
+     - Would this require downloading the file, over-writing the subsequence of bytes, and re-uploading a new file?
+       - Single replays, small projects, or if the mechanism is infrequently used could make this a valid deletion mechanism.
+       - The data could be encrypted, with some encryption key stored on the metadata row, making the byte sequence unreadable upon row delete.
+
+2. What datastore should we use to store the byte range information?
+
+   - Cassandra, Postgres, AlloyDB?
+   - Postgres likely won't be able to keep up long-term.
```
volokluev, line 78: Do you have any numbers to back up that claim? postgres can do a lot, espcially if yall get a dedicated postgres instance for this

text/0098-store-multiple-replay-segments-in-a-single-blob.md
volokluev: Am I correctly understanding that one blob would store the segments for an arbitrary amount of different replays?

cmanallen: I wouldn't say arbitrary but any number of segments from different replays and different projects.

volokluev: And the current procedure is storing one blob per segment?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
volokluev: oh yes we should use AlloyDB for this (it is just postgres, but managed meaning the likelihood of getting paged for it is much less).

cmanallen: Perfect 👍

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+   - The replay_id functions as an idempotency token in the outcomes consumer and prevents the customer from being charged for the same replay multiple times.
+4. Click tracking.
+   - Duplicate click events will be inserted for a replay, segment pair.
+   - This is an acceptable outcome and will not impact search behavior.
+
+## Security
+
+How do we prevent users from seeing segments that do not belong to them? The short-answer is test coverage. Recording another segment's byte range would be the same as generating the incorrect filename under the current system. These outcomes are prevented with robust test coverage.
+
+In the case of bad byte range math, we do have some implicit protection. Each segment is compressed independently. Fetching a malformed byte range would yield unreadable data. A bad byte range either truncates the compression headers or includes unintelligible bytes at the beginning or end of the sequence. If we manage to decompress a subset of a valid byte range the decompressed output would be malformed JSON and would not be returnable to the user.
```
bmckerry, line 206: What if the byte range accidentally includes a full other segment? (unlikely but worth considering imo)

bmckerry (commented):
Left a few comments about data deletions, but I don't feel good about coupling separate replays together like this.
How do you feel instead about buffering all segments from a single replay together and storing those as individual blobs? There would be less upside cost-wise, but (imo) it would simplify how we store our data, decrease storage costs somewhat, and make it so we only fetch one blob per replay. 
I'm not sure how much work we'll need to accommodate this, just something I wanted to float

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+# Drawbacks
+
+1. Deleting data becomes tricky. See "Unresolved Questions".
+
+# Unresolved Questions
+
+1. Can we keep deleted data in GCS but make it inaccessible?
+
+   - User and project deletes:
+     - We would remove all capability to access it making it functionally deleted.
```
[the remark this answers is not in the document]
cmanallen: First you would download the blob, then you would remove the bytes between A and B, then upload a new file, and finally update the segment rows to point to the new file and byte offsets.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+
+The response bytes will be decompressed, merged into a single payload, and returned to the user as they are now.
+
+# Drawbacks
+
+1. Deleting data becomes tricky. See "Unresolved Questions".
+
+# Unresolved Questions
+
+1. Can we keep deleted data in GCS but make it inaccessible?
```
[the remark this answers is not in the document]
cmanallen: What if its encrypted and the encryption key has been deleted?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Handling Consumer Restarts**
+
+If the consumer restarts with a non-empty buffer, the buffer's last item's offset will not be committed. When the consumer resumes it will start processing from the last offset committed (i.e. the last item in the last successfully-flushed-buffer). The buffer will be rebuilt exactly as it was prior to restart.
+
+**Storage Service Failure**
+
+If we can not communicate with the storage provider we have several options.
+
+1. Catch the exception and commit the offset anyway. This means all the segments in the buffer would be lost.
```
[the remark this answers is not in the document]
cmanallen: Agreed.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,208 @@
...
+   - This operation typically takes >1ms to complete.
+   - This operation only occurs for segment-0 events.
+   - Querying this information in a tight loop is not an ideal situation.
+     - Forwarding the project_id to a secondary Kafka consumer would free up resources on our main consumer and allow the secondary consumer to optimize for this type of workload.
+     - Alternatively, another method for looking up the project's `has_replay` flag could be found.
+
+**Duplicate Message Handling**
+
+1. Google Cloud Storage.
+   - Unique filename generation per buffer would mean that a segment could be present in multiple files.
```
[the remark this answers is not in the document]
cmanallen: We would write a record for every segment (including its duplicates) and delete a segment from multiple files if necessary.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: Then it would download that segment.  But that would be akin to a segment pointing to the wrong file name.  So its not a new failure mode.  With good test coverage neither of those cases should be possible.

cmanallen:
@bmckerry 

> How do you feel instead about buffering all segments from a single replay together and storing those as individual blobs? 

Not possible unfortunately.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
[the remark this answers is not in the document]
bmckerry: Circling back on this now, I checked with security and I was very wrong - it sounds like encrypting data we want to delete is how they would prefer data deletions to happen (there's a DACI on this exact topic [here](https://www.notion.so/sentry/DACI-Should-we-use-encryption-to-protect-customer-data-8f8f101ec64a4d4289b7cddebbbf53c0)).

mdtro:
I can help address some of the security concerns around data deletion and segment leaking. Symmetric encryption can solve both of these problems for us.

---

Specifically, I recommend using [envelope encryption](https://cloud.google.com/kms/docs/envelope-encryption) for this. With envelope encryption, we'll have two layers of keys. The following is rather broad and general guidelines, but we can work to make this more specific as needed. 🙂 

- the Data Encryption Key (DEK): *used to encrypt the actual data*
- the Key Encryption Key (KEK): *used to encrypt the Data Encryption Key* 

With the proposed design, we'd need a data encryption key (DEK) per segment. You could create DEKs per `replay_id` or per `project_id`, but then you're looking at centrally managing them. With the scale this is at, I'm not sure that's the route to go. By `replay_id` _might_ be feasible?

I'm not too familiar with the underlying design of Replay's storage, but it sounds like there is an option of storing data in a metadata row? If so, the wrapped DEK can be stored there safely in a column/key. "Wrapped" refers to the plaintext DEK being encrypted itself with the Key Encryption Key (KEK).

For the KEK, I recommend generating a unique key every day. That day's key is used to encrypt any DEKs associated with segments being written. With a key per day, we can just delete the oldest key on a 90 day rotate cycle to maintain compliance with our data retention standards. Data older than 90 days would be unusable once the associated KEK was deleted.

At this point, data deletion would be a matter of deleting the DEK for the particular segments (ex. setting the column to an empty string or null). The data is unusable at that point. In terms of accidentally over-reading a segment, the data would not be decrypt-able and a failure would be obvious (since you'd need two keys).

You get some added bonuses here too:
- data deletion is almost instantaneous (since you're just removing a key, not the actual bytes)
- there's no need to read and re-write files

cmanallen:
@mdtro You probably wouldn't recommend having a statically configured KEK live as an environment variable, right?  Probably not safe long-term.

I'm curious about failure mode of a daily generated KEK.  Let's say we have some periodic process that creates KEKs.  Let's also say this process experiences an outage and no one notices (or doesn't notice quick enough).  Our ingestion pipeline would begin to fail because a KEK for that day could not be located.

So I'm thinking what if we generate KEKs in batches?  That would mean we have a long lead time to fix any problems associated with generating KEKs.  Is that a security liability?  The keys won't be in use until their due date so I'm under the impression that if we were compromised we could conceivably generate new keys for the days that have not elapsed.

cmanallen:
> How do you feel instead about buffering all segments from a single replay together and storing those as individual blobs?

This question has been asked several different ways by several different people so I'd like to respond to it broadly.  Thank you to everyone who asked.  Its an important question with deep technical implications.  I'm feeling a little pressured to get a response out today so I will be very quick, direct, and I will skip a lot of opportunities for deeper technical explanations.

The first question I would ask is: why?  Why would we want to do that?

Does it provide a better cost structure in GCS?  The average replay is roughly 10 segments.  I want to store more than 200 segments per file.  So at a minimum we have a 20x difference in cost -- so no it does not have a better COGS outcome.  

Maybe storing the segments together improves deletes in some way?  With @mdtro's comment I think we've come to the conclusion that deleting files from GCS is not a concern and that encryption is both a viable and sanctioned alternative.  

Is it safer and more fault-tolerant?  The design of a system that aggregates like-segments will necessarily be less safe and more prone to distributed faults than our existing service.  Why?  Because its an _in addition to_ piece of infrastructure.  Its another component that can fail.  It does not replace an existing component.  ...But lets say it did (which drastically expands the scope), the state in the state machine has an undefined life-span.  Because Kafka can backlog its not the maximum life of a Replay.  So when we flush we can never be truly certain that the entire replay was contained within.  And so we're back to storing a replay in multiple files.

Does it improve the liveness of the data?  The opposite.  Data is retained for a minimum of an hour before its made available.  Unless you expose the stateful service's internals via REST API (another significant addition to the scope).

Even if this were a high value goal, do we have the organization capability to execute on this?  No.  Maybe a couple people in this company could construct such a stateful service (in a safe and fault tolerant way) but again we circle back to the first question: what is the utility of storing like-segments together?  And is it greater than the value those minds are providing elsewhere in the org?

Put simply:

Its technically possible to do.  But its difficult to create, more expensive, prone to cascading faults in production, reduces the quality of the replay product, and unnecessarily couples a system to replay which otherwise could be used by other products.

Hopefully this clears things up :-)

mdtro:
> @mdtro You probably wouldn't recommend having a statically configured KEK live as an environment variable, right? Probably not safe long-term.
> 
> I'm curious about failure mode of a daily generated KEK. Let's say we have some periodic process that creates KEKs. Let's also say this process experiences an outage and no one notices (or doesn't notice quick enough). Our ingestion pipeline would begin to fail because a KEK for that day could not be located.
> 
> So I'm thinking what if we generate KEKs in batches? That would mean we have a long lead time to fix any problems associated with generating KEKs. Is that a security liability? The keys won't be in use until their due date so I'm under the impression that if we were compromised we could conceivably generate new keys for the days that have not elapsed.

Ah, that's a really good point and I failed to leave out some detail. The wrapped DEK would include the `key id` of the KEK used to encrypt it. You wouldn't need to do any timestamp logic to determine which key to use. Rotating them each day doesn't need to be a perfectly timed orchestration. So if a KEK fails to rotate a particular day, you'd just use the existing current one.

- Encryption would always be done with the currently configured KEK.
- Decryption would be done with the KEK key id defined in the wrapped DEK.

I'd use [Cloud KMS](https://cloud.google.com/kms/docs) which has [some solid uptime guarantees](https://cloud.google.com/kms/sla).  Combined with the above behavior, you'd maintain resilience in the ingest pipeline.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,252 @@
...
+
+File parts can be grouped into like-retention-periods and deleted manually or automatically after expiry. However, in the case of replays, storage costs are minor. We will retain our encrypted segment data for the maximum retention period of 90 days.
+
+## Key Rotation
+
+If a KEK is compromised and needs to be rotated we will need to follow a four step process. First, we query for every row in the "file_part_byte_range" table whose DEK was encrypted with the old KEK. Second, we will decrypt every DEK with the old KEK. Third, we will encrypt the DEK with a new KEK. Fourth, the old KEK is dropped.
+
+DEKs are more complicated to rotate as it requires modifying the blob. However, because DEKs are unique to a byte range within a single file we have a limited surface area for a compromised key to be exploited. To rotate a DEK first download the blob, second decrypt the byte range with the compromised DEK, third generate a new DEK, fourth encrypt the payload with the new DEK, fifth encrypt the new DEK with any KEK, and sixth upload and re-write the metadata rows with the new offsets.
```
mdtro, line 118: This section is 🔥 . I was so excited to read it. Awesome work!

text/0098-store-multiple-replay-segments-in-a-single-blob.md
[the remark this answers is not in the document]
mdtro: With the recent updates to the RFC regarding the use of envelope encryption -- I would consider this resolved. 👍

text/0098-store-multiple-replay-segments-in-a-single-blob.md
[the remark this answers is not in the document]
mdtro: With the recent updates to the RFC regarding envelope encryption -- this is also resolved. 👍 

Deletes would not require re-writing the blob. Instead the relevant metadata row with it's DEK would be deleted. In practice, this makes the data completely unusable and is functionally deleted.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
mdtro: With the recent updates to the RFC regarding envelope encryption -- I would consider this mitigated. 👍

If bad byte range math happens, the extra bytes would not be decrypt-able without the matching Data Encryption Key.

Do we know how the system would behave if it were to be fed random bytes?

cmanallen: The system doesn't exist yet but presumably it would error.  If we were persistently writing bad byte ranges (buggy ingestion process) then the replays would be lost.

cmanallen: Also thanks @bmckerry for bringing this up and @mdtro for the great cryptography advice!  Team work :D

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: I'm so stoked to hear that!  Thank you, I really appreciate this feedback :D

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+
+# Background
+
+This document was originally written to respond to a percieved problem in the Session Replays recording consumer. However, upon exploring these ideas more it was determined that this could be more generally applied to the organization as a whole. For that reason I've made many of the names generic but have also retained many references to Session Replay.
+
+# Supporting Data
+
+Google Cloud Storage lists the costs for writing and storing data as two separate categories. Writing a file costs $0.005 per 1000 files. Storing that file costs $0.02 per gigabyte. For the average Session Replay file (with a retention period of 90 days) this works out to: $0.000000012 for storage and $0.00000005 for the write.
+
+In practical terms, this means 75% of our spend is allocated to writing new files.
```
fpacifici, line 24: While the cost is mostly due to write, at scale is this cost problematic? Like is it a blocker for reaching higher scale or are we simply looking for a more efficient option ?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+1. We will store multiple "parts" per file.
+   - A "part" is a distinct blob of binary data.
+   - It exists as a subset of bytes within a larger set of bytes (referred to as a "file").
+   - A "part" could refer to a replay segment or to a sourcemap or anything that requires storage in a blob storage service.
+2. Each "part" within a file will be encrypted.
+   - Encryption provides instantaneous deletes (by deleting the row containing the encryption key) and removes the need to remove the byte sub-sequences from a blob.
+   - We will use envelope encryption to protect the contents of every file.
+     - https://cloud.google.com/kms/docs/envelope-encryption
+   - Related, contiguous byte ranges will be encrypted independently of the rest of the file.
+   - We will use KMS to manage our key-encryption-keys.
```
fpacifici, line 37: KMS ?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: I do not believe cost will prevent us from reaching greater scale.  Write cost scale linearly.  Pricing does not quite scale linearly but if you're happy with how much GCS costs at low levels of demand you will be happy at peak demand.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: Google Key Management Service.  Subject to change.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+
+| id  | key | path     | start | stop  | dek      | kek_id | created_at          |
+| --- | --- | -------- | ----- | ----- | -------- | ------ | ------------------- |
+| 1   | A:0 | file.bin | 0     | 6241  | Aq3[...] | 1      | 2023-01-01T01:01:01 |
+| 2   | B:0 | file.bin | 6242  | 8213  | ppT[...] | 1      | 2023-01-01T01:01:01 |
+| 3   | A:1 | file.bin | 8214  | 12457 | 99M[...] | 1      | 2023-01-01T01:01:01 |
+
+- The key field is client generated identifier.
+  - It is not unique.
+  - The value of the key field should be easily computable by your service.
```
fpacifici, line 58: Why not ensuring this is unique ?

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+  - The DEK is the key that was used to encrypt the byte range.
+  - The key itself is encrypted by the KEK.
+    - **K**ey **E**ncryption **K**ey.
+  - Encryption is explored in more detail in the following sections.
+- The "kek_id" column contains the ID of the KEK used to encrypt the DEK.
+  - This KEK can be fetched from a remote **K**ey **M**anagement **S**ervice or a local database table.
+
+Notice each row in the example above points to the same file but with different start and stop locations. This implies that multiple, independent parts can be present in the same file. A single file can be shared by hundreds of different parts.
+
+Second, the Session Replay recording consumer will not commit blob data to Google Cloud Storage for each segment. Instead it will buffer many segments and flush them together as a single blob to GCS. Next it will make a bulk insertion into the database for tracking.
```
fpacifici, line 76: How is this buffering process going to look like? 
It will have to rely on persistent state to prevent you from loosing chunks in case of failover before a file is fully ingested while parts have been already committed in Kafka.

Also a more high level concern with this. 
If the idea is to build a generic system reusable by other features, relying on a specific consumer to do the buffering has the important implication that Kafka is always going to be a moving part of your system. Have you considered doing the buffering separately.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: At least once delivery guarantees.  If you're bulk inserting 1000 rows per upload then it becomes difficult to split out the rows that have duplicates.

But, thinking on this more... my plan is to use Kafka for this.  Ordering guarantees could mean that the file batch is deterministic (assuming we don't have a stateful countdown timer).  If 1 row exists in the database then they all must exist (therefore the batch was already written - commit offsets and move on).  I suppose re-balancing would not affect this?  A unique constraint could be possible.

That being said I think a deadline is important so files don't sit idle in the consumer for undefined periods of time.

cmanallen: Then again a read query against the unique key prior to the insert operation could satisfy this constraint.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+
+In practical terms, this means 75% of our spend is allocated to writing new files.
+
+# High Level Overview
+
+1. We will store multiple "parts" per file.
+   - A "part" is a distinct blob of binary data.
+   - It exists as a subset of bytes within a larger set of bytes (referred to as a "file").
+   - A "part" could refer to a replay segment or to a sourcemap or anything that requires storage in a blob storage service.
+2. Each "part" within a file will be encrypted.
```
fpacifici, line 32: Since encryption is used here entirely to make deletion of individual parts quicker, could you please expand on:
- whether we are confident that this is a use case worth optimizing for (deleting individual part quickly) 
- how would the process look like if we just accepted to rewrite files entirely in order to remove the part.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+# Questions
+
+1. If KEKs are managed in a remote service how do we manage outages?
+   - We manage it in the same way we manage an outage of any other remote service.
+     - We will need to backlog or otherwise 400/404.
+   - KEKs have the benefit of _probably_ not blocking ingest as the key will be cached for long stretches of time (24 hours) and can be used for longer periods of time if a new key can not be fetched.
+2. How will read efficiency be impacted if we rely on a remote service to decrypt blob data?
+   - It will have some cost but hopefully that cost is minimized by the constraints of your system.
+   - For example, Session Replay fetches multiple segment blobs in a single request. At most we will need to fetch two keys (and in the majority of cases a single key) to decrypt the segments.
+   - This key fetching latency is immaterial to the total latency of the request.
```
fpacifici, line 160: What is the self hosted open source implementation for this

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Consumer Buffering Mechanics
+
+The following section is highly specific to the Session Replay product.
+
+We will continue our approach of using _at-least once processing_. Each message we receive is guaranteed to be processed to completion regardless of error or interrupt. Duplicate messages are possible under this scheme and must be accounted for in the planning of each component.
+
+**Buffer Location and Behavior**
+
+The buffer is kept as an in-memory list inside the consumer process. For each message we receive we append the message to the buffer. Afterwards, we check if the buffer is full. If it is we flush. Else we wait for another message.
```
fpacifici, line 222: It seems this will constrain you to a fairly small buffer size. 
Also it ties the number of replicas of your consumer to the cost efficiency of the storage, which is quite undesirable:
Assuming you are never going to commit on kafka untill the buffer is flushed (if you did you would not be able to guarantee at least once):
- If you increase the number of replicas for any reason, each replicas takes less traffic, thus it takes longer to fill the buffer.
- Unless you tie the commit batch time and size to the number of replicas (which is undesirable - see the snuba consumer), increasing the number of replicas would increase the number of files written per unit of time. 
replicas and file size should not be connected to each other, otherwise we will have to keep in mind a lot of affected moving parts when scaling the consumer.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
```diff
@@ -0,0 +1,282 @@
...
+
+This is a somewhat simplified view of whats happening. In reality we will use deadline flushing in addition to a count or resource-usage based flushing. This ensures the buffer does not stay partially full indefinitely.
+
+**Buffer Flush**
+
+On flush the buffer will take every message in the list and merge them together into a single bytes object. This bytes object will then be uploaded to the storage service-provider. Upon successful upload the start and stop byte range values of each message are stored in a database in addition to other metadata such as their replay_id and segment_id. Finally, the last offset is committed to Kafka.
+
+**Handling Consumer Restarts**
+
+If the consumer restarts with a non-empty buffer, the buffer's last item's offset will not be committed. When the consumer resumes it will start processing from the last offset committed (i.e. the last item in the last successfully-flushed-buffer). The buffer will be rebuilt exactly as it was prior to restart.
```
fpacifici, line 232: What is the target size of each file you want to write and the target buffer size ? 
The approach of throwing away the in flight work is generally ok if the buffer are very small and the time it takes to reprocessing is negligible. If this is not the case (you want large files) you may create a lot of noise when kafka rebalances consumer groups and create a lot of backlog.

fpacifici (commented):
A few high level remarks:

What is the rationale for merging parts into single blob outside of Filestore instead of evolving Filestore to achieve the result? In this implementation it seems you are targeting a generic solution that is meant to be reusable in other scenarios, but if we go that way we will end up in a split world where one "filestore" is great for latency but unable to merge parts into files and the "other filestore" is horrible at latency but able to merge parts into files.
This seems an undesirable situation and I would strongly advice to consolidate into one solution that handles files well across the board instead (whether that is Filestore or a brand new service). 

Could you please expand on how this would be supported for open source customers? IT seems to depend on additional external services we do not have today, they must have a self hosted alternative. In general also it is a good idea to be wary of solutions that are substantially different in different environments. We tend to avoid them unless strictly necessary as maintenance becomes a lot harder and the chances to break the system is higher.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: > How is this buffering process going to look like?

Like buffering would in our Snuba consumers.  You keep an array of messages/rows/bytes in memory.  When it comes time to flush you zip them together in some protocol specific format.  In this case the protocol is byte concatenation.

> It will have to rely on persistent state to prevent you from loosing chunks in case of failover before a file is fully ingested while parts have been already committed in Kafka.

If we assume that the replays consumer pushes the bytes to another, generic Kafka consumer which buffers the files before upload then the persistent state will be the log.  Upload failures can be re-run from the last committed offset.  Persistent failures would have to be handled as part of a DLQ and would require re-running.  Potentially introducing a significant amount of latency between (in the replays case) a billing outcome and the replay recording being made available.

Assuming this buffering/upload step exists inside our existing consumer (i.e. not a generic service) then offsets will not be committed until after the batch has been uploaded.

> Also a more high level concern with this.
If the idea is to build a generic system reusable by other features, relying on a specific consumer to do the buffering has the important implication that Kafka is always going to be a moving part of your system. Have you considered doing the buffering separately.

I have considered that.  The idea being process A buffers a bunch of file-parts/offsets before sending the completed object to permanent storage either through direct interaction, filestore, or Kafka intermediary (or any number of intermediaries).  The problem is then that each call site is responsible for the buffer which is the "hardest" part of the problem.

Kafka is an important part of this system in my mind.  When I wrote this document I relied on the guarantees it provides.  I think if this is a generic service made available to the company then publishing to the topic should be a simple process for integrators.  The fact that its Kafka internally is an irrelevant implementation detail (to the caller).

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: > whether we are confident that this is a use case worth optimizing for (deleting individual part quickly)

It is necessary to support GDPR, project, and point deletes.  I consider it a high priority.  The alternative is rotating the file with the offending parts removed.

> how would the process look like if we just accepted to rewrite files entirely in order to remove the part.

The operations is no particular order:

1. Download the file
2. Get the active byte ranges from the database.
3. Remove all byte ranges from the file not found in the set of returned byte ranges.
4. Delete all references to the file in the database.
5. Upload the new file.
6. Insert new offset rows.
7. Delete the old file.

Repeat for every delete operation.  Deletes must be single-threaded per file to prevent concurrent access.  You can use kafka and partition on filename.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: Environment variable named `SECRET`.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: If the time to accumulate a batch is less than the time to upload a batch then you need to add a replica.  That's the only constraint.  You get more efficiency at peak load so its best to run our replicas hot.  The deadline will prevent the upload from sitting idle too long.  The total scale factor will be determined by the number of machines we can throw at the problem.

Multi-processing/threading, I think, will be deadly to this project.  So we will need a lot of single-threaded machines running.

I re-wrote this response several times.  Its as disorganized as my thoughts are on this.  Happy to hear critiques.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: These are the commit criterion I have (naively) envisioned.  All these values are adjustable.

- Commit if filesize >= 10MB.
- Commit if num_parts >= 1000.
- Commit if five-second-deadline exceeded.

10MB and 1000 parts can be scaled down to 1MB and 100 parts if they seem unrealistic.  Any lower and the concept I think has run its course.

> the time it takes to reprocessing is negligible

I believe this will be the case.  A generic consumer implementation should only be doing byte concatenation.  But depending on size it may take a while to fetch those files over the network to even begin buffering.

I would prefer this generic consumer deployed independently of getsentry so re-balances are less common.

cmanallen:
@fpacifici 

> What is the rationale for merging parts into single blob outside of Filestore instead of evolving Filestore to achieve the result? In this implementation it seems you are targeting a generic solution that is meant to be reusable in other scenarios, but if we go that way we will end up in a split world where one "filestore" is great for latency but unable to merge parts into files and the "other filestore" is horrible at latency but able to merge parts into files.
This seems an undesirable situation and I would strongly advice to consolidate into one solution that handles files well across the board instead (whether that is Filestore or a brand new service).

Why should these cases be mutually exclusive?  A buffering consumer could point to filestore as its storage destination.

That being said I'm happy to improve filestore in whatever capacity you think I'm able.  Just point me in a direction and I'll go.

> Could you please expand on how this would be supported for open source customers? IT seems to depend on additional external services we do not have today, they must have a self hosted alternative. In general also it is a good idea to be wary of solutions that are substantially different in different environments. We tend to avoid them unless strictly necessary as maintenance becomes a lot harder and the chances to break the system is higher.

I think I addressed these cases in my other responses but if not.  Things like KMS can be replaced with an environment variable and the Kafka consumer is part of the broader sentry ecosystem of technologies.  Everything else is just code.

cmanallen:
@fpacifici 

> Have you considered doing the buffering separately.

I've been thinking about this a lot and it made me realize that I goofed.  When I originally wrote this RFC it was for an in-memory buffer within the replays consumer.  You reduce service costs and simplify your network overhead by batching the files together.  That all makes sense to me.

Then this generic consumer idea emerged.  I got really excited about it and forgot one crucial thing.  Publishing to Kafka isn't free.  This idea only works if you totally disregard its network cost.  I've taken 1000 network calls and made it 1000 + 1 network calls.  Those 1000 publishes to Kafka could be faster but faster isn't free and free is what I was looking for.

So with that in mind I'm going to revert the RFC back to the point where its just talking about an in-memory buffer on the replays consumer.  So I think this RFC no longer has any org-wide impact other than whatever minor libraries or database tables emerge from it.

text/0098-store-multiple-replay-segments-in-a-single-blob.md
cmanallen: > Unless you tie the commit batch time and size to the number of replicas (which is undesirable - see the snuba consumer), increasing the number of replicas would increase the number of files written per unit of time.

I think this is an okay outcome.  If you double the number of replicas you halve the number of parts per file and double the number of files.  That reduces cost efficiency but the throughput efficiency remains the same for each replica.  Ignoring replica count, cost efficiency will ebb and flow with the variations in load we receive throughout the day.

We still come out ahead because the total number of files written per second is less than the current implementation which is 1 file per message.

cmanallen:
Following up from my July 13th comment from above.  Having reviewed the Kafka producer implementation I do believe a generic Kafka consumer is a viable path and should be explored.

mwarkentin:
Some of these encryption / deletion concerns came up again as part of https://getsentry.atlassian.net/browse/INC-1210 where bulk deletion of replays in GCS was doable-but-painful in GCS, and deemed infeasible at the scale needed on the Clickhouse side.

Revisiting crypto shredding is an action item: https://getsentry.atlassian.net/browse/IA-184

PR #99, opened 2023-05-30
Amend RFC 92 With New Options

cmanallen:
[Rendered RFC](https://github.com/getsentry/rfcs/blob/ebad7640b55368a216020e4d383fd86b378811dd/text/0092-replay-issue-creation.md)

### DIFF
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -78,4 +78,4 @@ When the SDK encounters a "slow click" it will append the slow click to the reco
 
 # Decisions
 
-No decision has been made.
+We have decided to go with option **3** and couple these new issues to the Session Replay product and use the Session Replay back-end to create issues. We'll very likely need to continue iterating on the detection and be able to iterate on the server gives us a lot more flexibility. The reach will be smaller in terms of total user adoption (require Replay) but the we believe finding the right value to noise ratio is challenging. Having a Replay becomes very important for users to understand what happened and assess impact. 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -65,6 +65,7 @@ When the SDK encounters a "slow click" it will append the slow click to the reco
 
 - We can sample slow click events on the back-end without worrying about input from other teams.
 - No code changes required by the Issues team.
+- May be able to incentivize folks to install Replay in order to see dead click issues
 
 **Cons:**
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -70,7 +70,7 @@ When the SDK encounters a "slow click" it will append the slow click to the reco
 
 - Requires the Session Replay SDK.
   - Significantly smaller pool of customers who will see "slow click" issues.
-- Requies code changes by the Session Replay back-end team.
+- Requires code changes by the Session Replay back-end team.
   - Requires addition of event sampling, issue platform integration, and merging of the replay-event and recording-event payloads.
   - Merging the replay-event and recording-event payloads together is not a trivial change and requires careful deployment.
 
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -32,13 +32,8 @@ When the SDK encounters a "slow click" it will use `sentry.captureException(slow
 
 - Requires SDK changes.
 - Can not change issue sampling parameters without an SDK update.
-  - Or can you? It depends on how protected this endpoint is. If the teams responsible for maintaining it are not interested in the replay team adding `if is_slow_click(event): then do special sampling stuff`.
-
-**Questions:**
-
-- Can we sample these issue events on the back-end?
-  - Yes its certainly possible.
-  - However, it depends on how protected this endpoint is. If the teams responsible for maintaining it are not interested in the replay team adding `if is_slow_click(event): then do special sampling stuff` then we will not be able to meet this criteria.
+  - Or can you?
+  - What if we sample within the error processing pipeline on the back-end?
   - If error processing is functionally immutable (i.e. there is no organizational will-power to allow product teams to sample within the error pipeline) then we will not be able to sample on the back-end.
 
 ### Option 2: SDK Creates Issues Through an Issues HTTP Endpoint
@@ -50,18 +45,16 @@ When the SDK encounters a "slow click" it will use `http.post("/issues", data=sl
 - Significantly larger customer reach (i.e. all javascript SDK customers).
 - Opportunity to upsell customers on Session Replay from a "slow click" event.
 - No code changes required on the Session Replay back end.
+- We can sample slow click issues on the backend.
+  - This would require buy-in and coordination from the Issues team.
+  - There is nothing preventing us from looking at an HTTP request and making a go/no-go decision.
 
 **Cons:**
 
 - Requires SDK changes.
 - Requies code changes by the Issues team to create a generic interface for creating issues.
-
-**Questions:**
-
-- Can we sample these issue events on the back-end?
-  - Yes its certainly possible.
-  - However, it depends on how protected this endpoint is. If the teams responsible for maintaining it are not interested in the replay team adding `if is_slow_click(event): then do special sampling stuff` then we will not be able to meet this criteria.
-  - If the issue interface is functionally immutable (i.e. there is no organizational will-power to allow product teams to sample within this interface) then we will not be able to sample on the back-end.
+  - This would likely disrupt our June 18 deadline.
+  - Unless the Issues team has excess capacity and a willingness to work on it immediately.
 
 ### Option 3: Replay SDK Pushes Issues to Session Replay Backend Which Raises an Issue
 
@@ -69,14 +62,15 @@ When the SDK encounters a "slow click" it will append the slow click to the reco
 
 **Pros:**
 
-- We can sample slow click events without worrying about input from other teams.
+- We can sample slow click events on the back-end without worrying about input from other teams.
 - No code changes required by the Issues team.
 - No SDK changes required as the SDK is already sending slow-click events.
 
 **Cons:**
 
-- Significantly smaller pool of customers who will see "slow click" issues.
-- No opportunity to upsell the Session Replay product.
+- Requires the Session Replay SDK.
+  - Significantly smaller pool of customers who will see "slow click" issues.
+  - No/Limited opportunity to upsell the Session Replay product.
 - Requies code changes by the Session Replay back-end team.
   - Requires addition of event sampling, issue platform integration, and merging of the replay-event and recording-event payloads.
   - Merging the replay-event and recording-event payloads together is not a trivial change and requires careful deployment.
--- a/text/0092-replay-issue-creation.md
+++ b/text/0092-replay-issue-creation.md
@@ -17,43 +17,72 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
 
 # Options Considered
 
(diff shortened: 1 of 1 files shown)


text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+  - There is nothing preventing us from looking at an HTTP request and making a go/no-go decision.
 
 **Cons:**
 
-1. Poor rollout could impact service availability during testing period.
-2. Requires coordination between the SDK and Ingest to create new issue types.
+- Requires SDK changes.
+- Requies code changes by the Issues team to create a generic interface for creating issues.
+  - This would likely disrupt our June 18 deadline.
+  - Unless the Issues team has excess capacity and a willingness to work on it immediately.
```
JoshFerge, line 57: it also seems like we'd also want to do this in Relay -- we try not to have python endpoints be hit by SDKs. seems like it would be a large amount of work.

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
-   - E.g. "Experienced 10 occurences in the past hour".
-4. Replay is guaranteed to be sampled.
+- Significantly larger customer reach (i.e. all javascript SDK customers).
+- Opportunity to upsell customers on Session Replay from a "slow click" event.
+- No code changes required on the Session Replay back end.
+- We can sample slow click issues on the backend.
+  - This would require buy-in and coordination from the Issues team.
+  - There is nothing preventing us from looking at an HTTP request and making a go/no-go decision.
 
 **Cons:**
```
JoshFerge, line 52: the big problem i see is _noise_. from our tests it seems like dead clicks could drown out error quota quite quickly.

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
 
 # Options Considered
 
-### Option 1: SDK Generates Issues Through a Generic Interface
+### Option 1: SDK Creates Issues Through the Error Interface
```
bruno-garcia, line 20: How about just:
```suggestion
### Option 1: SDK Creates Issues Through the SDK
```

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
 
-When the SDK encounters a "replay issue" it will make an HTTP request to a generic interface which will handle the issue creation process.
+When the SDK encounters a "slow click" it will use `sentry.captureException(slowClick)` to create an error issue on the Issues platform.
```
bruno-garcia, line 22: It'll be an envelope with an event in it. We could skip the "Issues Platform" bit?

```suggestion
When the SDK encounters a "slow click" it will use `sentry.captureEvent(slowClick)`.
```

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+
+- We can sample slow click events on the back-end without worrying about input from other teams.
+- No code changes required by the Issues team.
+- No SDK changes required as the SDK is already sending slow-click events.
+
+**Cons:**
+
+- Requires the Session Replay SDK.
+  - Significantly smaller pool of customers who will see "slow click" issues.
+  - No/Limited opportunity to upsell the Session Replay product.
```
bruno-garcia, line 73: They would already be using Replay, so we could omit this item. one would argue the feature itself is the upsell (Use Replay to get Dead Click Issues):

```suggestion
```

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+- Requies code changes by the Session Replay back-end team.
```
bruno-garcia, line 74:
```suggestion
- Requires code changes by the Session Replay back-end team.
```

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+
+### Option 3: Replay SDK Pushes Issues to Session Replay Backend Which Raises an Issue
+
+When the SDK encounters a "slow click" it will append the slow click to the recording events payload. The back-end will search for these events and then publish them to the Issues platform's kafka consumer.
+
+**Pros:**
+
+- We can sample slow click events on the back-end without worrying about input from other teams.
+- No code changes required by the Issues team.
+- No SDK changes required as the SDK is already sending slow-click events.
```
bruno-garcia, line 67: We wouldn't need any further changes? I imagine we wouldn't keep it in its current form. For example, not emitting anything for short durations, for example.

If that's true, perhaps the "No SDK change X SDK change" pro/con items could all be removed. Or adjusted with "Complex vs Simple SDK changes" or any other more specific description.

@mydea @billyvg can probably give us the best approach with this

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
 
-1. Uses errors quota.
-2. Unclear if we're able to use dynamic thresholds.
-   - E.g. "Experienced 10 occurences in the past hour".
-3. The replay containing the issue could not be sampled.
+- Requires SDK changes.
+- Can not change issue sampling parameters without an SDK update.
+  - Or can you?
+  - What if we sample within the error processing pipeline on the back-end?
+  - If error processing is functionally immutable (i.e. there is no organizational will-power to allow product teams to sample within the error pipeline) then we will not be able to sample on the back-end.
```
bruno-garcia, line 37:
```suggestion
  - If error processing is functionally immutable (i.e. there is no organizational will-power to allow product teams to sample within the error pipeline) then we will not be able to sample on the back-end.
- Consumes from user error quota: If too noisy, might give the impression we're 'wasting' their error quota
```

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
 
-### Option 2: SDK Generates Issues Through a Session-Replay Specific Interface
+### Option 2: SDK Creates Issues Through an Issues HTTP Endpoint
 
-The SDK publishes a "replay issue" to the Replay back-end. The back-end will decide to process the issue or not.
+When the SDK encounters a "slow click" it will use `http.post("/issues", data=slowClick)` to create an error issue on the Issues platform.
```
bruno-garcia, line 41: Not sure this is feasible. Given we'd need the whole mechanism of retry/backpressure, serialization, etc. It would need to be a new `envelopeType` and use the same `captureEnvelope` mechanism of the SDK instead.

text/0092-replay-issue-creation.md
cmanallen: I think we should change it to `captureException` to be clear the mechanism we're using.  "SDK ... Through the SDK" is a bit circular.

text/0092-replay-issue-creation.md
cmanallen: I think we're saying the same things here.  My extra text is confusing.  Happy to remove it.

text/0092-replay-issue-creation.md
cmanallen: My thinking was "As a result of choosing option X we would or would not need to ask the SDK team to do something".  The SDK team can always make changes to the detection strategy but the outcome of this RFC wouldn't compel changes.

I can remove this text as its a minor point and it seems to cause confusion.

text/0092-replay-issue-creation.md
cmanallen: I'm not sure the exact implementation details.  This was an idea proposed by the Issues team.  A generic /issues endpoint which product teams could make requests to.  My http client psuedo-code shouldn't be interpreted as production code.  I'm only trying to indicate that there's a new /issues HTTP endpoint which is submitted to by the SDK.

bruno-garcia: Would this be rolled out on Relay? How do we handle all the ingestion scaling, PoP's etc? The endpoint could be used internally by the consumer if that helps somehow, but to be called by the clients, I imagine we'll need to stick to `/envelope` and Relay. But happy keep this option here, just raising my concerns.

cmanallen: Not sure of the implementation details.  I would imagine the endpoint would be added to Relay and messages forwarded to the Issues Kafka consumer.

mitsuhiko:
Out of the options here this really should be option 3. I don’t like the idea if creating issues directly from the SDK. On the one hand for abuse reasons, on the other for flexibility reasons if we want to change how these issues behave.

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,65 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+  - Unless the Issues team has excess capacity and a willingness to work on it immediately.
+
+### Option 3: Replay SDK Pushes Issues to Session Replay Backend Which Raises an Issue
+
+When the SDK encounters a "slow click" it will append the slow click to the recording events payload. The back-end will search for these events and then publish them to the Issues platform's kafka consumer.
+
+**Pros:**
+
+- We can sample slow click events on the back-end without worrying about input from other teams.
+- No code changes required by the Issues team.
```
jas-kas, line 67:
```suggestion
- No code changes required by the Issues team.
- May be able to incentivize folks to install Replay in order to see dead click issues
```

bruno-garcia (commented):
:shipit:

bruno-garcia (approved):
Since they inverted the buttons, it really gets me

JoshFerge (approved):

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,65 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
+
+**Cons:**
+
+- Requires the Session Replay SDK.
+  - Significantly smaller pool of customers who will see "slow click" issues.
+- Requires code changes by the Session Replay back-end team.
+  - Requires addition of event sampling, issue platform integration, and merging of the replay-event and recording-event payloads.
+  - Merging the replay-event and recording-event payloads together is not a trivial change and requires careful deployment.
 
 # Unresolved questions
```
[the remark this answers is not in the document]
ryan953: We should put both of these bullets as Cons inside (especially) option 1

text/0092-replay-issue-creation.md
```diff
@@ -17,43 +17,66 @@ These issues can only be detected on the SDK. The Replay back-end will never hav
...
 
 # Decisions
 
-We have decided to couple these new issues to the Session Replay product and use the Session Replay back-end to process issue events. Generic issue interfaces are not well supported at the time of writing. Using the Session Replay back-end we can accomplish our product goals. Should an HTTP interface for creating issue events be created this RFC can be re-addressed.
+We have decided to go with option **3** and couple these new issues to the Session Replay product and use the Session Replay back-end to create issues. We'll very likely need to continue iterating on the detection and be able to iterate on the server gives us a lot more flexibility. The reach will be smaller in terms of total user adoption (require Replay) but the we believe finding the right value to noise ratio is challenging. Having a Replay becomes very important for users to understand what happened and assess impact.
```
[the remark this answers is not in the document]
mydea: good point, and very true 👍

Issue #28, opened 2022-10-18
[DECISION] Add RFC for top level span creation method


AbhiPrasad:
ref:
- https://github.com/getsentry/sentry-laravel/pull/598
- https://github.com/getsentry/sentry-java/issues/1154

- https://github.com/getsentry/sentry-javascript/pull/3753

Introduce a `Sentry.trace` method that wraps a callback with a created span.

sl0thentr0py:
in ruby we just added 
https://github.com/getsentry/sentry-ruby/pull/1783
we can rename in a major for consistency if this becomes unified API

AbhiPrasad:
Wrote out why this is need for OpenTelemetry here: https://github.com/getsentry/develop/issues/720

PR #101, opened 2023-06-12
rfc(decision): Revamping the SDK Performance API

AbhiPrasad:
[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/revamping-the-sdk-performance-api/text/0101-revamping-the-sdk-performance-api.md)

closes https://github.com/getsentry/rfcs/issues/28

This RFC proposes to revamp the performance API in the SDKs. The new API aims to accomplish the following:

- Align both the internal schemas and top level public API with OpenTelemetry and their SDKs.
- De-emphasize the concept of transactions from users using a Sentry performance monitoring SDK.
- Optimize for making sure performance data is always sent to Sentry.
- Open the SDK up for future work where we support batch span ingestion instead of relying on a transaction solely.

The RFC proposal comes in two steps.

1. Introduce three new top level methods, `Sentry.startActiveSpan` and `Sentry.startSpan` and their language specific variants as well as `Sentry.setMeasurement`.

2. ~Introduce a new span schema that is aligned with OpenTelemetry.~ Out of scope for this RFC, will be addressed later.

### DIFF
--- a/text/0101-revamping-the-sdk-performance-api.md
+++ b/text/0101-revamping-the-sdk-performance-api.md
@@ -186,96 +186,42 @@ To remove the overhead of understanding transactions/spans and their differences
 
 The current transaction schema inherits from the error event schema, with a few fields that are specific to transactions.
 
-```rs
-// Based on https://github.com/getsentry/relay/blob/2ad761f64db3df9b4d42f2c0896e1f6d99c16f49/relay-general/src/protocol/event.rs
-pub struct Event {
-    /// Transaction name of the event.
-    /// This field is called `name` by most SDKs and is accessed by `transaction.setName()` and `transaction.getName()`
-    pub transaction: Annotated<String>,
+A full version of this protocol can be seen in [Relay](https://github.com/getsentry/relay/blob/2ad761f64db3df9b4d42f2c0896e1f6d99c16f49/relay-general/src/protocol/event.rs), but here are some of the fields that are important:
 
-    /// Timestamp when the event was created. Some SDKs alias this to `end_timestamp` but convert
-    /// it to `timestamp` when serializing to send to Sentry.
-    pub timestamp: Annotated<Timestamp>,
-
-    /// Timestamp when the event has started.
-    pub start_timestamp: Annotated<Timestamp>,
-
-    /// Custom tags for this event.
-    pub tags: Annotated<Tags>,
-
-    /// Spans for tracing.
-    pub spans: Annotated<Array<Span>>,
-
-    /// Measurements which holds observed values such as web vitals.
-    pub measurements: Annotated<Measurements>,
-}
-```
+| Transaction Field | Description                                                                                                                                                | Type            |
+| ----------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------- |
+| `name`            | Name of the transaction. In ingest and storage this field is called `transaction`                                                                          | String          |
+| `end_timestamp`   | Timestamp when transaction was finished. Some SDKs alias this to `end_timestamp` but convert it to `timestamp` when serializing to send to Sentry.         | String \| Float |
+| `start_timestamp` | Timestamp when transaction was created.                                                                                                                    | String \| Float |
+| `tags`            | Custom tags for this event. Identical in behaviour to tags on error events.                                                                                | Object          |
+| `spans`           | A list of child spans to this transaction.                                                                                                                 | Span[]          |
+| `measurements`    | Measurements which holds observed values such as web vitals.                                                                                               | Object          |
+| `contexts`        | Contexts which holds additional information about the transaction. In particular, `contexts.trace` has additional information about the transaction "span" | Object          |
 
 The transaction also has a trace context, which contains additional fields about the transaction.
 
-```rs
-// https://github.com/getsentry/relay/blob/2ad761f64db3df9b4d42f2c0896e1f6d99c16f49/relay-general/src/protocol/contexts/trace.rs
-pub struct TraceContext {
-    /// The trace ID.
-    #[metastructure(required = "true")]
-    pub trace_id: Annotated<TraceId>,
-
-    /// The ID of the span.
-    #[metastructure(required = "true")]
-    pub span_id: Annotated<SpanId>,
-
-    /// The ID of the span enclosing this span.
-    pub parent_span_id: Annotated<SpanId>,
-
-    /// Span type (see `OperationType` docs).
-    pub op: Annotated<OperationType>,
-
-    /// Whether the trace failed or succeeded. Currently only used to indicate status of individual
-    /// transactions.
-    pub status: Annotated<SpanStatus>,
-}
-```
+| Transaction Field | Description                                                                                                                   | Type   |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------- | ------ |
+| `trace_id`        | Trace ID of the transaction. Format is identical between Sentry and OpenTelemetry                                             | String |
+| `span_id`         | Span ID of the transaction. Format is identical between Sentry and OpenTelemetry                                              | String |
+| `parent_span_id`  | Parent span ID of the transaction. Format is identical between Sentry and OpenTelemetry                                       | String |
+| `op`              | Operation type of the transaction. [Standardized by Sentry spec](https://develop.sentry.dev/sdk/performance/span-operations/) | String |
+| `status`          | Status of the transaction. Sentry maps status to HTTP status codes, while OpenTelemetry has a fixed set of status'            | String |
 
 The current span schema is as follows:
 
-```rs
-// https://github.com/getsentry/relay/blob/2ad761f64db3df9b4d42f2c0896e1f6d99c16f49/relay-general/src/protocol/span.rs
-pub struct Span {
-    /// Timestamp when the span was ended.
-    #[metastructure(required = "true")]
-    pub timestamp: Annotated<Timestamp>,
-
-    /// Timestamp when the span started.
-    #[metastructure(required = "true")]
-    pub start_timestamp: Annotated<Timestamp>,
-
-    /// Human readable description of a span (e.g. method URL).
-    pub description: Annotated<String>,
-
-    /// Span type (see `OperationType` docs).
-    pub op: Annotated<OperationType>,
-
-    /// The Span id.
-    #[metastructure(required = "true")]
-    pub span_id: Annotated<SpanId>,
-
-    /// The ID of the span enclosing this span.
-    pub parent_span_id: Annotated<SpanId>,
-
-    /// The ID of the trace the span belongs to.
-    #[metastructure(required = "true")]
-    pub trace_id: Annotated<TraceId>,
-
-    /// The status of a span
-    pub status: Annotated<SpanStatus>,
-
-    /// Arbitrary tags on a span, like on the top-level event.
-    pub tags: Annotated<Object<JsonLenientString>>,
-
-    /// Arbitrary additional data on a span, like `extra` on the top-level event.
-    pub data: Annotated<Object<Value>>,
-}
-```
+| Span Field        | Description                                                                                                                                                                                           | Type            |
+| ----------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------- |
+| `description`     | Description of the span. Same purpose as transaction `name`                                                                                                                                           |
+| `trace_id`        | Trace ID of the span. Format is identical between Sentry and OpenTelemetry                                                                                                                            | String          |
+| `span_id`         | Span ID of the span. Format is identical between Sentry and OpenTelemetry                                                                                                                             | String          |
+| `parent_span_id`  | Parent span ID of the span. Format is identical between Sentry and OpenTelemetry                                                                                                                      | String          |
+| `end_timestamp`   | Timestamp when span was finished. Some SDKs alias this to `end_timestamp` but convert it to `timestamp` when serializing to send to Sentry.                                                           | String \| Float |
+| `start_timestamp` | Timestamp when span was created.                                                                                                                                                                      | String \| Float |
+| `op`              | Operation type of the span. [Standardized by Sentry spec](https://develop.sentry.dev/sdk/performance/span-operations/)                                                                                | String          |
+| `status`          | Status of the span. Sentry maps status to HTTP status codes, while OpenTelemetry has a fixed set of status'                                                                                           | String          |
+| `tags`            | Custom tags for this span.                                                                                                                                                                            | Object          |
+| `data`            | Arbitrary additional data on a span, like `extra` on the top-level event. We maintain [conventions for span data keys and values](https://develop.sentry.dev/sdk/performance/span-data-conventions/). | Object          |
 
 As you can see, the fields on the transaction/span differ in a few ways, the most noteable of which is that transactions have `name` while spans have `description`. This means that spans and transactions are not interchangeable, and users have to know the difference between the two.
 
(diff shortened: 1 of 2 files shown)


text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+// span is finished, so parent is put back onto scope
+Sentry.getCurrentHub().getScope().setSpan(parentSpan);
+```
+
+Most users do the above also when nested within their application, as often you assume that a transaction is defined that you can attach (very common in a web server context).
+
+To add instrumentation to their applications, users have to know concepts about hubs/scopes/transactions/spans and understand all the different nuances and use cases. This can be difficult, and presents a big barrier to entry for new users.
+
+Also, since transactions/spans are different classes (span is a parent class of transaction), users have to understand the impacts that the same method will have on both the transaction/span. For example, currently calling `setTag` on a transaction will add a tag to the transaction event (which is searchable in Sentry), while `setTag` on a span just adds it to the span, and the field is not searchable. `setData` on a span adds it to `span.data`, while `setData` on a transaction is undefined behaviour (some SDKs throw away the data, others add it to `event.extras`).
```
marandaneto, line 91: Isn't it the other way around? the transaction is a parent class of span, at least on Mobile.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+// ): Promise<T>;
+```
+
+In the ideal case, `startActiveSpan` should generally follow this code path.
+
+1. Get the active span from the current scope
+2. If the active span is defined, create a child span of that active span based on the provided `spanContext`, otherwise create a transaction based on the provided `spanContext`.
+3. Run the provided callback
+4. Finish the child span/transaction created in step 2
+5. Remove the child span/transaction from the current scope and if it exists, set the previous active span as the active span in the current scope.
```
marandaneto, line 143: I'm wondering if this solves (or not) the use case that you have an active span (span A), and spawn multiple threads to download something, and each of those threads will start a new span (span B and C), but then the relationship of the spans are messed up because they "race each other", for example, span B and C should be a child of span A, but what I see is that span C is a child of span B.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+sentry.StartSpanFromContext(ctx, spanCtx)
+```
+
+Or when continuing from headers in javascript:
+
+```js
+Sentry.startSpanFromHeaders(spanCtx, headers);
+```
+
+Since we want to discourage accessing the transaction object directly, the `Sentry.setMeasurement` top level method will also be introduced. This will set a custom performance metric if a transaction exists.
```
marandaneto, line 176: Would be the case that you set a `measurement` to a specific span, which is a span in our data model, and we don't have the `measurements` field in the span data model, only in the Transaction data model, how would we make this difference? are we planning to introduce a new data model and give up on Transaction and Span?

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+| `tags`            | Custom tags for this span.                                                                                                                                                                            | Object          |
+| `data`            | Arbitrary additional data on a span, like `extra` on the top-level event. We maintain [conventions for span data keys and values](https://develop.sentry.dev/sdk/performance/span-data-conventions/). | Object          |
+
+As you can see, the fields on the transaction/span differ in a few ways, the most noteable of which is that transactions have `name` while spans have `description`. This means that spans and transactions are not interchangeable, and users have to know the difference between the two.
+
+In addition, user's have the burden to understand the differences between `name`/`description`/`operation`. `operation` in particular can be confusing, as it overlaps with transaction `name` and span `description`. In addition, `operation` is not a required field, which means that it is not clear what the default value should be.
```
marandaneto, line 231: IIRC `operation` was always required in our data model.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+The new span schema is as follows:
+
+| Span Field        | Description                                                  | Type            | Notes                                                                                                                                                                                                                                                              |
+| ----------------- | ------------------------------------------------------------ | --------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `name`            | The name of the span                                         | String          | Should be low cardinality. Replacing span `description`                                                                                                                                                                                                            |
```
marandaneto, line 243: I believe is important to keep `description`, how would I set the SQL statement of a DB query if it has to be low cardinality? One can say you can do it under the `data` field, so the `name` then would be similar to the former `operation`? some operations just don't have a unique name, is the grouping of transactions still per `name`?

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,268 @@
...
+
+For the purposes of this RFC, the version on the span schema will be set to 2. This will indicate to all consumers that the new span schema is being used.
+
+Just like both the old Sentry schema and the OpenTelemetry schema, we keep the same fields for `span_id`, `trace_id`, `parent_span_id`, `start_timestamp`, and `end_timestamp`. We also choose to rename `description` to `name` to match the OpenTelemetry schema.
+
+Having both the `name` and `op` fields is redundant, but we choose to keep both for backwards compatibility. There are many features in the product that are powered by having a meaningful operation, more details about this can be found in the documentation around [Span operations](https://develop.sentry.dev/sdk/performance/span-operations/). In the future, we can choose to deprecate `op` and remove it from the schema.
+
+The most notable change here is to formally introduce the `attributes` field, and remove the `span.data` field. This is a breaking change, but worth it in the long-term. If we start accepting `attributes` on transactions as well, we more closely align with the OpenTelemetry schema, and can use the same conventions for both spans and transactions.
```
marandaneto, line 260: Can we keep the `data` that is gonna be deprecated and forwarded to `attributes` for compatibility?
eg
`span.setData(x, y)` just calls `span.setAttribute(x, y)` internally.

text/0101-revamping-the-sdk-performance-api.md
AbhiPrasad: I guess the caveat here is that the hub/scope needs to be properly forked and correctly follow the execution context.

I guess in single scope instances, this logic breaks down re: parent-child relationships. Let me amend the RFC to reference this.

text/0101-revamping-the-sdk-performance-api.md
AbhiPrasad: In the OpenTelemetry conventions, it's recommended db spans be named <db.operation> <db.name>.<db.sql.table> - like `SELECT bestDb.orders`, but the only thing they require is that something is low cardinality, so the db query also works here. In most cases, we can just add whatever we were setting in `span.description` to `span.name` (as long as it's low cardinality).

Grouping of transactions/spans would still be by name, but there is more freedom about what that should be. The product instead relies on span attributes to make this work.

text/0101-revamping-the-sdk-performance-api.md
AbhiPrasad: Yup this is a good idea! I will also put more details into migrating between the different span schemas.

text/0101-revamping-the-sdk-performance-api.md
marandaneto: Not sure why it's a mandatory field in [Mobile](https://github.com/getsentry/sentry-java/blob/98fb9c76b168e19a601f82e878a5d539c1698e1a/sentry/src/main/java/io/sentry/ISpan.java#L17) then, either it was documented like that or is an oversight on our end, thanks for the link.

text/0101-revamping-the-sdk-performance-api.md
marandaneto: I see, DB queries are high cardinality though, at least for bigger databases.
Most of the time interceptors/hooks just give you a `String`, so to build `SELECT bestDb.orders` requires parsing the String on the client which is also not ideal.
Alright, this is more an implementation detail anyway, thanks.

philipphofmann: I'm OK removing `op` and adding a new low cardinality format of OpenTelemetry as you described @AbhiPrasad, but I think the description we currently have for db queries, HTTP requests, and such is highly useful. Where would that end up now?

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+
+Transactions also have no mechanism for arbitrary additional data like spans do with `data`. Users can choose to add arbitrary data to transactions by adding it to the `contexts` field (as transactions extend the event schema), but this is not obvious and not exposed in every SDK. Since contexts are already well defined in their own way, there is no way of using [Sentry's semantic conventions for span data](https://develop.sentry.dev/sdk/performance/span-data-conventions/) for transactions.
+
+### New Span Schema
+
+To simplify how performance data is consumed and understood, we are proposing a new span schema that the SDKs send to Sentry. The new span schema aims to be a superset of the [OpenTelemetry span schema](https://github.com/open-telemetry/opentelemetry-proto/blob/4967b51b5cb29f725978362b9d413bae1dbb641c/opentelemetry/proto/trace/v1/trace.proto) and have a minimal top level API surface. This also means that spans can be easily converted to OpenTelemetry spans and vice versa.
+
+The new span schema is as follows:
```
philipphofmann, line 238: Can we also keep `origin` https://develop.sentry.dev/sdk/performance/trace-origin/? It's handy to know what precisely created a span.

philipphofmann (commented):
Excellent that you try to address some critical issues of our performance API @AbhiPrasad 👏. I gave this a first quick pass, but I'll have to take some more time after my vacation next week and think about how the new API will play out on mobile.

text/0101-revamping-the-sdk-performance-api.md
marandaneto: @philipphofmann the span attributes, aka `span.data`

text/0101-revamping-the-sdk-performance-api.md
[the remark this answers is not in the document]
marandaneto: Today spans have tags, but they are not indexed, so they can just be span attributes.
But I'd love to see indexed fields from spans as well, I believe this is important.

marandaneto: `origin` can also be part of the span attributes if needed, but good point.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+# Improvements
+
+## New SDK API
+
+The new SDK API has the following requirements:
+
+1. Newly created spans must have the correct trace and parent/child relationship
+2. Users shouldn’t be burdened with knowing if something is a span/transaction
+3. Spans only need a name to identify themselves, everything else is optional.
+4. The new top level APIs should be as similar to the OpenTelemetry SDK public API as possible.
```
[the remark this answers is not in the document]
marandaneto: I'd be interested in that as well, what are the pro x cons of moving closer to opentelemetry rather than our own model?
I know a few of them but it'd be nice to be written down.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+| `status`          | Status of the span                                           | String          | An optional final status for this span. Can have three possible values: 'ok', 'error', 'unset'. Same as OpenTelemetry's Span Status                                                                                                                                |
+| `attributes`      | A set of attributes on the span.                             | Object          | This maps to `span.data` in the current schema for spans. There is no existing mapping for this in the current transaction schema. The keys of attributes are well known values, and defined by a combination of OpenTelemtry's and Sentry's semantic conventions. |
+| `measurements`    | Measurements which holds observed values such as web vitals. | Object          |                                                                                                                                                                                                                                                                    |
```
iker-barriocanal, line 251: Is this a good time to structure the data further and define what `Object` looks like for each field? The more we structure the data, the better product experience we can provide on the server side.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+- Optimize for making sure performance data is always sent to Sentry.
+- Open the SDK up for future work where we support batch span ingestion instead of relying on a transaction solely.
+
+# Planned Work
+
+The RFC proposal comes in two steps.
+
+1. Introduce three new top level methods, `Sentry.startActiveSpan` and `Sentry.startSpan` and their language specific variants as well as `Sentry.setMeasurement`.
+
+This allows us to de-emphasize the concept of hubs, scopes and transactions from users, and instead have them think about just spans. Under the hood, `Sentry.startActiveSpan` and `Sentry.startSpan` should create transactions/spans as appropriate. `Sentry.setMeasurement` is used to abstract away `transaction.setMeasurement` and similar.
```
cleptric, line 22: `Sentry.startActiveSpan` should be the default, hence called `Sentry.startSpan`.
Introducing more top-level functions (even undocumented) increases the API surface unnecessarily and adds more confusion.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+3. Run the provided callback
+4. Finish the child span/transaction created in step 2
+5. Remove the child span/transaction from the current scope and if it exists, set the previous active span as the active span in the current scope.
+
+If the provided callback throws an exception, the span/transaction created in step 2 should be marked as errored. This error should not be swallowed by `startActiveSpan`.
+
+`startActiveSpan` only provides the correct parent-child relationship if your platform has proper support for forking scopes. For platforms have a single hub/scope (like the mobile SDKs), this method will not lead to the correct parent-child relationship. The SDK will have to provide a different method for these platforms. One option is for `startActiveSpan` to always attach the span it creates to the root span (the transaction), which means users don't get exact parent-child relationships, but they do get relative relationships between spans using relative durations.
+
+`Sentry.startSpan` will create a span, but not set it as the active span in the current scope.
```
cleptric, line 150: Any use-cases come to mind why this function is necessary?

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+
+For example with go:
+
+```go
+sentry.StartSpanFromContext(ctx, spanCtx)
+```
+
+Other SDKs can also add a `startSpanFromContext` method where they can make context a variable databag that accepts headers, environment variables, etc. This is useful for easy trace propagation.
+
+Since we want to discourage accessing the transaction object directly, the `Sentry.setMeasurement` top level method will also be introduced. This will set a custom performance metric if a transaction exists. If a transaction doesn't exist, this method will do nothing. In the future, this method will attach the measurement to the span on the scope, but for now it'll only attach it to the transaction.
```
cleptric, line 175: What about `Sentry.setMetric`? `setMeasurement` might set the expectation to measure something, which is not the case.

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+
+Just like both the old Sentry schema and the OpenTelemetry schema, we keep the same fields for `span_id`, `trace_id`, `parent_span_id`, `start_timestamp`, and `end_timestamp`. We also choose to rename `description` to `name` to match the OpenTelemetry schema.
+
+Having both the `name` and `op` fields is redundant, but we choose to keep both for backwards compatibility. There are many features in the product that are powered by having a meaningful operation, more details about this can be found in the documentation around [Span operations](https://develop.sentry.dev/sdk/performance/span-operations/). In the future, we can choose to deprecate `op` and remove it from the schema.
+
+The most notable change here is to formally introduce the `attributes` field, and remove the `span.data` field. This is a breaking change, but worth it in the long-term. If we start accepting `attributes` on transactions as well, we more closely align with the OpenTelemetry schema, and can use the same conventions for both spans and transactions.
+
+### Shimming the old schema
```
cleptric, line 261: What are we going to do with the `extra` stuff?

text/0101-revamping-the-sdk-performance-api.md
HazAT: What if we turn it around - instead of having `startActiveSpan` we just have `startSpan` that does by default what `startActiveSpan` does (meaning putting it on the scope) - I suppose this is what people would expect 99% of the time. We could add an argument that prevents setting it on the scope if someone sets it.

marandaneto: That's what we do on Mobile, `startTransaction` with a `bindToScope` param, but it's `bindToScope=false` by default.

If this flag is enabled by default, because of single Hub/Scope on Mobile, might be overwriting each other transactions as a side-effect, not sure if it's a good idea, at least on the Mobile world, maybe keeping `bindToScope=false` by default makes more sense in this case.

The other option is that if it's enabled by default, but there's a living transaction in the scope, you just create a span, but maybe this is not the desired behavior and we'd need a way to opt-out, in this case, people would need to force `bindToScope=false`.

Ergonomics isn't great again because they will not know if there's a living transaction in the scope.

text/0101-revamping-the-sdk-performance-api.md
marandaneto: Mobile has single Hub/Scope and this is needed in case you want your own transaction instead of the auto-instrumented ones, measuring something side by side.
For example, a auto-instrumented transaction during screen loading, but also a service doing the X job, that has nothing to do with the screen loading transaction.

text/0101-revamping-the-sdk-performance-api.md
marandaneto: https://github.com/getsentry/rfcs/pull/101#discussion_r1230910664

text/0101-revamping-the-sdk-performance-api.md
```diff
@@ -0,0 +1,273 @@
...
+## New SDK API
+
+The new SDK API has the following requirements:
+
+1. Newly created spans must have the correct trace and parent/child relationship
+2. Users shouldn’t be burdened with knowing if something is a span/transaction
+3. Spans only need a name to identify themselves, everything else is optional.
+4. The new top level APIs should be as similar to the OpenTelemetry SDK public API as possible.
+
+There are two top level methods we'll be introducing to achieve this: `Sentry.startActiveSpan` and `Sentry.startSpan`. `Sentry.startActiveSpan` will take a callback and start/stop a span automatically. In addition, it'll also set the span on the current scope. Under the hood, the SDK will create a transaction or span based on if there is already an existing span on the scope.
```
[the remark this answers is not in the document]
krystofwoldrich: Active for `active in the scope` in the scope make sense, but since we try to shield users from the internal scope and hub. 

`startActiveSpan` looks to me as `active` for the automatic finish.
 `startSpan` as a manual variant of the `startAtiveSpan`

AbhiPrasad:
After a lot of deliberation finally pushed updates to this RFC.

We've moved the span schema changes out of scope of this RFC as this requires further discussion in ingest/product. As a result the RFC now only focuses on introducing the new top level methods for the performance API.

The RFC has also been amended to give more freedom to SDK authors to decide what methods to best introduce, although the recommendation is to use `startActiveSpan` and `startSpan`. This was done because certain languages/runtimes might have more idiomatic patterns for span creation, and we should prefer that over forcing a single unified API everywhere. We see this in OpenTelemetry where each otel SDK has it's own set of span creation methods.

```go
// golang
ctx, span := tracer.Start
```

```csharp
// .NET
using var myActivity = MyActivitySource.StartActivity
```

```ruby
# ruby
MyAppTracer.in_span("do_work") do |span|
  # do some work that the 'do_work' span tracks!
end
```

```java
# Java
Span childSpan = tracer.spanBuilder("child")
    .setParent(Context.current().with(parentSpan))
    .startSpan();
```

PR #104, opened 2023-06-20
rfc(informational): String Parameterization Output Spec

gggritso:
This RFC proposes a spec for the output of string parameterization. At Sentry, the SDKs, Relay, and Ingest all do some amount of parameterization and sometimes they disagree on the output format. This causes inconsistent UX, duplicated code, and various bugs. The RFC proposes a spec for what parameterized strings (whether transaction names, span description, etc.) should look like.

It's meant as a starting point for us to discuss, explore, and agree on a consistent parameterization output. This is motivated mostly by Starfish, where successful parameterization is a key feature.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/string-parameterization-output-spec/text/0104-string-parameterization-output-spec.md)

### DIFF
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -25,7 +25,7 @@ The main goal of parameterization is to remove dynamic values from strings in or
 
 Firstly, when looking at application data (traces, URLs, spans, etc.) it is most useful to look at parameterized values, because they represent aggregate behaviour. In the routing example, the `/users/:id` route is a useful entity to examine, while the unparameterized URLs are not. The first and most important outcome of parameterization is successful aggregation. Aggregation makes it possible to store, query, and compare data easily.
 
-The second outcome is downstream processing. Successful and consistent parameterization allows downstream features like Performace Issues, Span Groups, and Starfish to work correctly. Parameterization allows identifying which spans are abstractly "the same", which powers downstream features, and UIs.
+The second outcome is downstream processing. Successful and consistent parameterization allows downstream features like Performance Issues, Span Groups, and Starfish to work correctly. Parameterization allows identifying which spans are abstractly "the same", which powers downstream features, and UIs.
 
 Lastly, good parameterization will reduce cardinality, and therefore reduce data storage costs and data processing times.
 
@@ -49,7 +49,7 @@ In another example, the SDKs might pass the query `SELECT * FROM country WHERE c
 
 In another example, Ingest and Relay both do some parameterization of URLs, for different purposes. Relay will parameterize host names like `someclient.api.com` into `*.api.com` for span metrics. Ingest does not do that for span groups. This creates inconsistency in the system, where span groups might not match span metrics.
 
-Lastly, we do not give format parameterization recommendatiosn to our users. If users implement manual parameterization (e.g., in a `before_send` hook) incorrectly, their spans will be ineligible for downstream features like Performance issues.
+Lastly, we do not give format parameterization recommendations to our users. If users implement manual parameterization (e.g., in a `before_send` hook) incorrectly, their spans will be ineligible for downstream features like Performance issues.
 
 This spec documents the desired output of a parameterizer, regardless of context. This makes it possible for different parameterizers to agree, conform to a standard, share code, and create consistent correct behaviour in our systems.
 
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -148,13 +148,15 @@ Given:
 GET http://someservice.com/users/tom/
 ```
 
-If the parameterizer determines that `tom` is a route parameter named `:name`, it must return:
+If the parameterizer determines that `tom` is a route parameter named `:userName`, it must return:
 
 ```
-GET http://someservice.com/users/{name}
+GET http://someservice.com/users/{userName}
 ```
 
-**Note:** The parameterizer should replace the smallest possible substring. e.g., the URL `/service/id-17` should be parameterized as `/service/id-{user_id}` rather than `/service/{user_id}`.
+**Note:** The parameterizer should replace the smallest possible substring. e.g., the URL `/service/id-17` should be parameterized as `/service/id-{userId}` rather than `/service/{userId}`.
+
+**Note:** The parameterizer must output parameter names in `camelCase` format.
 
 ### Unknown Route Segments
 
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -7,7 +7,7 @@
 
 ## Summary
 
-This RFC describes a spec for the output of a string parameterizer. A parameterizer is a system that accepts text strings, detects dynamic parameters in those strings, and replaces those parameters with placeholders. A parameter can be a value in a SQL query, a path segment in a URL, or any value interpolated into a string. A successful parameterizer can check a string (or a set of strings), and determine whether any part of those strings corresponds to a dynamic value.
+This RFC describes a spec for the output of string parameterizers, and proposes that we adopt the spec and recommend it to our users. A parameterizer is any system or function that accepts text strings, detects dynamic parameters in those strings, and replaces those parameters with placeholders. A parameter can be a value in a SQL query, a path segment in a URL, or any value interpolated into a string. A successful parameterizer can check a string (or a set of strings), and determine whether any part of those strings corresponds to a dynamic value.
 
 Parameterization applies to many aspects of distributed tracing, like transaction names, span description, and breadcrumbs. For example, consider a traced application that returns cities in given countries by fetching them from a SQL database. Each SQL query will create a corresponding span, and the query will be the span’s description. Over its runtime, the application might generate these descriptions:
 
@@ -29,9 +29,13 @@ A successful parameterizer will ingest these span descriptions, and determine th
 SELECT * FROM cities WHERE country = ?;
 ```
 
-where `?` is a country code parameter that varies from query to query.
+where `?` is a country code parameter that varies from query to query. Examples or parameterizers in Sentry include:
 
-This spec outlines standards for the output of a parameterizer, including but not limited to the desired string formatting, payload structure, and character escaping. The spec is organized by data type. Parameterization methods and infrastructure are outside the scope of this spec.
+- `beforeSend` hooks written by users to strip out URL segments for manual instrumentation
+- Django instrumentation that replaces URLs with corresponding Django routes
+- Relay's URL parameterizer
+
+This RFC describes a spec that provides standards for the output of a parameterizer. The spec explains the desired string formatting, the payload structure, and character escaping. Parameterization methods and infrastructure are outside the scope of this spec.
 
 ## Motivation
 
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -26,7 +26,7 @@ SELECT * FROM cities WHERE country = 'AU';
 A successful parameterizer will ingest these span descriptions, and determine that `"CA"`, `"US"`, and `"AU"` were probably a parameter inserted dynamically by the application depending on some condition. The three query descriptions are examples of this parameterized query:
 
 ```sql
-SELECT * FROM cities WHERE county = ?;
+SELECT * FROM cities WHERE country = ?;
 ```
 
 where `?` is a country code parameter that varies from query to query.
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -65,6 +65,7 @@ This spec documents the desired output of a parameterizer, regardless of context
 - Parameterize *every* parameter. If a string contains multiple parameters, parameterize each one according to the relevant rules
 - Re-order unordered parameters in alphabetical order if it does not change the meaning of the string (e.g., see “Query Strings”)
 - Do not alter the string in any other way (e.g., do not re-format the whitespace, quotes, or any other symbols)
+- Output syntax-highlightable, but not necessarily valid strings (e.g., parameterized SQL queries don't have to be executable as-is, but they should be syntax-highlightable)
 
 ## Parameterizing SQL Commands
 
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -1,0 +1,35 @@
+- Start Date: 2023-06-20
+- RFC Type: informational
+- RFC PR: https://github.com/getsentry/rfcs/pull/104
+- RFC Status: draft
+
+# Summary
+
+One paragraph explanation of the feature or document purpose.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Background
+
+The reason this decision or document is required. This section might not always exist.
+
+# Supporting Data
+
+[Metrics to help support your decision (if applicable).]
+
+# Options Considered
+
+If an RFC does not know yet what the options are, it can propose multiple options. The
+preferred model is to propose one option and to provide alternatives.
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+# Unresolved questions
+
+- What parts of the design do you expect to resolve through this RFC?
+- What issues are out of scope for this RFC but are known?
--- a/text/0104-string-parameterization-output-spec.md
+++ b/text/0104-string-parameterization-output-spec.md
@@ -249,6 +249,30 @@ Return:
 SADD orgs:*:emails jane@myorg.com
 ```
 
+## Structured JSON
+
+Databases like MongoDB structure operations as JSON payloads. In structured JSON languages, **the parameterizer must replace every leaf parameter with a single unquoted `?` character**.
+
+Given:
+
+```
+{
+  status: {
+    $in: ["active"]
+  }
+}
+```
+
+Return:
+
+```
+{
+  status: {
+    $in: [?]
+  }
+}
+```
+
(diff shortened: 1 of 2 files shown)


text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+- **Parameter**: A substring in a span description that represents a dynamic value that was substituted in by code. e.g., in the string `/users/17`, `17` is a likely parameter
+- **Parameterization**: The act of determining parameters, and replacing them with a substitution character. Also sometimes called "normalization"
+- **Parameterizer**: A piece of code, or a system that performs parameterization
+- **Substitution character**: A character or string of characters that denote a parameter. e.g., in the string `SELECT * FROM countries WHERE id = ?;` the `?` is a substitution character
+
+## Overall Guidelines
+
+- Parameterize *every* parameter. If a string contains multiple parameters, parameterize each one according to the relevant rules
+- Re-order unordered parameters in alphabetical order if it does not change the meaning of the string (e.g., see “Query Strings”)
```
gggritso, line 64: I'm not sure this is a good idea. I think the parameterized string should be as close to the original string as possible. If the parameterized string is later used to create a _hash_ or some kind of other unique identifier, that part of the code can decide to re-arrange things in order to ensure consistent hashing.

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+SQL queries usually appear in descriptions of `db` spans. In a SQL query (or any SQL statement) a parameter is any value that is either generated randomly, or supplied from a variable in the code. The following symbols are *not* considered parameters, and should not be altered or re-ordered:
+
+- column names
+- table names
+- aliases
+
+### Simple Parameters
+
+For individual parameters, **the parameterizer must replace the parameter with a single unquoted `?` character**.
```
untitaker, line 77: is the intent to get rid of all the variations like `%s` or `%(named_param)s`? Those templates are often 1:1 forwarded from the underlying DB driver, of which e.g. the Python SDK is mostly agnostic. Different DBs technically have different ways to substitute strings. They are very similar to each other as far as postgres/mysql/sqlite is concerned, but what about random non-SQL stuff like MongoDB queries?

Not clear to me where you want this normalization to happen. I think it would be good to only do it serverside in Relay, make some educated guesses on how string formatting works (similar to how we do it already for logentry) and preserve the original template somewhere in case the parametrizer misbehaves.

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+- `SELECT * FROM countries WHERE id = ?`
+- `SELECT * FROM countries WHERE id = %s`
+- `SELECT * FROM countries WHERE id = $1`
+
+As another example, the SDKs might pass the query `SELECT * FROM country WHERE code IN (?, ?)` to Ingest, which will further parameterize that as `SELECT * FROM country WHERE code IN (?)`. The string `(?, ?)` is technically only partially parameterized from the system's perspective.
+
+This has important downstream effects on user experience. A good parameterizer will correctly group similar strings together in a way that represents the original intent. If parameterization is successful, users will see groups that make sense to them. Good parameterization will also reduce cardinality, and therefore reduce data storage costs. A bad parameterizer will fill the users’s lists of transactions and spans with noisy similar descriptions that drown out the signal.
+
+This spec documents the desired output of a parameterizer. This makes it possible for different parameterizers to agree, conform to a standard, share code, and create consistency in the system.
```
untitaker, line 52: I am not sure why this consistency is even desired. Ensuring that all SQL uses `?` as parameter can feel "foreign" to a user familiar with a particular database. In the case of C-style formatting, `%(named_parameter)` vs `%(named_parameter)s` also has different meaning. The spec also locks in named parameters in some cases to be formatted with curly braces, which only feels native in some languages (rust, python)

Is there something specific we want to do in the UI with those parameters, such as syntax-highlight them? It might be possible to use _meta protocol from Relay here, though it is a very complex solution. Basically what it does is to allow you to annotate ranges of characters of a string with custom data:

```
{
   "myvalue": "/foo/{bar}/baz",
   "_meta": {"myvalue": {"": "rem": [["@parametrizer","s",6,10]]
}
```

iker-barriocanal: Expanding this a bit more: this goal tries to set a standard across all organizations, while specific organizations may follow a single pattern and don't have different formats. In the latter case, we'd be reformatting their queries for no real benefit. I don't think we should have a different approach per organization but approach this on a language basis, as Markus suggests (we can always have a fallback in Relay).

text/0104-string-parameterization-output-spec.md
iker-barriocanal: What is the problem we're trying to solve with this point? Is the following reasoning fair?

1. SDKs capture parameters in a certain order, consistently. The order they capture the parameters is always the same.
2. We can consider the parameters to not be ordered, according to sorting rules we've considered "correct".
3. A set of parameters that is always in the same order is already ordered, by some rules that we may not know/have defined.
4. Thus, changing the order doesn't effectively make any improvements on the ordering, while introducing more complexity.

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+```
+GET http://someservice.com/users/tom/
+```
+
+If the parameterizer determines that `tom` is a route parameter named `:name`, it must return:
+
+```
+GET http://someservice.com/users/{name}
+```
```
iker-barriocanal, line 153: As with the section above trying to set a common substitution character, setting an arbitrary format may be confusing for users manually setting custom instrumentation. Is this something we're ok with?

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+Return:
+
+```
+GET http://someservice.com:8080/
+```
+
+### Auth Strings
+
+**If the URL contains authentication strings, the parameterizer must replace them with a `*` character.**
```
iker-barriocanal, line 213: [Advanced data scrubbing](https://docs.sentry.io/product/data-management-settings/scrubbing/advanced-datascrubbing/#methods) is a powerful feature Sentry offers to users to deal with sensitive data. Do we think it's a good idea to ignore this feature and redact the strings in a different way?

jjbayer: Specifically, in advanced data scrubbing, the user can chose whether they want to [remove, mask, hash, or replace](https://docs.sentry.io/product/data-management-settings/scrubbing/advanced-datascrubbing/#methods) the original value (which is then documented in the `_meta` field, as @untitaker described [above](https://github.com/getsentry/rfcs/pull/104/files#r1236533531)).

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,264 @@
...
+
+Here are three ways different SDKs might parameterize a SQL query:
+
+- `SELECT * FROM countries WHERE id = ?`
+- `SELECT * FROM countries WHERE id = %s`
+- `SELECT * FROM countries WHERE id = $1`
+
+As another example, the SDKs might pass the query `SELECT * FROM country WHERE code IN (?, ?)` to Ingest, which will further parameterize that as `SELECT * FROM country WHERE code IN (?)`. The string `(?, ?)` is technically only partially parameterized from the system's perspective.
+
+This has important downstream effects on user experience. A good parameterizer will correctly group similar strings together in a way that represents the original intent. If parameterization is successful, users will see groups that make sense to them. Good parameterization will also reduce cardinality, and therefore reduce data storage costs. A bad parameterizer will fill the users’s lists of transactions and spans with noisy similar descriptions that drown out the signal.
```
jjbayer, line 50: Everything in this paragraph is true, but how would standardizing the placeholder format help with this?

Standardizing could be a good idea, but this section could use some more convincing points as to why that is IMO. Do we have examples of cases where the product fails to deal with placeholders because they are in the wrong format, or a string is altered by two different parameterizers and therefore looks messy (e.g. `/{my}/<parameterized>/:string`?).

gggritso: Great feedback, thank you! I will update this section. There are actually a few goals here, and I didn't explain them fully.

1. Downstream processing edge cases. When SDKs use different placeholder formats, all downstream processing of descriptions has to handle each kind of placeholder. We ran into this with performance issues, suspect spans, and now span metrics. All three of them have to handle `?` vs. `%s` vs. `$1` as database placeholders. If a new SDK comes in with a fourth format, all have to be updated
2. We haven't seen any UI cases of mixed parameterization, but that's what I'm trying to prevent. For example, transaction names on the frontend and backend are both URL-like, but look different. `/api/0/organizations/{organization_slug}/events-trends-stats/` vs. `/projects/:projectId/`. Why curly brackets in one, but colons in the other? Right now they're never mixed, but it's bad precedent

You're right, a standard placeholder does _not_ inherently help with cardinality. One aspect of the spec that _does_ help with cardinality is demanding that lists of arguments like `IN (? ? ? ?)` are collapsed. Otherwise they'd all have different descriptions. Not a very strong case, but it's one example!

text/0104-string-parameterization-output-spec.md
gggritso: > SDKs capture parameters in a certain order, consistently. The order they capture the parameters is always the same.

Is this always true?

The original goal was to make sure that different parameter order doesn't affect parameterization, in cases where the SDKs do _not_ capture parameters in a certain order consistently. Can SDKs guarantee that? If yes, we don't need this point. If not, we might want to specify that parameterization ensures the order

text/0104-string-parameterization-output-spec.md
gggritso: > is the intent to get rid of all the variations like %s or %(named_param)s?

The intent is for the SDKs to use the same parameterization character, if that's what you mean!

> but what about random non-SQL stuff like MongoDB queries?

This is a good question, I can add a section below for parameterization structures queries like MongoDB (and Snuba, technically!)

> Not clear to me where you want this normalization to happen.

This spec have no opinion at all on _where_ parameterization happens. I mention this early in the description, but I should make that more explicit. We parameterize things in all kinds of places (like the SDKs, and serverside), to your point.

I do agree that preserving the original string makes a lot of sense, I cover that point a little further below in "Storing Parameterization Results"

text/0104-string-parameterization-output-spec.md
gggritso: > Ensuring that all SQL uses ? as parameter can feel "foreign" to a user familiar with a particular database.

So far, the product folks don't think this will be a problem, but I'm curious about this. Do you know of a SQL-like database that uses a different character?

> The spec also locks in named parameters in some cases to be formatted with curly braces, which only feels native in some languages (rust, python)

I waffled on this a lot. I think curlies are familiar to almost all developers. Python, Ruby, JavaScript, Rust, PHP all use some kind of curly-like syntax for interpolation though they're not all the same.

> Is there something specific we want to do in the UI with those parameters, such as syntax-highlight them?

This is a great point, thanks for raising. I will think about this more. No, the spec is not concerned with ideas like syntax highlighting, but the `_meta` protocol is interesting!

gggritso: > I don't think we should have a different approach per organization but approach this on a language basis

I think I know what you mean, but I'm not sure, could you elaborate?

text/0104-string-parameterization-output-spec.md
gggritso: 🤔 that's an interesting point! I think this might _reduce_ confusion, because if we publish a spec on how to parameterize strings, users with custom instrumentation can refer to our spec, and their parameterization will be compatible with what our system expects

text/0104-string-parameterization-output-spec.md
iker-barriocanal: I believe I wasn't clear enough. If SDKs don't modify the order in any way, then SDKs always extract it the same way. This means that the order is whatever users wrote, and it doesn't matter if it's alphabetically sorted or not. Does this make sense?

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,293 @@
...
+SADD orgs:17:emails jane@myorg.com
+```
+
+Return:
+
+```
+SADD orgs:*:emails jane@myorg.com
+```
+
+## Structured JSON
```
iker-barriocanal, line 252: How are structured JSONs sent? Is this part of `span.data`?

text/0104-string-parameterization-output-spec.md
jjbayer: > This spec have no opinion at all on where parameterization happens.

@untitaker correct me if I'm wrong, but I think the concern is not about where the parameterization (i.e. replacing IDs with placeholders) happens, but where we ensure that these placeholders adhere to the standardization.

As @untitaker pointed out,

> Those templates are often 1:1 forwarded from the underlying DB driver, of which e.g. the Python SDK is mostly agnostic.

so we probably do not want to add the complexity of parsing and converting db-specific placeholders to standard-conform placeholders in the SDK itself.

text/0104-string-parameterization-output-spec.md
gggritso: The intention here is to cover cases where JSON data is stringified into a string field. For example, Snuba spans do this with their descriptions. This is not a common case as far as I know!

text/0104-string-parameterization-output-spec.md
gggritso: I'm curious to hear more about this, since converting db-specific placeholders to stands-conform is exactly what I'm suggesting! Is it a lot of work? Right now, the placeholders conform to the language of the SDK (or the framework) rather than the underlying database type. Unifying them would make downstream processing easier (see updated motivation section)

That said, I'm not suggesting we _have_ to unify them, but at a minimum I would like to document these placeholders, so there's a reference for which ones exist, and where, so any downstream processing can be consistent

text/0104-string-parameterization-output-spec.md
iker-barriocanal: Technically, the example below isn't valid JSON. Do we have a spec of what the language we're looking for looks like?

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,293 @@
...
+  "description": "POST example.com/user/123456",
+}
+```
+
+The parameterizer must output:
+
+```json
+{
+  "description": "POST example.com/user/123456",
+	"description.scrubbed": "POST example.com/user/%s"
```
jjbayer, line 291: There is an existing protocol for string substitutions, that [the UI partially understands](https://github.com/getsentry/sentry/blob/200624923c0066ae77710aba593f574ad59bb5c6/static/app/components/tagsTable.spec.tsx#L6-L44). It might make sense to reuse that. Can't find any documentation on it, but here's an [example](https://github.com/getsentry/relay/blob/ec72a5d6330581fab71330d3fa278bcdf4ffe488/relay-general/src/pii/snapshots/relay_general__pii__convert__tests__sdk_client_ip_stripped.snap#L5-L26) and [spec](https://github.com/getsentry/relay/blob/ec72a5d6330581fab71330d3fa278bcdf4ffe488/relay-general/src/types/meta.rs#L412-L430).

text/0104-string-parameterization-output-spec.md
jjbayer: Unifying them is definitely possible, but I wonder whether users actually _want_ them unified. E.g. A react user might prefer seeing `/api/:user` over `/api/{user}`. This is ultimately a product decision though, I have no strong opinion.

An alternative way to make downstream processing aware of what parts of the string are placeholders is to store annotations outside of the string itself, e.g. by using the `_meta` protocol mentioned further below.

text/0104-string-parameterization-output-spec.md
jjbayer: I should have picked a better example, sorry. The point is that `_meta` has an optional `"val"` field we can write the original value to:

https://github.com/getsentry/relay/blob/a127f96ee4870f9d676e3d0f39ab91ab8c557fcd/relay-general/src/types/meta.rs#L427-L429

gggritso: Ah interesting, thanks!

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,293 @@
...
+
+Return:
+
+```
+SADD orgs:*:emails jane@myorg.com
+```
+
+## Structured JSON
+
+Databases like MongoDB structure operations as JSON payloads. In structured JSON languages, **the parameterizer must replace every leaf parameter with a single unquoted `?` character**.
```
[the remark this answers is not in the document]
gggritso: That's fair! I think this "structured JSON" section might be out of scope completely

text/0104-string-parameterization-output-spec.md
```diff
@@ -0,0 +1,293 @@
...
+```
+
+Return:
+
+```
+{
+  status: {
+    $in: [?]
+  }
+}
```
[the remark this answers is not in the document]
gggritso: The JavaScript SDK allows these kinds of keys without quotes, yeah! You're right, this isn't even JSON, it's actually a structured JS object, which is way looser than JSON

gggritso:
Long overdue, but I'm going to close this PR! Huge thanks to everyone who contributed to this! As we've worked on Starfish (primarily on the database view) the Relay team has put in a _ton_ of work on parametrization, and it's _still changing_ and getting better. I don't think this RFC serves a purpose until that work settles down, and even then, it would be more productive to either write an RFC for databases specifically, or just extract the code that does the parameterization.

Thank you @jjbayer @iker-barriocanal @untitaker 👍🏻 We'll come back to this in the future (maybe).

PR #105, opened 2023-06-21
feat: Update RFC#91 with new token format

mydea:
This PR updates RFC #91 with an adjusted token format. We did some experimentation and started implementation and noticed shortcomings of the JWT format, so we decided to change course here to a more simple base64 format.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/fn/update-org-token-format/text/0091-ci-upload-tokens.md)

### DIFF
--- a/text/0091-ci-upload-tokens.md
+++ b/text/0091-ci-upload-tokens.md
@@ -62,7 +62,7 @@ sntrys_eyJpYXQiOjE2ODczMzY1NDMuNjk4NTksInVybCI6bnVsbCwicmVnaW9uX3VybCI6Imh0dHA6L
 ```
 
 * `PREFIX`: `sntrys_` - this is static and helps to identify this is a Sentry token.
-* `FACTS`: A padding-less base64 encoded JSON string of the facts.
+* `FACTS`: A base64 encoded JSON string of the facts.
 * `SECRET`: A random secret part for the token. We may use `b64encode(secrets.token_bytes(32)).decode("ascii").rstrip("=")`, but this is an implementation detail. 
 
 A serialized token is added a custom prefix `sntrys_` (sentry structure) to make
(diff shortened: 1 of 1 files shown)


text/0091-ci-upload-tokens.md
```diff
@@ -271,6 +272,12 @@ globally unique IDs.  However this today does not work for a handful of reasons:
    for frontend + backend deployment scenarios being able to use one token to manage releases
    across projects might be desirable.
 
+## Why not JWT?
+
+We initially set out to try to use JWT as a format. However, since we are not interested in signing the tokens (which is a fundamental concept of JWT), this lead to problems. Skipping signing means we have to use `algorithm='none'`, which is not very well supported. When using this algorithm, the resulting tokens always end in `.`, as the final part would be based on the key, which is missing. Having a trailing `.` after each token is an unnecessary error source (users may not copy it, ...). We _could_ try to handle this when decoding, but this would still make this technically invalid JWT.
```
markstory, line 277:
```suggestion
We initially set out to try to use JWT as a format. However, since we are not interested in signing the tokens (which is a fundamental concept of JWT), this led to problems. Skipping signing means we have to use `algorithm='none'`, which is not very well supported. When using this algorithm, the resulting tokens always end in `.`, as the final part would be based on the key, which is missing. Having a trailing `.` after each token is an unnecessary error source (users may not copy it, ...). We _could_ try to handle this when decoding, but this would still make this technically invalid JWT. 
```

text/0091-ci-upload-tokens.md
```diff
@@ -115,26 +116,26 @@ unaware of the structure behind structural tokens nothing changes.
...
-  'sentry_url': 'https://sentry.io/',
-  'sentry_region_url': 'https://eu.sentry.io/',
-  'sentry_org': 'myorg'
-}
+```py
+def parse_token(token: str):
+    if not token.startswith("sntrys_") or token.count('_') != 2:
+        return None
+
+    # Note: We add == to the end of the string, because we remove the base64 padding when generating the token
```
mdtro, line 133: Why do we remove the padding when generating the token? Can we just keep the padding in place so that client-side decoding is simple?

mydea: the idea is to have "nicer" tokens without the `==` in them, from a users perspective. I don't have a super strong feeling here, maybe we also leave this out as this is not the end of the token string anymore (as the secret is at the end) 🤔  It doesn't matter that much if there is `==` in the middle of the token I'd say.

mydea: After giving it some more thought, I will adjust this to allow padding for the encoded payload, as it is in the middle of the token and there are also other special characters (e.g. / or -) in the base64 encoded string, it makes it easier to allow this. We just make sure to have no padding at the end of the secret.

PR #106, opened 2023-06-22
Better replacement for Artifact Indexes

Swatinem:
We want to come up with a way to support SourceMap processing or in particular access to files needed for SourceMap
processing that supports more customer use-cases, is correct, and is efficient to implement and scale.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/swatinem/artifact-index/text/0106-artifact-indices.md)

### DIFF
--- a/text/0106-artifact-indices.md
+++ b/text/0106-artifact-indices.md
@@ -68,6 +68,7 @@ Built on top of `ReleaseFile`s, this does not index each individual file, but ra
 - **con**: The index needs to be merged/updated when uploading new bundles.
 - **con**: For disjoint / unique file names, the index will grow very large.
 - **con**: Needs a `Release` object as foreign key.
+- **con**: Can lead to "merge conflicts" with concurrent uploads touching the same index.
 
 ## Artifact Bundle
 
@@ -122,6 +123,7 @@ index, a flat-file format was proposed. However the exact format is rather an in
 - **con**: Needs to re-create / replicate most of the artifact-index logic, and inherits all of its cons:
   - updating / merging the index on uploads
   - the index growing monotonically on disjoint uploads
+- **con**: Automatic deletions (expired files) need to be taken into account when maintaining index.
 
 ## Index individual files in the Database
 
@@ -148,11 +150,12 @@ The table could look something the following:
 The idea was brought up to only index individual files if the number of uploaded bundles per release exceeds a certain
 threshold `N`.
 
-This takes advantage of the fact that the downstream consumer (Symbolicator) will query all the returned `N` bundles
-for the needed files as well.
+This takes advantage of the fact that the downstream consumer (Symbolicator) will use the `manifest.json` index in
+the returned `N` bundles, and no additional server-side index might be needed in that case.
 
 - **mixed**: All the pros/cons of the above _individual files_, just with a lower number of indexed files.
 - **pro**: No additional overhead for the one-bundle-per-release use-case.
+- **pro**: The consumer (Symbolicator) side already has all the logic for dealing with `manifest.json` indices.
 - **con**: Needs jobs and logic to backfill a per-file index if the threshold was exceeded.
 
 ## _Store_ individual files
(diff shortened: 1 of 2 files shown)


text/0106-artifact-indices.md
```diff
@@ -0,0 +1,175 @@
...
+
+## _Store_ individual files
+
+Instead of just _indexing_ files individually and still referring to the bundle they are included in, it might be possible
+to _store_ files themselves individually, possibly deduplicated by a content hash.
+
+- **pro**: Solves file decompression performance of Symbolicator.
+- **pro**: We _assume_ that most files inside of bundles are unused.
+- **pro**: We _assume_ that a lot of files will not change across bundles (polyfills, dependencies, etc).
+- **con**: We need a better way to store many small files.
```
loewenheim, line 166: Better in what way? Is it too expensive, too hard to organize, or something else?

Swatinem: Lets see if @mitsuhiko wants to chime in here, as he was very vocal against doing this. I believe the reason is that depending on storage vendor, you are billed by number of files, which can get extremely expensive storing a ton of small files.

text/0106-artifact-indices.md
```diff
@@ -0,0 +1,175 @@
...
+
+This would involve creating a replacement for artifact indexes based on the newer artifact bundles. Instead of a `JSON`
+index, a flat-file format was proposed. However the exact format is rather an insignificant detail.
+
+- **pro**: Would allow querying / accessing the needed files.
+- **pro**: Would solve one pain-point of artifact indexes: No more `Release` object needed.
+- **con**: Requires downloading / querying the index on API servers.
+- **con**: Needs to re-create / replicate most of the artifact-index logic, and inherits all of its cons:
+  - updating / merging the index on uploads
+  - the index growing monotonically on disjoint uploads
```
iambriccardo, line 124:
```suggestion
- **pro**: Would allow querying / accessing the needed files.
- **pro**: Would solve one pain-point of artifact indexes: No more `Release` object needed.
- **con**: Requires downloading / querying the index on API servers.
- **con**: Needs to re-create / replicate most of the artifact-index logic, and inherits all of its cons:
  - updating / merging the index on uploads
  - the index growing monotonically on disjoint uploads
- **con**: Maintenance of the index can be tedious due to automatic deletions (e.g., what to do if the bundle mentioned in the flat-file is expired?).
```

text/0106-artifact-indices.md
```diff
@@ -0,0 +1,175 @@
...
+- **pro**: No need to download / process any index on API servers.
+- **pro**: Would work well with auto-expiration / partitioning if implemented correctly.
+- **con**: High load on Database and its indices.
+- **con**: Might require a custom solution to better compress `release` and `url`, possibly by using some kind of interning.
+
+The table could look something the following:
+
+- org / project
+- (optional) release + dist
+- url
```
iambriccardo, line 142: Here we need to also consider that we would have to perform the lookup with the `url` variants, thus we would have to use the `IN` operator or a chain of `OR` which means that we will have to leverage at most the `org` and `release/dist` indexes for fast lookup.

text/0106-artifact-indices.md
```diff
@@ -0,0 +1,175 @@
...
+- date_added / touched
+- artifact bundle id
+
+## Hybrid Bundles / Individual files
+
+The idea was brought up to only index individual files if the number of uploaded bundles per release exceeds a certain
+threshold `N`.
```
iambriccardo, line 149:
```suggestion
The idea was brought up to only index individual files if the number of uploaded bundles per release exceeds a certain
threshold `N` (which could easily be `1` since we want to optimize the case in which there is no need to create an index, since we can just leverage the `manifest.json` in the bundle).
```

text/0106-artifact-indices.md
```diff
@@ -0,0 +1,195 @@
...
+
+# Motivation
+
+We have discovered that the transition to non-indexed Artifact Bundles and the moving of SourceMap processing to Symbolicator have caused some customer Stack Traces to regress. We traced these regressions back to our inability to resolve the files needed for SourceMap processing through our current method of querying for these files.
+
+## Customer Use-Cases
+
+Customers may use our product in a multitude of ways. Although we may want to guide and encourage users to use the product in a specific manner, such as _uploading only a single bundle per release_, this may not be practical and it restricts the customers' freedom. Furthermore, it may not align with the direction we wish to pursue.
+
+We spent some time analyzing all the edge cases that our system permits and e
```
liamjones, line 18: This sentence is cut off

iambriccardo: Fixed, thanks!

PR #113, opened 2023-09-06
ref: Amend span creation names for RFC 101

AbhiPrasad:
Discussed offline we decided to change the names of the span creator methods.

- `startActiveSpan` -> `startSpan`
- `startSpan` -> `startInactiveSpan`

lmk what you all think!

### DIFF
--- a/text/0101-revamping-the-sdk-performance-api.md
+++ b/text/0101-revamping-the-sdk-performance-api.md
@@ -18,9 +18,9 @@ Note: In a previous version of this RFC there was a focus on updating the span s
 
 # Planned Work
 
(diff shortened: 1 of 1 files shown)


mydea (approved):
:shipit: sounds good to me!

sentrivana:
I wasn't present during the discussion that led to this change so I don't want to approve on behalf of folks that were but wanted to say this change sounds great! Definitely makes the API more intuitive. :shipit:

Lms24 (approved):
Nice! I'm still not a fan of the "start" naming in the callback-based API (would have preferred something like `trace`) but this is just bike-shedding at this point and I also see the value in keeping this at least _somewhat_ consistent across SDKs.  Thanks for updating!

sl0thentr0py (approved):
merci

PR #115, opened 2023-10-03
Custom Metrics in SDKs

mitsuhiko:
This is a work in progress RFC for how metrics fit into SDKs. I will update motivation and text here shortly.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/metrics/text/xxxx-metrics.md)

### DIFF
--- a/text/xxxx-metrics.md
+++ b/text/xxxx-metrics.md
@@ -155,7 +155,7 @@ metrics.
   the total sum of emitted values, `count` is a running counter of the number of values reported).
 
 * **distribution**(*key*, *value*, *unit* = `"none"`, *tags* = `{}`, *timestamp* = `null`):
-  Records a value for a distribution.  The aggregation state is a lis of all values encountered.
+  Records a value for a distribution.  The aggregation state is a list of all values encountered.
   This means that distributions are reduced to their quantiles only post emission.  This also means
   that in practical terms distributions are likely to be more expensive on the SDK side in terms of
   memory consumption than many other metric types and might cause early flushes.
--- a/text/xxxx-metrics.md
+++ b/text/xxxx-metrics.md
@@ -1,0 +1,318 @@
+- Start Date: YYYY-MM-DD
+- RFC Type: feature
+- RFC PR: <link>
+- RFC Status: draft
+
+# Summary
+
+This RFC lays out the data model behind metrics in Sentry.  Metrics are time series
+that can be pre-aggregated on the SDK side, optionally can be extracted from
+transactions in relay and pop up in the product in different areas.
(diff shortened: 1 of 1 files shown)


text/xxxx-metrics.md
```diff
@@ -0,0 +1,318 @@
...
+  SDK side is a single float.
+
+* **gauge**(*key*, *value* = `1.0`, *unit* = `"none"`, *tags* = `{}`, *timestamp* = `null`):
+  Emits a gauge value.  For more information about gauges see the metric types.  On the SDK side
+  the aggregation state is 5 counters (`last` with the most recently emitted value, `min` with the
+  smallest value observed, `max` with the largest value observed, `sum` which is a float sum of
+  the total sum of emitted values, `count` is a running counter of the number of values reported).
+
+* **distribution**(*key*, *value*, *unit* = `"none"`, *tags* = `{}`, *timestamp* = `null`):
+  Records a value for a distribution.  The aggregation state is a lis of all values encountered.
```
philipphofmann, line 158:
```suggestion
  Records a value for a distribution.  The aggregation state is a list of all values encountered.
```

text/xxxx-metrics.md
```diff
@@ -0,0 +1,318 @@
...
+features.
+
+# SDK Design
+
+SDKs are expected to provide basic metric emission APIs modelled after common APIs inspired
+by statsd.
+
+## Aggregations
+
+SDKs are expected to aggregate locally for windows of up to 10 seconds.  The aggregator shall be
```
philipphofmann, line 126: I guess that mobile and browser SDKs can apply a different strategy cause the metrics frequency will be lower. One strategy could be to attach these buckets to an already scheduled envelope, which we do for [client reports](https://develop.sentry.dev/sdk/client-reports/).

text/xxxx-metrics.md
```diff
@@ -0,0 +1,318 @@
...
+running in the background and regularly flush out items.  Upon shutdown the SDK needs to ensure
+that unflushed buckets are drained.  SDKs area also encouraged to ensure that the total aggregation
```
philipphofmann, line 128: For Mobile, do we need to ensure that we do not lose any buckets when crashing, or is losing some buckets acceptable? I would say losing buckets is unacceptable. One way to achieve this is to sync the buckets across multiple layers, as we already do for the scope.

philipphofmann (commented):
Exciting RFC 😃

philipphofmann:
We can close this as we haven't moved forward with metrics in the SDKs. Please reopen if you disagree, @mitsuhiko.

PR #116, opened 2023-10-12
rfc(feature): Sentry Semantic Conventions

AbhiPrasad:
closes https://github.com/getsentry/team-webplatform-meta/issues/107

The purpose of this RFC is to formalize the a set of semantic conventions in Sentry, aligning with [OpenTelemetry's semantic conventions](https://opentelemetry.io/docs/concepts/semantic-conventions/). These will be a standardized naming scheme for operations and data that will be shared across the SDKs, ingest, and the product.

This moves us closer to OpenTelemetry, helps reduce friction when creating new product features that rely on conventions around data, and helps us avoid the need to create new conventions for data that is already covered by OpenTelemetry.

At the current time, these semantic conventions are meant to apply to spans, breadcrumbs, and metrics, but not to errors/transactions/replays/crons. This may change in the future.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/sentry-semantic-conventions/text/0116-sentry-semantic-conventions.md)


text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+Right now we have a variety of top level fields that need to be mapped to attributes, and some of them match existing OpenTelemetry semantic conventions. We should map these fields to attributes as follows (the dots in the initial fields represent nested fields):
+
+Top level fields:
+
+- `release` -> `sentry.release`
+- `environment` -> `sentry.environment`
+
+[User interface](https://develop.sentry.dev/sdk/event-payloads/user/):
+
+- `user.id` -> `sentry.user.id`
```
mydea, line 124: can we use `enduser.id` here instead? See https://opentelemetry.io/docs/specs/otel/trace/semantic_conventions/span-general/#general-identity-attributes, this was also requested here: https://github.com/getsentry/sentry-javascript/issues/9177

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Attributes Schema Versioning and Implementation
+
+The attributes schema will be versioned and published as a JSON Schema, which will be the representation of the well-known attributes that Sentry ingestion and product relies on.
+
+The JSON schema will also be codegened into packages for JavaScript, Python, and Rust, so that the schema can be used in Relay, the frontend and the backend. There has to be care put in to make sure that the codegened packages are compatible with the JSON schema, and that the JSON schema is backwards compatible, but we can introduce tests to make sure that this is the case.
+
+By having a versionig/packaging structure like this, it also makes it much easier for the SDKs to identify what parts of the product are using the data they are sending.
+
+## Mapping Sentry Specific Fields to Attributes
```
mydea, line 113: So some things that are missing here, at least things I needed to manually handle in the POTEL experiments, are:

* span origin (`sentry.origin` ?)
* span op (`sentry.op` ?)
* span source (`sentry.source` ?)
* sample rate (`sentry.sample_rate` ?)

we also kind of have a parent sampled flag, but this may also just remain an internal thing I guess.

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+
+[SDK interface](https://develop.sentry.dev/sdk/event-payloads/sdk/)
+
+- `sdk.name` -> `sentry.sdk.name`
+- `sdk.version` -> `sentry.sdk.version`
+- `sdk.integrations` -> `sentry.sdk.integrations`
+- `sdk.packages` -> `sentry.sdk.packages`
+
+For the contexts defined in as part of our contexts payload, we can flatten it so that the context name become the prefix to the attribute key. For example the `os` context would become `os.name`, `os.version`, `os.kernel_version`, etc. Some of these new keys will have to be renamed to better match OpenTelemetry's semantic conventions, but that can be approached in a case by case basis.
+
```
mydea, line 150: One other thing that we may have to handle, but maybe also want to do in a follow up, are breadcrumbs? The timed events in OTEL also use the same attributes schema. In the POTEL experiment I have:

* `sentry.breadcrumb.type`
* `sentry.breadcrumb.level`
* `sentry.breadcrumb.event_id`
* `sentry.breadcrumb.category`
* `sentry.breadcrumb.data` --> this is a bit tricky as it is arbitrary JSON, currently I store this as serialized JSON string.

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+- `user.geo.country_code` -> `sentry.user.geo.country_code`
+- `user.geo.region` -> `sentry.user.geo.region`
+
+[Request interface](https://develop.sentry.dev/sdk/event-payloads/request/)
+
+- `request.method` -> [`http.request.method`](https://github.com/open-telemetry/semantic-conventions/blob/cadfe53949266d33476b15ca52c92f682600a29c/model/trace/http.yaml#L19)
+- `request.url` -> [`url.full`](https://github.com/open-telemetry/semantic-conventions/blob/cadfe53949266d33476b15ca52c92f682600a29c/model/trace/http.yaml#L47)
+- `request.query_string` -> [`url.query_string`](https://github.com/open-telemetry/semantic-conventions/blob/cadfe53949266d33476b15ca52c92f682600a29c/model/trace/http.yaml#L47)
+- `request.cookies` -> `http.request.cookies`
+- `request.cookies` -> `http.request.headers`
```
mydea, line 139:
```suggestion
- `request.headers` -> `http.request.headers`
```

Is the idea of cookies / headers to store a JSON string or to store them in nested dot notation, e.g.

```js
span.setAttribute('request.headers.content-type', 'application/json');
```

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+
+For the purposes of rollout, we recommend this field is adopted by spans, metrics (DDM), and breadcrumbs first, and then adopted by crons after it has been further validated given there is no arbitrary context on crons at the moment.
+
+This does not replace tags, which should still be part of their respective signal payloads. The product can make decisions on how to promote different attributes to become tags. This happens with contexts and span data today.
+
+## Attributes Schema
+
+Attribute keys should be unique and well-known, and should not be used for multiple purposes. In OpenTelemetry the attribute values cannot be dictionaries, only primitives and arrays of primitives. We will try to follow this convention as well, but we will allow dictionaries as attribute values for better backwards compatibility with existing Sentry fields.
```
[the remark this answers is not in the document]
AbhiPrasad: Yes in the otel schema they have to be flattened. I think we should be more liberal than that though to start off, we can eventually get to a flattened state.

text/0116-sentry-semantic-conventions.md
AbhiPrasad: Yeah I'm not sure this will ever get to an exhaustive state, more just important examples, but I think we need a good analog for trace context. Will add that.

text/0116-sentry-semantic-conventions.md
AbhiPrasad: Yes, will update, thank you for the reminder.

text/0116-sentry-semantic-conventions.md
AbhiPrasad: I'm still undecided if we should do 

> we will allow dictionaries as attribute values for better backwards compatibility with existing Sentry fields

The idea here was to use a JSON string because it is the easiest to adapt from what we have, but maybe we should strictly enforce the otel attributes schema of no nesting. Still going back and forth on it.

text/0116-sentry-semantic-conventions.md
AbhiPrasad: Ideally breadcrumbs data matches the attributes, we can structure them the same way otel does it's logs signal.

for span events that are attached that map to breadcrumbs, we probably want to "unroll" breadcrumb data into the top level keys for attributes on span events.

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+
+For the contexts defined in as part of our contexts payload, we can flatten it so that the context name become the prefix to the attribute key. For example the `os` context would become `os.name`, `os.version`, `os.kernel_version`, etc. Some of these new keys will have to be renamed to better match OpenTelemetry's semantic conventions, but that can be approached in a case by case basis.
+
+## Review process
+
+Generally as per OpenTelemetry's semantic conventions, we can split up the attributes into required and optional attributes into various categories (browser, mobile, http, etc). This means we can have each category have it's own set of owners that can review and approve changes to the attributes in that category. This will help us scale the review process for the attributes.
+
+## Backwards Compatibility
+
+We will support the `data` field as an alias for `attributes` for backwards compatibility. This means that breadcrumbs and spans can adopt the `data` field immediately. There is no timeline for us to remove the `data` field, but new signals like crons or metrics should use `attributes` as the field name.
```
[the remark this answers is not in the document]
AbhiPrasad: If an SDK sends stuff in the data field we should be good to go for backwards compat in breadcrumbs/spans.

I think we're not going to make these changes to errors (too much friction to change) or transactions (getting phased out by spans anyway).

text/0116-sentry-semantic-conventions.md
mydea: yeah, these were just the things I noticed we need to kind of maintain feature parity - esp. origin, op and source I believe? sample rate is maybe less important I guess.

text/0116-sentry-semantic-conventions.md
mydea: to clarify, I do store these as regular attributes on a timed event :) But we still need universally understood attributes on there, I guess? timed events have only a name other than the attributes. Not sure what we do with these fields on the breadcrumb right now though, tbh.

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * null or undefined attribute values are invalid and will result in undefined behavior.
+ */
+export type AttributeValue =
+  | string
+  | number
+  | boolean
+  | Array<string>
+  | Array<number>
+  | Array<boolean>
```
lforst, line 85: This will mean no mixed arrays? eg `['foo', 42, true]`?

It's probably fine but just raising to be sure somebody thought about this.

text/0116-sentry-semantic-conventions.md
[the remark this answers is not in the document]
lforst: I am also wondering how to represent stack traces in this format. It's an important use case we cannot dismiss.

It's not just stacktraces. `debug_meta` for example too.

Ideally we have some sort of plan of attack or migration path lined out in this doc for fields like these.

text/0116-sentry-semantic-conventions.md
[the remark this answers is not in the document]
lforst: I thought this was about errors too. At least we mention errors in the intro of this RFC. Are errors in scope or not? Maybe we should clarify this?

text/0116-sentry-semantic-conventions.md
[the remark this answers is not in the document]
mydea: FWIW, OTEL has the built in type `EXCEPTION_STACKTRACE` which is just a string.

AbhiPrasad:
I spent a bunch of time mulling over various decisions, but I've decided on a couple things

1. The convention attributes can no longer be nested - they must be flat top level on the attributes object. We will provide mappings for all possible values in the event schema to address this
2. The conventions are only applying to spans/breadcrumbs/metrics. Crons/errors/replays/transactions are not addressed in this case, but those signals can choose to adopt the conventions if they wish. This is to assist with ease of adoption and reduce hard to deal with changes
3.  More explicitly wrote down mappings for op, sample_rate and other sentry specific performance monitoring values

text/0116-sentry-semantic-conventions.md
```diff
@@ -0,0 +1,164 @@
+- Start Date: 2023-10-12
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/116
+- RFC Status: draft
```
AbhiPrasad, line 4:
```suggestion
- RFC Status: implementation
```

AbhiPrasad:
I'm moving this RFC into implementation state. We're going to work on creating a repo to store our conventions, and publish libraries that both sentry and relay can use.

PR #119, opened 2023-10-25
New strategy to use Rust in Sentry/Python

Swatinem:
We want to improve the adoption of Rust within the Sentry/Python codebase, making it easier to do so,
and removing some hurdles that make current Rust usage very annoying and brittle.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/swatinem/rust-python/text/0119-rust-in-sentry.md)

### DIFF
--- a/text/0119-rust-in-sentry.md
+++ b/text/0119-rust-in-sentry.md
@@ -1,6 +1,6 @@
 - Start Date: 2023-10-25
 - RFC Type: decision
-- RFC PR: <link>
+- RFC PR: https://github.com/getsentry/rfcs/pull/119
 - RFC Status: draft
 
 # Summary
--- a/text/0119-rust-in-sentry.md
+++ b/text/0119-rust-in-sentry.md
@@ -58,8 +58,28 @@ Rust crate.
 - **pro**: Simpler release / update flow. Ideally things would only require a single commit to `sentry`.
 - **con**: Setting up and maintaining a workflow in the Monolith repo that suits every developer might be more complex.
 
+# Proposed Workflow
+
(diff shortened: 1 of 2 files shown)


text/0119-rust-in-sentry.md
```diff
@@ -0,0 +1,85 @@
...
+With this approach, each Rust crate we are interested in would have its own Python Bindings.
+
+- **pro**: We have smaller more targeted crates/packages to maintain and to publish.
+- **pro/con**: Both pro/con is that each crate/package is in charge of its own workflows.
+- **con**: This would still suffer from the problems around having to maintain a public SemVer API.
+- **con**: It would also rely on publishing to PyPI, and then updating that version within Sentry.
+- **con**: Having multiple smaller Python extension modules will increased the fixed per-module overhead.
+  In particular, each module would have its own bundled copy of the Rust `std`.
+
+## Create a single Sentry-specific bindings package
```
untitaker, line 47: i'm in favor of a "other" bindings package for functionality that does not fit into any of the more focussed packages, especially if we are already adding this functionality into packages it doesn't belong into.

I think we would want to keep the arroyo bindings small/focussed because they are potentially useful outside of sentry, therefore I wouldn't want to couple them and their build process to a bunch of native stuff.

my concern with a large bindings crate in general is not even the python wheel size, but having a complicated build process. i think having a mega-package that links rdkafka/breakpad/openssl/... sounds like a recipe for slow CI and hard to debug failures.

loewenheim (approved):

Swatinem:
This is now alive 🎉

PR #120, opened 2023-11-06
DRAFT: Introduce global & request scopes for JavaScript SDK

mydea:
**WORK IN PROGRESS**

This is an RFC to rework hubs & scopes for the JavaScript SDK. Subject to change and very much WIP!

[Rendered RFC](https://github.com/getsentry/rfcs/blob/fn/js-scope/text/0120-js-sdk-scope.md)

Note: We decided to split this up into two RFCs. [One](https://github.com/getsentry/rfcs/pull/122) for the general Hub/Scope/Client changes, and this one will be more specifically tailored to JS and it's specific needs (which may or may not apply to other SDKs/languages as well, and may or may not be used by SDKs going forward).

### DIFF
--- a/text/0120-js-sdk-scope.md
+++ b/text/0120-js-sdk-scope.md
@@ -5,16 +5,18 @@
 
 # Summary
 
-We want to streamline Hub & Scope usage for the JavaScript SDK and beyond, to also better align with OpenTelemetry models etc.
+This RFC proposes changes on top of [RFC #122](https://github.com/getsentry/rfcs/pull/122). 
+While that RFC proposes general changes to merge Hub & Scope, this RFC goes further and proposes a way to ensure we can get the correct breadcrumb & context data for any events, specifically in JavaScript but potentially also in other SDKs.
+
+Things proposed in this RFC _do not_ have to be followed/implemented by all languages/SDKs, but any language/SDK may pick things from it that are relevant.
 
 # Motivation
 
-Generally, Hubs and Scopes are a very Sentry-specific concept, and can be a bit hard to explain. 
-Also, how Hub & Scope forking behaves is currently a bit tricky, and does not play well e.g. with how OpenTelemetry contexts work.
+Especially when working with OpenTelemetry, the way how we currently always assign data to the current scope can break down in unexpected ways.
 
-In OpenTelemetry, contexts are forked very often, because Contexts are copy-on-write - each time you want to set _anything_ on a context, you need to fork it.
-This frequent forking leads to a lot of problems with how sentry scopes apply. 
-Because any function or callback may have a forked context, attaching data inside of e.g. such a callback would lead to this never being applied to events correctly:
+Because OpenTelemetry has to fork a context each time _anything_ is changed on it (copy-on-write), you end up with a lot of deeply nested scopes.
+
+This leads to many scenarios where data is added to the "wrong" scope, from a users perspective:
 
 ```js
 app.get('/my-route', function() {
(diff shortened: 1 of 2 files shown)


text/0120-js-sdk-scope.md
```diff
@@ -0,0 +1,187 @@
...
+Note that the global scope _is not_ the initial scope, but a separate, "empty" scope. This way, you cannot accidentally pollute it. 
+You can access it via `getGlobalScope()`:
+
+```js
+Sentry.init();
+
+const scope = Sentry.getScope(); // <-- the initial scope, that other scopes will inherit from
+const globalScope = Sentry.getGlobalScope(); // <-- the global scope that will be applied to all events
+
+scope !== globalScope
```
krystofwoldrich, line 80: Does `Sentry.init` create both the `scope` and `globalScope` and set the same client on both of them or only the `globalScope` is tied to the created client and `scope` has `NoopClient`?

Do we always have these two scopes? What would be the difference for envs like RN, where we have currently only one Hub and Scope?

text/0120-js-sdk-scope.md
```diff
@@ -0,0 +1,187 @@
...
+## Breadcrumbs Scope
+
+If you need to manually isolate breadcrumbs, e.g. for a queue function, you can do this with the following new function:
+
+```js
+function myQueueJob() {
+  Sentry.withScope((scope) => {
+    scope.captureBreadcrumbsHere();
+    // do something
+  });
```
krystofwoldrich, line 134: I'm not sure if I get the Breadcrumbs Scope isn't it just a copy of the initial Scope (in this example.)?

krystofwoldrich: Or does this mean that the breadcrumb will be added to the `breadcrumbs scope` of the parent/global?

text/0120-js-sdk-scope.md
mydea: This is a bit of an implementation detail, I think I would implement it in a way that the global scope is created when it is first fetched. E.g. something like this:

```js
function getGlobalScope() {
  const client = getClient();
  const globalScope = getGlobalScopeForClient(client); // this could either live on the client, or e.g. in a weakmap
  if (globalScope) {
    return globalScope;
  }
  return createGlobalScopeForClient(client);
}
```

Does that make sense to you? 

I think it may also be reasonable to _not_ have this in RN if we only have a single hub/scope. (Just to clarify, does that mean in RN we do not have `withScope()` or `Scope.clone()`? )

text/0120-js-sdk-scope.md
mydea: The big difference is that the breadcrumbs scope is applied with the breadcrumbs it has at event creation time, not at the time when the scope is forked. Think of this slightly contrived example, but it should explain the problem:

```js
Sentry.withScope(scope => {
  Sentry.addBreadcrumb({ message: 'breadcrumb 1' }); 
  // scope.breadcrumbs: ['breadcrumb1']

  Sentry.withScope(scope2 => {
    // scope2.breadcrumbs: ['breadcrumb1']
    Sentry.addBreadcrumb({ message: 'breadcrumb2' }); 
    // scope2.breadcrumbs: ['breadcrumb1', 'breadcrumb2']
  });

  // scope.breadcrumbs: ['breadcrumb1']

  Sentry.withScope(scope3 => {
    // scope3.breadcrumbs: ['breadcrumb1']
    Sentry.captureException(error);
  });
});
```

While this appears a bit contrived, in an OTEL world scopes are automatically forked all the time, so you get into this scenario quite fast there. With this proposal, if you define a breadcrumbs scope, instead you'd have this:

```js
Sentry.withScope(scope => {
  scope.captureBreadcrumbsHere();
  Sentry.addBreadcrumb({ message: 'breadcrumb 1' }); 
  // scope.breadcrumbs: []
  // scope.breadcrumbsScope.breadcrumbs: ['breadcrumb1']

  Sentry.withScope(scope2 => {
    // scope2.breadcrumbs: []
    // scope.breadcrumbsScope.breadcrumbs: ['breadcrumb1']
    Sentry.addBreadcrumb({ message: 'breadcrumb2' }); 
    // scope2.breadcrumbs: []
    // scope2.breadcrumbsScope.breadcrumbs: ['breadcrumb1', 'breadcrumb2']
  });

  Sentry.withScope(scope3 => {
    // scope3.breadcrumbs: []
    // scope3.breadcrumbsScope.breadcrumbs: ['breadcrumb1', 'breadcrumb2']
    Sentry.captureException(error);
  });
});
```

Does that make sense? :D

text/0120-js-sdk-scope.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+# Drawbacks
+
+* TODO
+
+# Unresolved questions
+
+* Do we prefer `getGlobalScope()` or `getClientScope()`?
+* What actual API should we use for the scope breadcrumbs? `scope.captureBreadcrumbsHere()` is just an initial draft...
```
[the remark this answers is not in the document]
mydea: this is just about the API to define that you want to have an isolated breadcrumbs scope there. You'd always add breadcrumbs via `Sentry.addBreadcrumbs()`, like you do today - this API will just tell `addBreadcrumbs()` under the hood where to put the breadcrumb!

text/0120-js-sdk-scope.md
mydea: Also, in RN you maybe don't need this at all. For Browser JS, most probably just having all breadcrumbs be on the global scope always may be enough. (TBD)

text/0120-js-sdk-scope.md
krystofwoldrich: RN has `withScope()` or `Scope.clone()` but it's basically not used. One exception is cloning the scope for specific event -> `captureException(error, withScope)`.

krystofwoldrich: From https://github.com/getsentry/rfcs/pull/122 I understand that scopes hold reference to clients, 1:1.

Will that be the same with the global scope or it will be available to all created clients?

text/0120-js-sdk-scope.md
krystofwoldrich: Thanks for the explanation. That makes sense. Yeah for RN there won't be much difference from holding the breadcrumbs on the global scope, but I understand the difference. 

The `captureBreadcrumbsHere` enables us to select the scope where from which the `breadcrumbs` will be shared.

cleptric:
Superseded by https://github.com/getsentry/rfcs/pull/122

PR #121, opened 2023-11-07
rfc(decision): DRAFT: Environment Variable conventions

cleptric:


Lms24 (commented):
General:

I think this makes sense and at least partially enables easier configuration of the SDK. 

I have an important question though: Is the goal of supporting env variables to read them at SDK init time and store the values as long as the application is alive? Or should SDKs also observe updates to env variables after they're initialized?

* If only at init time, all good, we can make this work fairly easily.
* In case of the latter, we need to be aware though that this has deeper implications on how SDKs currently deal with options. For instance, we currently don't guarantee support for "updating" of `getClient().getOptions()`. It's possible (at least in JS) but very likely, we have some instrumentation (e.g. an integration) that hard-copied e.g. sample rates when they initialized, stored them away and wouldn't receive the update value. If we were to support dynamic updates of options, we need to ensure that this doesn't happen anywhere. IOW, we'll need to re-query the env variables a lot. 

Also, users (and we) need to be aware that setting options via env variables will only work for primitive values. Adding integrations or callbacks (beforeSend) like this will not work.

JS Specific and also dependent on my initial question:

1. Since Browser and Server SDKs inherit from a core SDK package, we need to find a way to a) only add env variable reads in the server side or b) make these reads a no-op in the browser
2. Both but especially b) will have a bundle size impact. Negligible in server-land, something we need to keep an eye on in browser-land.

cleptric:
Just for init, everything else, as you rightly pointed out, has too many side effects.

Lms24 (approved):
In this case, let's do it!

mwarkentin:
Sorry, tapped something while reviewing that marked it as ready and I don't see a way to unmark it on GH mobile now.

text/0121-draft-environment-variable-conventions.md
```diff
@@ -0,0 +1,45 @@
...
+
+A first list of environment variables that all server sides SDKs should support could look like:
+
+- `SENTRY_DSN`
+- `SENTRY_RELEASE`
+- `SENTRY_ENVIRONMENT`
+- `SENTRY_SAMPLE_RATE`
+- `SENTRY_TRACES_SAMPLE_RATE`
+- `SENTRY_PROFILES_SAMPLE_RATE`
+
```
antonpirker, line 33: I would also add `SENTRY_DEBUG` for easier turning debug output on/off.

antonpirker (approved):
looks good

text/0121-draft-environment-variable-conventions.md
```diff
@@ -0,0 +1,46 @@
...
+A first list of environment variables that all server sides SDKs should support could look like:
+
+- `SENTRY_DSN`
+- `SENTRY_RELEASE`
+- `SENTRY_ENVIRONMENT`
+- `SENTRY_SAMPLE_RATE`
+- `SENTRY_TRACES_SAMPLE_RATE`
+- `SENTRY_PROFILES_SAMPLE_RATE`
+- `SENTRY_DEBUG`
```
lforst, line 33: Just stumbled across this RFC. We could also list `SENTRY_TRACE` and `SENTRY_BAGGAGE`, which could be picked up to continue traces.

cleptric: Yep, there is/was another RFC for that, but I can add it and actually move this RFC forward :grimacing:

JPeer264:
https://github.com/getsentry/sentry-docs/pull/16326 has been merged. I guess this RFC can be closed (I'll close it, feel free to reopen)

PR #122, opened 2023-11-08
DRAFT: Merge Hub & Scope in SDKs

mydea:
**WORK IN PROGRESS**

This is a WIP RFC about merging the concepts of hubs & scopes in the SDK into a singular Scope concept.

[Rendered RFC](todo)

### DIFF
--- a/text/XXXX-sdk-hub-scope-merge.md
+++ b/text/XXXX-sdk-hub-scope-merge.md
@@ -1,0 +1,148 @@
+- Start Date: 2023-11-07
+- RFC Type: feature
+- RFC PR: TODO
+- RFC Status: draft
+
+# Summary
+
+We want to streamline Hub & Scope usage for the SDK.
+
+# Motivation
+
+Generally, Hubs and Scopes are a very Sentry-specific concept, and can be a bit hard to explain. 
+Also, how Hub & Scope forking behaves is currently a bit tricky, and does not play well e.g. with how OpenTelemetry contexts work.
+
+This RFC aims to streamline this by merging the concepts of Hub and Scope into a singular Scope concept.
+
+# Background
+
+TODO
+
+# Proposed Solution
+
+Hubs will be removed as a concept for SDK users. 
+Instead, from a users perspective only Scopes remain, which will become the singular entity to hold context data etc.
+
+Scopes will be _similar_ to how they work today, but not entirely the same. 
+Scopes can have data (e.g. tags, user, ...) added to them the same way you can do today. 
+This RFC _does not_ aim to change any of the data that is kept on the scope and is applied to events.
+
+The following APIs will be removed:
+
+* `getCurrentHub()`
+* `configureScope()` (instead just get the scope and set on it directly)
+* Any APIs currently on the Hub only: 
+  * `hub.pushScope()`
+  * `hub.popScope()`
+  * `hub.isOlderThan()`
+  * `hub.bindClient()`
+  * `hub.getStack()`
+  * `hub.getStackTop()`
+  * `hub.run()` (use `withScope()` instead)
+
+Instead, we will introduce some new APIs:
+
+```ts
+// get the currently active scope. replacement for `getCurrentHub().getScope()`
+export function getScope(): Scope;
+
+// get the currently active client. May return a NOOP client. Replacement for `getCurrentHub().getClient()`.
+export function getClient(): Client;
+
+// make a scope the current scope. Replacement for `makeMain(hub)`
+export function makeCurrentScope(scope: Scope): void;
+```
+
+The following APIs already exist but will behave differently:
+
+* `withScope()` will still work, but it will actually fork an execution context. So this will roughyl the same as doing `hub.run()` today.
+
+APIs that are currently on the hub should instead be called directly on the scope (e.g. `scope.captureException()` etc.), or via a global method (e.g. `Sentry.captureException()`).
+
+## Clients
+
+Instead of a client being optional, there will now _always_ be a client. It may be a Noop Client that does nothing, if `init()` has not been called yet.
+
+Each client gets a fresh initial scope when it is initialized:
+
+```js
+const client = new Client();
+client.getScope(); // <-- fresh, initial scope for this client
+scope.getClient(); // <-- get client of scope
+```
+
+You can make a scope the current one like this:
+
+```js
+makeCurrentScope(client.getScope());
+// this scope/client will now be available via e.g. `getScope()` or `getClient()`
+// this is what init() calls under the hood
+```
+
+The current scope may be kept similar to how we currently keep the current hub, but this is SDK specific and not part of this RFC.
+
+When calling `getScope()` before a scope was made the current one (=before init was called), we will return a scope for a noop client. 
+A noop client is a regular client that simply does not send anything.
+
+This way, the API for `getClient()` can always return a client, and users do not have to guard against this being undefined all the time. 
+We may also expose a util like `sentryIsInitialized()` that checks if the current client is a Noop client (which currently you could have checked as `getCurrentHub().getClient() === undefined`).
+
+If you want to have multiple isolated clients, you can achieve this easily with this new setup:
+
+```js
+const client1 = new Client();
+const client2 = new Client();
+
+const scope1 = client1.getScope();
+const scope2 = client2.getScope();
+
+scope1.captureException(); // <-- isolated from scope2
+```
+
+## Scopes
+
+Scopes behave similar to how they behave today. 
+When a scope is forked via `withScope()`, a new scope is created that inherits all data currently set on the parent scope.
+
+The main change to Scopes is that they do not push/pop anymore, but instead fork an execution context (in languages where this makes sense/is possible).
+Basically, `withScope()` should behave like `hub.run()` does today.
+
+`client.getScope()` should return the current scope of this client in the current execution context.
+
+From a users perspective, this should mostly not be noticeable - they can always run `getScope()` to get the current scope, or `withScope(callback)` to fork a new scope off the current scope.
+
+You can make a scope the current one via `makeCurrentScope(scope)`, which should bind the scope to the current execution context (or a global, in SDKs without execution context).
+
+## What about other Hub references?
+
+While the Hub is mainly exposed via `getCurrentHub()`, it is also used as argument or similar in many places. 
+These occurences should be updated to instead take a scope or a client.
+
+## What about backwards compatibility?
+
+We should strive to provide a wrapper/proxy `getCurrentHub()` method that still exposes the key functionality to ease upgrading. E.g.:
+
+```js
+import {getScope, getClient} from '../internals';
+
+function getCurrentHub() {
+  return {
+    getClient,
+    getScope
+  }
+}
+```
+
+We need to decide what to keep in that proxy and what not.
+
+## What about globals?
+
+This RFC does not propose any concrete way to store the current scope. This is up to the concrete SDK and may behave the same way as it currently does for the hub, or differently if that makes more sense in a given scenario.
+
+# Drawbacks
+
+* This changes _a lot_ of public APIs and behavior.
+
+# Unresolved questions
+
+TODO


text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+## Clients
+
+Instead of a client being optional, there will now _always_ be a client. It may be a Noop Client that does nothing, if `init()` has not been called yet.
+
+Each client gets a fresh initial scope when it is initialized:
+
+```js
+const client = new Client();
+client.getScope(); // <-- fresh, initial scope for this client
+scope.getClient(); // <-- get client of scope
```
cleptric, line 71: Aren't we introducing reference cycles with this?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+
+You can make a scope the current one like this:
+
+```js
+makeCurrentScope(client.getScope());
+// this scope/client will now be available via e.g. `getScope()` or `getClient()`
+// this is what init() calls under the hood
+```
+
+The current scope may be kept similar to how we currently keep the current hub, but this is SDK specific and not part of this RFC.
```
cleptric, line 82: This should be higher up, as 

```ts
// get the currently active scope. replacement for `getCurrentHub().getScope()`
export function getScope(): Scope;
```

already touches on this subject.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+
+```js
+const client1 = new Client();
+const client2 = new Client();
+
+const scope1 = client1.getScope();
+const scope2 = client2.getScope();
+
+scope1.captureException(); // <-- isolated from scope2
+```
```
cleptric, line 100: 👍

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+* `getCurrentHub()`
+* `configureScope()` (instead just get the scope and set on it directly)
+* Any APIs currently on the Hub only: 
+  * `hub.pushScope()`
+  * `hub.popScope()`
+  * `hub.isOlderThan()`
+  * `hub.bindClient()`
+  * `hub.getStack()`
+  * `hub.getStackTop()`
+  * `hub.run()` (use `withScope()` instead)
```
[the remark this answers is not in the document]
mydea: but is this _only_ available on the hub? (In JS you can also do `Sentry.startTransaction()`). I can also take this into this list explicitly 👍

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+// get the currently active client. May return a NOOP client. Replacement for `getCurrentHub().getClient()`.
+export function getClient(): Client;
+
+// make a scope the current scope. Replacement for `makeMain(hub)`
+export function makeCurrentScope(scope: Scope): void;
+```
+
+The following APIs already exist but will behave differently:
+
+* `withScope()` will still work, but it will actually fork an execution context. So this will roughyl the same as doing `hub.run()` today.
```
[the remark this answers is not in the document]
mydea: ah, good point - wasn't 100% sure if we have this everywhere. This forks a hub for a new execution context. Do we have similar/other APIs for this in other languages?

text/0122-sdk-hub-scope-merge.md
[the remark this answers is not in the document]
cleptric: It's also exported as a global API.

text/0122-sdk-hub-scope-merge.md
mydea: Yes, we are, I am not 100% sure if this is a problem 🤔 I _think_ at least in JS this should be fine as they should be garbage collectable if they are the only things referencing themselves (* I think). If this is problematic we can also leave the `scope.getClient()` API away I'd say - wdyt?

text/0122-sdk-hub-scope-merge.md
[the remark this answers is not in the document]
cleptric: https://github.com/getsentry/sentry-go/blob/master/hub.go#L118C1-L126

text/0122-sdk-hub-scope-merge.md
cleptric: I would leave this out of the scope of the RFC, and each platform can add it if they see the need.

text/0122-sdk-hub-scope-merge.md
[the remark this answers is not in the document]
mydea: Ah, right. I think there should/can still be an API to just clone a scope without any execution context stuff (e.g. in JS you can do `Scope.clone(oldScope)`, which is fine I'd say. Should I include this in the RFC, wdyt?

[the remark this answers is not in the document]
cleptric: Let's do it!

text/0122-sdk-hub-scope-merge.md
markushi: What's the reason for adding these APIs? As of now [clients are state-less](https://develop.sentry.dev/sdk/unified-api/#client), so IMHO it would make more sense to move the ownership into the scope instead. E.g. `Sentry.getScope().getClient()` and `Sentry.getScope().setClient()`.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+
+```ts
+// get the currently active scope. replacement for `getCurrentHub().getScope()`
+export function getScope(): Scope;
+
+// get the currently active client. May return a NOOP client. Replacement for `getCurrentHub().getClient()`.
+export function getClient(): Client;
+
+// make a scope the current scope. Replacement for `makeMain(hub)`
+export function makeCurrentScope(scope: Scope): void;
```
markushi, line 53: Just for reference: This would be called `setCurrentScope` in Java world,  similar to the [`Sentry.setCurrentHub`](https://github.com/getsentry/sentry-java/blob/main/sentry/src/main/java/io/sentry/Sentry.java#L83C4-L83C4) we have now.

text/0122-sdk-hub-scope-merge.md
markushi, line 58:
```suggestion
* `withScope()` will still work, but it will actually fork an execution context. So this will roughly do the same as doing `hub.run()` today.
```

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+
+Scopes behave similar to how they behave today. 
+When a scope is forked via `withScope()`, a new scope is created that inherits all data currently set on the parent scope.
+
+The main change to Scopes is that they do not push/pop anymore, but instead fork an execution context (in languages where this makes sense/is possible).
+Basically, `withScope()` should behave like `hub.run()` does today.
+
+`client.getScope()` should return the current scope of this client in the current execution context.
+
+From a users perspective, this should mostly not be noticeable - they can always run `getScope()` to get the current scope, or `withScope(callback)` to fork a new scope off the current scope.
```
markushi, line 112: This has been a major pain point in the past: `withScope()` was used in places where we actually wanted to change the current scope, but instead created a fork and then people were wondering why the changes were not visible on the original scope. Not sure if feasible within this RFC, but renaming `withScope()` to `forkScope()` would help here.

markushi (commented):
Nice RFC! I left a few comments, looking good overall! My only concern is about the ownership of scope / clients, see my comments.

text/0122-sdk-hub-scope-merge.md
mydea: The idea was to make it easier to create (multiple) clients, but no strong feelings, we can also leave this on the scope only and require to do e.g. `scope.setClient(client)` or something like this. Then you'd do:

```js
const client1 = new Client();
const client2 = new Client();

const scope1 = new Scope();
const scope2 = new Scope();
scope1.setClient(client1);
scope2.setClient(client2);

scope1.captureException(...);
```

Which I guess is fine too!

text/0122-sdk-hub-scope-merge.md
mydea: maybe this is nicer anyhow - actually the `make` prefix is not great anyhow IMHO, so I'll just update this to propose `setCurrentScope` generally.

text/0122-sdk-hub-scope-merge.md
mydea: I have no strong feelings about `withScope`, interested to hear what others think on this! I'd be fine with either.

mydea:
Updated this based on feedback in regard to scope<>client relationship.

Now, a scope holds a reference to a client, which can be set via `scope.setClient()`. The client is inherited by child scopes normally. `scope.getClient()` returns a Noop client if none has been set yet (=init has not been called yet).

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+* `withScope()` will still work, but it will actually fork an execution context. So this will roughly do the same as doing `hub.run()` today in languages that have that, which forks an execution context. 
+
+APIs that are currently on the hub should instead be called directly on the scope (e.g. `scope.captureException()` etc.), or via a global method (e.g. `Sentry.captureException()`).
+
+## Clients
+
+Instead of a client being optional, there will now _always_ be a client. It may be a Noop Client that does nothing, if `init()` has not been called yet.
+
+A scope has a reference to a client. By default it will reference a noop client. You can bind a client to a scope via `scope.setClient()`. 
+The client is inherited by forked scopes.
```
adinauer, line 67: How does this behave if the client is replaced on the parent of a forked scope. Does the client on the forked scope inherit the new client or is the one from back when the fork happened still in use?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+
+```js
+const client1 = new Client();
+const scope = new Scope();
+
+scope.getClient(); // <-- returns a noop client by default
+
+scope.setClient(client1);
```
adinauer, line 75: Is this going to become a thing for normal users of SDKs or just some edge use case for users who e.g. want more than a single client?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+scope.getClient(); // <-- returns client1
+```
+
+The current scope may be kept similar to how we currently keep the current hub, but this is SDK specific and not part of this RFC.
+
+When calling `getScope()` before a scope was made the current one (=before init was called), we will return a scope for a noop client.
```
adinauer, line 81: What `init` are we referring to here? Sentry.init as in initializing the SDK or init of the scope?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,148 @@
...
+When a scope is forked via `withScope()`, a new scope is created that inherits all data currently set on the parent scope.
+
+The main change to Scopes is that they do not push/pop anymore, but instead fork an execution context (in languages where this makes sense/is possible).
+Basically, `withScope()` should behave like `hub.run()` does today in languages that have execution context forking.
+
+`client.getScope()` should return the current scope of this client in the current execution context.
+
+From a users perspective, this should mostly not be noticeable - they can always run `getScope()` to get the current scope, or `withScope(callback)` to fork a new scope off the current scope.
+
+You can make a scope the current one via `setCurrentScope(scope)`, which should bind the scope to the current execution context (or a global, in SDKs without execution context).
```
adinauer, line 114: I presume this means `Sentry.setCurrentHub(Sentry.getCurrentHub().clone())` will become something like `Sentry.setCurrentScope(Sentry.getScope().clone())`, is that correct?

adinauer (commented):
Left some questions / comments. Generally I'm surprised to see such a focus on client and its APIs. In Java SDK these are rarely touched by clients and rather hidden away.

text/0122-sdk-hub-scope-merge.md
mydea: I'd say that generally a client is always inherited, if a user overwrites it somewhere down the line than going forward this overwritten client will continue to be inherited. Basically:

```js
scope1 - client1
FORK -> scope2 - client1
  scope2.setClient(client2)
  FORK -> scope3 - client2
```

does that make sense?

text/0122-sdk-hub-scope-merge.md
mydea: just an edge case I'd say! Normally users should not do that, that happens inside of `init()`.

text/0122-sdk-hub-scope-merge.md
mydea: Yeah, I mean `Sentry.init()` (will update to be more specific!)!

text/0122-sdk-hub-scope-merge.md
mydea: Yes, that's the idea! 👍

mydea:
> Left some questions / comments. Generally I'm surprised to see such a focus on client and its APIs. In Java SDK these are rarely touched by clients and rather hidden away.

This is something that should not concern most users, but for those who need it (multiple clients...) we need to make sure stuff still works/is possible to do. But yes, we should def. focus on making the "regular" use case most convenient and abstracted away!

text/0122-sdk-hub-scope-merge.md
adinauer: What I meant was:

```
scope1 - client1
FORK -> scope 2 - client 1
scope1.setClient(client2)
  scope2.getClient() <-- what does this return?
```

mydea: sorry, wanted to show this but my note was not specific enough xD

```js
scope1.getClient() // >> client1
FORK -> scope2
  scope2.setClient(client2)
  scope2.getClient() // >> client2
  FORK -> scope3
    scope3.getClient() // >> client2
```

adinauer: Hmm but if you now set a different `client` on `scope1`, would that affect forked scopes (`scope2` and `scope3`) as well?

mydea: ah, you mean would it affect _already forked_ scopes? I'd say not - I think forking should only happen at fork-time and after that the scopes should not affect each other anymore. Does that sound reasonable?

adinauer: It sure does sound easier to implement and reason about. Probably best then to start out this way as I think that's also how hubs behave at the moment.

mydea: yes, they do, I think! 👍 I will update the RFC to make this explicit/clear.

antonpirker:
I read it and everything looks reasonable. How clients and the noop client are handled i have not yet fully figured out for python. But I guess I need to tinker with some code to fully get the picture and give meaningful feedback. Will try to make time for this this week.

philipphofmann:
I don't have much to add from the Mobile perspective because, on Mobile, we encourage our users to call the static APIs `Sentry.captureException`, etc., and hardly any users interact with the hub or client. I don't see a problem merging the scope and the hub for Mobile.

> Generally I'm surprised to see such a focus on client and its APIs. In Java SDK these are rarely touched by clients and rather hidden away.

I agree with @adinauer. I would actually love to make the client internal/private. I would prefer passing options to the scope, and that's all the flexibility you get.

On Cocoa, we don't support creating your own implementation of a hub, client, or transport, and nobody ever complained. Our only complaints so far are that some customers want to send the data to Sentry and their backend, but that has no impact on this RFC.

Correct me if I missed something on other mobile platforms, @krystofwoldrich, @romtsn, @markushi, @brustolin, @stefanosiano.

adinauer:
> make the client internal/private

There are clients who customize clients, e.g. to support multi DSN or send events somewhere else as well. While rare this is still a use case we probably shouldn't drop without good reason.

mydea:
> > make the client internal/private
> 
> There are users who customize clients, e.g. to support multi DSN or send events somewhere else as well. While rare this is still a use case we probably shouldn't drop without good reason.

IMHO the goal should be to hide the client away as much as possible by default (which we already do anyhow generally, via `init()` - nobody will generally see a client reference in this scenario), but allow to configure it for "edge cases" (as @adinauer mentioned). But I also think that this can be decided in detail on a SDK-by-SDK basis, where it makes sense to adjust this, do it! (FWIW in JS there are _some_ scenarios where you need to interact with a client manually)

text/0122-sdk-hub-scope-merge.md
mydea: I updated this so only the scope has a reference to the client, but the client remains stateless.

text/0122-sdk-hub-scope-merge.md
mydea: 👍 moved this up.

text/0122-sdk-hub-scope-merge.md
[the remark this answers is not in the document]
mydea: updated this accordingly!

text/0122-sdk-hub-scope-merge.md
mydea: added a note describing this!

text/0122-sdk-hub-scope-merge.md
mydea: Added a sentence making clear that this is meant to replace `makeMain` or `setCurrentHub` APIs.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,161 @@
...
+
+From a users perspective, this should mostly not be noticeable - they can always run `getScope()` to get the current scope, or `withScope(callback)` to fork a new scope off the current scope.
+
+You can make a scope the current one via `setCurrentScope(scope)`, which should bind the scope to the current execution context (or a global, in SDKs without execution context). This is a replacement for the current APIs like `makeMain(hub)` or `setCurrentHub(hub)`.
+
+You can still clone scopes manually the same way as before, e.g. via `Scope.clone(oldScope)` or a similar API. In contrast to `withScope()`, this will _not_ fork an execution context.
+
+You can update the client of a scope via `scope.setClient(newClient)`. This will not affect any scope that has already been forked off this scope, but any scope forked off _after_ the client was updated will also receive the updated client.
+
+## Should users care about Clients?
```
mydea, line 122: I added this section as it came up a few times in PR feedback. The idea is to make it explicit that the client should _not_ be a main user-facing concept, but more a power user feature.

mydea:
I applied all the feedback and moved this out of draft - I think this is ready to "really" review!

krystofwoldrich:
From RN point of view I don't we can merge Hub and Scope. 99% time users can use `Sentry.captureException` static API, the same as in browser JS.

RN Uses only one Hub and extends browser JS implementation of Scope to synchronize with the native SDKs (iOS, Android).

Since RN heavily depends on browser JS SDKs, we will have to adopt it basically immediately when this changes in JS. Event if the native SDKs haven't made the change at the time, it should not affect the scope synchronizing.

mydea:
> From RN point of view I don't we can merge Hub and Scope. 99% time users can use `Sentry.captureException` static API, the same as in browser JS.
> 
> RN Uses only one Hub and extends browser JS implementation of Scope to synchronize with the native SDKs (iOS, Android).
> 
> Since RN heavily depends on browser JS SDKs, we will have to adopt it basically immediately when this changes in JS. Event if the native SDKs haven't made the change at the time, it should not affect the scope synchronizing.

This change will def. happen in a major only, so there should be enough time etc. to sync this up with RN!
The general Scope API should also remain more or less similar to right now, so IMHO it should be possible to continue doing more or less what you're doing right now in RN (*small tweaks will probably be necessary, but nothing we can't figure out!)

philipphofmann:
After talking to more mobile folks, we don't see a problem with moving forward with this, but we don't see a need to change that cause it doesn't cause any problems in our SDKs. We will most likely make that change in one of our  next major versions, but we won't do a major version because of that.

mydea:
After talking about this with @mitsuhiko , I updated this RFC to again include the (updated) concepts of a global scope, as well as an isolation scope.

The idea is that ALL SDKs should have the concepts. Although it may be less relevant for some - but in these cases, they will simply do nothing anyhow. But esp. for isomorphic envs like JS, the same APIs should exist for server & client.

Please re-read the RFC for the added sections. We had some talks before about the fact that this is not equally relevant for all SDKs (which is why I initially split this into two RFCs), but @mitsuhiko reiterated that we should do this in an aligned way.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+You can still clone scopes manually the same way as before, e.g. via `Scope.clone(oldScope)` or a similar API. In contrast to `withScope()`, this will _not_ fork an execution context.
+
+You can update the client of a scope via `scope.setClient(newClient)`. This will not affect any scope that has already been forked off this scope, but any scope forked off _after_ the client was updated will also receive the updated client.
+
+Every scope is always tied to a client.
+
+## Global Scope
+
+In addition to the currently active scope, there will also be a new special scope, the **Global Scope**.
+The global scope is _not_ the initial scope, but a special scope that belongs to a client and is applied to any event that belongs to this client.
```
cleptric, line 138: This would make the client stateful all of a sudden. Not the end of the world, but feels odd.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+You can get the current global scope via `getGlobalScope()`. There _may_ be a function `setGlobalScope(scope)` to update the global scope - or SDKs can decide that there is no need to update the global scope, you can only mutate it.
+
+If you call `getGlobalScope()` before a client is initialized, we should still get a global scope back (tied to a Noop client). Once an actual client is initialized, the global scope of the noop client should be merged into the new global scope for the new client. This should ensure that even if you call `getGlobalScope().setTag(...)` before the SDK is initialized, no data is lost.
```
cleptric, line 142: Any idea where the clients will be stored? I feel like we need a Hub for that *scnr

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+The reason that the global scope is not the same as the initial scope of a client, is that you cannot accidentally mutate it - nothing ever inherits off the global scope.
+
+## Isolation Scopes
+
+Furthermore, there can also be **Isolation Scopes**.
+Similar to the global scope, these are also applied to events. However, isolation scopes can be created, either by us internally (the most common scenario), or also by users. The new APIs for this are:
```
cleptric, line 149: The client already holds the global scope, and now it also holds an isolation scope. But we do not expect to create a new client for each request, so the client should hold a stack of isolation scopes IMO, which will be tricky. Can we somehow clarify how this all should work?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+// This will NOT make this scope the isolation scope, but will create a new isolation scope (based on the currently active isolation scope, if one exists)
+scope.isolate();
+
+// Similar to `withScope`, but it forks a new scope AND sets a new isolation scope for this context
+export function withIsolationScope(callback: (scope) => void): void;
+```
+
+You can fetch the currently active isolation scope via `getIsolationScope()`. You can define a new isolation scope via `scope.isolate()`, which will define a new isolation scope for this scope, and for all scopes that will be forked off this scope. When a client is created & bound, an initial isolation scope will immediately be created, similar to the global scope for a client.
+
+An isolation scope is attached to the current execution context, similar to the active scope. There is always exactly one active isolation scope. If you call `getIsolationScope()` before a client has been created, a noop isolation scope is returned, which should be merged in once a client is actually created (same as with the global scope).
```
cleptric, line 165: How can there be only one active isolation scope if my application handles multiple requests concurrently?

text/0122-sdk-hub-scope-merge.md
mydea: Not necessarily. This does not mean that the global scope is defined on the client, but that the global scope has a reference to the client :) the mechanism to look up the global scope can/should probably be more in parallel to getting the active scope. Think of it like this:

```js
const executionContext1; // this is one execution context somewhere
const executionContext2; // this is some other execution context somewhere else

executionContext1.globalScope --> Scope
executionContext2.activeScope --> Scope
executionContext3.isolationScope --> Scope
```

So all three of these scopes can be put on the execution context (or wherever the active scope is put). How _exactly_ this is done would be up to the SDK/language, but the idea is that the client remains stateless.

text/0122-sdk-hub-scope-merge.md
mydea: clients should be referenced on the scope! Will update this to make this clear. Each scope will always have a client attached (if you just do e.g. `new Scope()`, it will have a Noop Client). You lookup the client via the scope, same as you would today look it up via the hub!

text/0122-sdk-hub-scope-merge.md
mydea: Added a comment above hopefully explaining the idea here - was probably not fully clear from how I wrote this down in the RFC. TLDR: active scope, global scope, isolation scope should be stored "next to each other" on each execution context. nothing should be stored directly on the client, which should remain stateless.

text/0122-sdk-hub-scope-merge.md
mydea: There is one active isolations cope __in each execution context__. Will update this to clarify 😅 You are of course right that there can be multiple active isolation scopes in your application, the same way you can have multiple active scopes in your application at the same time!

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+APIs that are currently on the hub should instead be called directly on the scope (e.g. `scope.captureException()` etc.), or via a global method (e.g. `Sentry.captureException()`).
+
+The current scope may be kept similar to how we currently keep the current hub, but this is SDK specific and not part of this RFC.
+
+## Clients
+
+Instead of a client being optional, there will now _always_ be a client. It may be a Noop Client that does nothing, if `init()` has not been called yet.
+
+A scope has a reference to a client. By default it will reference a noop client. You can bind a client to a scope via `scope.setClient()`. 
+The client is inherited by forked scopes.
```
lforst, line 80: I think `scope.setClient()` is potentially dangerous to have due to its mutable nature. What use-cases do we need it for? If possible I would avoid it at all cost.

IMO much better would be to do `client.withScope()` for any cases where `scope.setClient()` would currently be applicable.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+This RFC does not propose any concrete way to store the current scope. This is up to the concrete SDK and may behave the same way as it currently does for the hub, or differently if that makes more sense in a given scenario.
+
+# Drawbacks
+
+* This changes _a lot_ of public APIs and behavior.
+
+# Unresolved questions
+
+TODO
```
lforst, line 305:
```suggestion
- What is the user experience like if there is no instrumentation to set isolation scopes?
```

text/0122-sdk-hub-scope-merge.md
mitsuhiko: I think `client.withScope` is confusing and raises questions. The client does not own scopes, so why would it make up one to bind itself? Logically the client now is a property of the scope like any other thing. Rebinding it on the scope is fine. The client on the scope needs to be mutable anyways to make rebinding possible, even if it's just internal. So we don't win anything by not exposing this.

lforst: I am sus but let's go for `scope.setClient()` then. 😄 My FP brain doesn't like this API - it's too imperative.

text/0122-sdk-hub-scope-merge.md
sentrivana: I find `forkScope` better, also considering this:

> You can still clone scopes manually the same way as before, e.g. via `Scope.clone(oldScope)` or a similar API. In contrast to `withScope()`, this will _not_ fork an execution context.


Both essentially create a new scope off of an existing one, but in different ways. It'd be good to have the name communicate what the difference is.

text/0122-sdk-hub-scope-merge.md
sentrivana: I was also initially confused by this, but I think it's just a matter of wording this a bit differently. It says:

> but a special scope that belongs to a client and is applied to any event that belongs to this client

This makes it sound like the client owns the scope, even though technically it's the other way around. Maybe we can write something like

```suggestion
The global scope is _not_ the initial scope, but a special scope that has a client and is applied to any event that belongs to its client.
```

szokeasaurusrex:
I agree with @antonpirker's assessment that these changes seem reasonable at first glance for the Python SDK.

My one suggestion for the RFC would be to consider adding a visualization, similar to the one on the [unified API docs page](https://develop.sentry.dev/sdk/unified-api/#simplified-visualization), to describe how these changes will affect the API at a high level.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+Example for hooking into external auto-instrumentation (e.g. OpenTelemetry):
+
+```ts
+let onRequestHook: (span: Span) => void;
+
+// This method is not defined by us, but is some external code
+// Here just for demonstration purposes of how that may be implemented
+function otelWrapHttpServerRequest(original: Function): Function {
+  // Fork an execution context for this server request,
```
sl0thentr0py, line 197: what is the semantic we are trying to achieve with isolate? is it per execution context? then why are we using 'fork an execution context' here. sorry but it really isn't clear to me what we're trying to achieve with this example. What kind of data lives on the `withScope` and what happens after you do `scope.isolate`?

sl0thentr0py (changes requested):
did a first round, i need to think things through

but overall it is quite unclear what the purpose of isolation scope is. Can you maybe add a non javascript example with a thread, for instance? What would be do on a `thread.start` monkeypatch?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+An isolation scope is attached to the current execution context, similar to the active scope. There is always exactly one active isolation scope. If you call `getIsolationScope()` before a client has been created, a noop isolation scope is returned, which should be merged in once a client is actually created (same as with the global scope).
+
+Similar to the global scope, an isolation scope is always a separate scope, so nothing will inherit off it - except for a potential superseding isolation scope. 
+If an isolation scope is created, and there is already an isolation scope in the current execution context, then the new isolation scope should be forked off the previous one (with copy-on-write).
+
+### When to create an isolation scope
+
+For most server-side SDKs, an isolation scope will be created for each request being processed. 
+Roughly, it will equate to each time we currently fork a hub.
```
sentrivana, line 173: Why not just fork the current active scope? What does the isolation scope offer compared to just setting the active scope? Maybe a clarification here would be good.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+export function setCurrentScope(scope: Scope): void;
+
+// get the currently active global scope
+export function getGlobalScope(): Scope;
+
+// get the currently active isolation scope
+export function getIsolationScope(): Scope;
+
+// similar to `withScope`, but defines an isolation scope
+export function withIsolationScope(callback: (scope: Scope) => unknown): unknown;
```
kamilogorek, line 64: I'm not sure what those mean to do tbh. Sounds like it'd be the same as `getScope` and `withScope`.

sentrivana:
🥔 for this RFC! Left some comments.

What I'd find especially helpful is some sort of mapping of the current concepts (e.g. creating a new hub) and/or common cases (e.g. wrapping a request processor) to the proposed concepts in this RFC (e.g., creating an isolation scope). You're already doing this in couple places in the RFC, but maybe it makes sense to sum up the most common cases in a table somewhere with a small explanation.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+You can update the client of a scope via `scope.setClient(newClient)`. This will not affect any scope that has already been forked off this scope, but any scope forked off _after_ the client was updated will also receive the updated client.
+
+Every scope is always tied to a client.
+
+## Global Scope
+
+In addition to the currently active scope, there will also be a new special scope, the **Global Scope**.
+The global scope is _not_ the initial scope, but a special scope that belongs to a client and is applied to any event that belongs to this client.
+
+You can get the current global scope via `getGlobalScope()`. There _may_ be a function `setGlobalScope(scope)` to update the global scope - or SDKs can decide that there is no need to update the global scope, you can only mutate it.
```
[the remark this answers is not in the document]
antonpirker: Was about to ask what happens with the existing global scope. :-)

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+// Similar to `withScope`, but it forks a new scope AND sets a new isolation scope for this context
+export function withIsolationScope(callback: (scope) => void): void;
+```
+
+You can fetch the currently active isolation scope via `getIsolationScope()`. You can define a new isolation scope via `scope.isolate()`, which will define a new isolation scope for this scope, and for all scopes that will be forked off this scope. When a client is created & bound, an initial isolation scope will immediately be created, similar to the global scope for a client.
+
+An isolation scope is attached to the current execution context, similar to the active scope. There is always exactly one active isolation scope. If you call `getIsolationScope()` before a client has been created, a noop isolation scope is returned, which should be merged in once a client is actually created (same as with the global scope).
+
+Similar to the global scope, an isolation scope is always a separate scope, so nothing will inherit off it - except for a potential superseding isolation scope.
```
antonpirker, line 167: I do not understand this. Can you explain?

antonpirker (commented):
First @mydea thanks for tackling this! 

The things that confuse me right now:
We have a "global scope", "initial scope", "currently active scope", "current active isolation scope" and the "initial isolation scope". 

I do not see the difference between "global scope" and "initial scope", in my mind they could be the same.

Also the difference between the "currently active scope" and an "isolation scope" is not clear. I did not get the use case for an "isolation scope".

text/0122-sdk-hub-scope-merge.md
antonpirker: I mean the `Similar to the global scope, an isolation scope is always a separate scope, so nothing will inherit off it - except for a potential superseding isolation scope.` part.

text/0122-sdk-hub-scope-merge.md
cleptric: Isn't `withScope()` mainly used to create a temporary scope which is only applied to the callback?
I don't see a good reason to rename/remove this.

text/0122-sdk-hub-scope-merge.md
mydea: The tricky thing to convey (and actually, we need to specifically spec this out), is that each client has a separete global scope. 

At least, I think this is what makes most sense 🤔 When talking with @mitsuhiko both options are possible (global scope is _truly_ global, or global scope is basically a "client scope"). 

Personally I tend to the idea that each client should have it's own global scope - simply because thinking about e.g. MFE and similar, we probably don't want to "pollute" other clients when doing `getGlobalScope().setTag()`. 

Wording wise, we could say:

> The global scope is _not_ the initial scope, but a separate, special scope. The global scope is different for each client, and is applied to every event.

Or something like this?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+
+## Scopes
+
+Scopes behave similar to how they behave today. 
+When a scope is forked via `withScope()`, a new scope is created that inherits all data currently set on the parent scope.
+
+The main change to Scopes is that they do not push/pop anymore, but instead fork an execution context (in languages where this makes sense/is possible).
+Basically, `withScope()` should behave like `hub.run()` does today in languages that have execution context forking.
+
+`client.getScope()` should return the current scope of this client in the current execution context.
```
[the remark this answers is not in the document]
mydea: Ah, damn, this should have been removed already (this is from an earlier iteration of this RFC). Client should not hold a scope reference, I'll remove this. Sorry about the confusion there.

text/0122-sdk-hub-scope-merge.md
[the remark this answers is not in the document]
mydea: will clarify this!

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,305 @@
...
+  public captureEvent(event: Event, additionalScope?: Scope) {
+    // Global scope is always applied first
+    const scopeData = getGlobalScope().getScopeData();
+
+    // Apply isolations cope next
+    const isolationScope = getIsolationScope();
+    merge(scopeData, isolationScope.getScopeData());
+    
+    // Now the scope data itself is added
+    merge(scopeData, scope.getScopeData());
```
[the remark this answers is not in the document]
mydea: it's actually wrong, should be `this` - will fix the example!

text/0122-sdk-hub-scope-merge.md
mydea: I will add a graphic to hopefully explain this better :D

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+Global scopes will be documented in a new section under "Enriching Events > Global Data", which will explain how to use `getGlobalScope()` and when/how data added there will be applied.
+
+### Isolation Scope
+
+Furthermore, there are also **Isolation Scopes**.
+
+An isolation scope is similar to a global scope, but tied to a specific section of your app/code that should be isolated.
+In a server side language, an isolation scope will commonly be created per request. Each request gets an isolated scope that you can assign data to, which should be applied to all events that are processed in this request. However, an isolation scope is not limited to requests - you could also have an isolation scope for a scheduled job, or for a zone in your UI (in theory).
+
+The isolation scope can be used to e.g. add user or request data that should be applied to any event for the current request (or similar isolation unit). Similar to the global scope, it does not matter where you call it, but you can always be sure that data is correctly attached/available. This can especially be useful when writing hooks or plugins where you may not easily get access to the scope of a request:
```
AbhiPrasad, line 233: Let's explicitly call out that breadcrumbs should be added on the isolation scope.

mydea: I mention this further down here: https://github.com/getsentry/rfcs/blob/fn/merge-hub-scope/text/0122-sdk-hub-scope-merge.md#what-should-be-called-from-top-level-methods

but it's a general challenge for this RFC that stuff feels like it should be explained earlier that comes up later, making it hard to keep it somewhat readable 😬

text/0122-sdk-hub-scope-merge.md
mydea: IMHO generally we should strive to replace this usage of `withScope` - only to add a temporary event processor - with different pattern(s). (not possible for _all_ cases, but most of them). So generally there should be little (not none) need to use it manually!

text/0122-sdk-hub-scope-merge.md
AbhiPrasad: My comment here was I think that the breadcrumbs use case helps build the best intuition here why we need a isolation scope concept, it's also why we need something like this while opentelemetry does not fwiw.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+
+// Somewhere else
+async function setupGlobalData() {
+  const revision = await getRevision();
+  // No need to worry or care about where this is called, this will _always_ work!
+  Sentry.getGlobalScope().setTag("revision", "abc");
+}
+```
+
+The global scope is _not_ the initial scope, but a separate, special scope. The global scope is different for each client, in order to avoid polluting this on accident when working with multiple completely separate clients.
```
philipphofmann, line 199: I'm a bit confused as to why the global scope is different for each client. You set the client to the scope, but you also need to keep a global scope for every client. If I got this right, do we need some bookkeeping of global scopes for clients?

philipphofmann, line 199: I am confused as to why the global scope is different for each client. If I got this right, do we need some bookkeeping of global scopes for clients? How is that going to work?
I would expect that the global scope is global 😄. It applies to all events.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+// Auto-instrumentation automatically forks an execution context with a new isolation scope
+app.get("/my-route", function () {
+  const isolationScope = Sentry.getIsolationScope();
+  isolationScope !== initialIsolationScope;
+  // Isolation scope inherits from previously current isolation scope
+  isolationScope.getTags(); // -> tag1: aa
+});
+```
+
+Note that in environments where you do not have multiple execution contexts (e.g. Browser JS), there is basically only a single isolation scope that is fundamentally the same as the global scope. In these environments, global & isolation scope can be used completely interchangeably. However, for consistency we still have these concepts everywhere - but users do not need to care about them, and we can reflect this in the SDK specific documentation.
```
philipphofmann, line 279: I don't think adding the isolation scope is a good idea when it's the same as the global scope. People will ask questions and need clarification about what the difference is. If it's the same, I would drop the isolation scope. We can always add it.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+
+### How to document isolation scopes
+
+Isolation scopes will be documented in a new section under "Enriching Events > Per request data" (or similar, based on SDK language & needs), which will explain how to use `getIsolationScope()` and when/how data added there will be applied.
+
+Global methods to set data, like `Sentry.setTag()` or `Sentry.setUser()`, will set on the isolation scope. Only if you want to set on a different scope, you need to manually get it (e.g. `getGlobalScope().setTag()`).
```
philipphofmann, line 285: On mobile, global methods should change the global scope, as we don't need an isolation scope, IMO.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+  scope.isolate(); // Add an isolation scope to the already forked scope
+};
+```
+
+## Applying scopes
+
+Scopes are applied in this order to events:
+
+```ts
+class Scope {
```
philipphofmann, line 334: So, clients will not be aware that there is a scope. The `Scope` class merges data into the event before passing it to the client. Did I get this correctly?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+
+# Drawbacks
+
+- This changes _a lot_ of public APIs and behavior.
+
+# Unresolved questions
+
+- How should we handle this in docs?
+- Naming: Can we come up with something better than Isolation Scope?
+- Global Scope: Should it be per-client (current RFC) or truly global?
```
philipphofmann, line 456: IMO, if you call it global, then it should be truly global. If you want it to be per-client, I would choose a different name.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+In addition to the current scope, this RFC introduces two new concepts:
+
+- [Global Scope](#global-scope)
+- [Isolation Scope](#isolation-scope)
+
+These scopes will be explained in detail below. They are _additional concepts_ that can be used by SDK internals & power users to make sure data can be assigned to the correct place, without relying on internals about how/when/where a scope or execution context is forked.
+
+The following chart gives an overview of how these scopes work and to which events they would be applied:
+
+![image](./0122-images/scope-inheritance.png)
```
philipphofmann, line 150: Is the green scope in that diagram the current scope? If yes, you could update the diagram by renaming scope to current scope to make it more straightforward.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+
+### How to store/reference scopes
+
+Similar to how we currently store the current hub, we also need to store the current scope. The exact mechanism to do that is up to the concrete SDK/language. In environments with multiple execution contexts (or thread locals), we should attach the current scope to the execution context (or thread local). In other environments, it may be attached to some global (e.g. in Browser JS). This RFC does not propose to change this, so generally the current scope may be stored the same was as the current hub used to be.
```
philipphofmann, line 154: If we store the current context globally on mobile and Brower JS, where there is no execution context, how is the current scope different from the global scope?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,458 @@
...
+
+```js
+Sentry.init();
+setupGlobalData();
+
+// Somewhere else
+async function setupGlobalData() {
+  const revision = await getRevision();
+  // No need to worry or care about where this is called, this will _always_ work!
+  Sentry.getGlobalScope().setTag("revision", "abc");
```
philipphofmann, line 195: On Mobile, `Sentry.setTag("revision", "abc")` already does what `Sentry.getGlobalScope().setTag("revision", "abc");` does, as the static API of `Sentry.` has a hub bound including a scope.

philipphofmann (commented):
Thanks for the detailed RFC @mydea 👏 

I like the idea of merging hub & scope, but I don't think an isolation scope is needed for Mobile. It will only confuse users. Furthermore, I'm confused about where to store the current scope on Mobile and Browser JS. If we keep it globally, it's basically the same as the global scope.

text/0122-sdk-hub-scope-merge.md
mydea: This is something I am not 100% sure about yet. On the one hand, it kind of makes sense that a global scope is _truly_ global. Thinking about this some more, probably you are right, and this should be truly global. If you have multiple clients, it's probably up to you to simply not use the global scope if you care about this not being polluted!

text/0122-sdk-hub-scope-merge.md
mydea: it's not the same - the global scope will always be the same, while the isolation scope _can_ be changed per context. In environments where there are no execution contexts (e.g. browser or probably mobile), they are more or less the same. But the idea is generally to not really expose this to users too much - users should just use `Sentry.doX` and not `Sentry.getCurrentScope().doX`, and we route it to the correct scope under the hood. that should work just as well in browser or mobile.

So the concept should still be there for consistency sake, and it _does_ work (e.g. you can call it), it's just not commonly useful by default - which is fine, nobody is forced to use it, and users should not have to think about it! _but_ for example looking at browser, if somebody decides to add and setup zone.js or something like this, or browsers eventually add something like execution contexts, we _can_ then also wire this up to work there as well.

text/0122-sdk-hub-scope-merge.md
mydea: it can/should still set on the isolation scope, which in absence of execution contexts will be functionally basically global. But _if_ execution contexts are ever added in some way, this will then also "just work" - that's the idea here, at least.

text/0122-sdk-hub-scope-merge.md
mydea: that's the idea, this is also how it works today (at least in JS and I _think_ more or less everywhere?) - this RFC does not propose to change this from it is today, except that the special scope data is _also_ applied to the events!

text/0122-sdk-hub-scope-merge.md
mydea: In e.g. mobile or browser, we'll need to implement the current scope somehow differently. Either we still need to do it via some kind of stack, or with some other mechanism - the idea is to get as close to this desired UX as possible, with the tech that is available!

text/0122-sdk-hub-scope-merge.md
philipphofmann: I'm OK with your proposal.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,455 @@
...
+
+## What about globals?
+
+This RFC does not propose any concrete way to store the active scope. This is up to the concrete SDK and may behave the same way as it currently does for the hub, or differently if that makes more sense in a given scenario.
+
+# Drawbacks
+
+- This changes _a lot_ of public APIs and behavior.
+
+# Unresolved questions
```
philipphofmann, line 449: We still need to figure out how scope sync across multiple layers works. For example, on RN we sync the scope to Cocoa, which syncs the scope to SentryCrash. I guess the SDKs should only sync the global scopes. We don't sync the scope from a lower layer, for example, Cocoa to RN.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+// This would be our custom sentry configuration
+onRequestHook = () => {
+  const scope = getCurrentScope();
+  scope.isolate(); // Add an isolation scope to the already forked scope
+};
+```
+
+## Applying scopes
+
+Scopes are applied in this order to events:
```
markushi, line 346: Given the example on how scopes are merged I'm wondering what happens the other way round when scopes are read? E.g. Would `Sentry.getCurrentScope().getRelease()` return a value when the release was set on the global scope ~ `Sentry.getGlobalScope().setRelease("1.0.0")`

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+This RFC does not propose any concrete way to store the active scope. This is up to the concrete SDK and may behave the same way as it currently does for the hub, or differently if that makes more sense in a given scenario.
+
+# Drawbacks
+
+- This changes _a lot_ of public APIs and behavior.
+
+# Unresolved questions
+
+- How should we handle this in docs?
+- Naming: Can we come up with something better than Isolation Scope?
```
markushi, line 473: Just throwing in some ideas: How about having a Global Scope / Root Scope / Current Scope? Or if we want to be similar to otel, naming it Context instead of scope; although it's not exactly the same as context is just a part of the scope as I understood it. Then we'd have Global Execution Context / Root Execution Context / Current Execution Context.
Not super happy with my ideas either, as root could imply some tree structure 🤷

text/0122-sdk-hub-scope-merge.md
mydea: It's a good question! I would say it should only return the value on the actual scope being accessed. If you want to have some cascading behavior it should be a top-level method like `Sentry.getRelease()` or something like this?

text/0122-sdk-hub-scope-merge.md
mydea: We also had root as a name already in earlier iterations here 😅 which was also found to not be super understandable by some folks. It's a hard naming problem 😬

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+
+It also proposes the new concepts of global & isolated scopes.
+
+# Background
+
+Hubs and Scopes have been part of the unified SDK API for a long time.
+While they have generally served us well, the shortcomings of the current systems have become apparent over time.
+Apart from generally being unclear to users - what is applied when, why, how? - there are also problems with the way we currently instruct users to add event data.
+
+A major reason for this change is also to better align with OpenTelemetry. In OpenTelemetry (OTEL), the equivalent to our scope is the "Context". However, the Context is immutable in OTEL, and works via copy-on-write - so each time you want to change _anything_ on the context, you fork it. This means that you have much more forking than you have in Sentry currently, and also since you always fork a context when you change the active span (as the span is set on the context, and updating it means forking a context), you can quickly get into deeply nested context situations due to auto instrumentation.
```
adinauer, line 25: This RFC doesn't seem to mention where to store the current span. While we might no longer have a Sentry span after completely switching over to OTEL, there'll likely be intermediate steps as well as some SDKs that can't yet move to pure OTEL spans.

Where we could put it:
- Isolation span
	- I don't think that's a good idea as
		- It's a potential footgun allowing users to hide the current span set on isolation scope by the SDK by setting a span on the current span
		- When a span ends we have to restore the previous span on the isolation scope
- Current Span
	- Since we're forking current scope anytime a new span is created anyways this is a good candidate
	- A drawback here is that it requires a more complex copy-on-write implementation that's per property instead of simply copying the whole scope. It seems the number one reason for copying scope will be creating a new span.
- Execution Context
	- This would keep current span separate from current scope and thus not cause as many current scope copies

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+## Should users care about Clients?
+
+Generally speaking, for most regular use cases the client should be mostly hidden away from users.
+Users should just call `Sentry.init()`, which will setup a client under the hood. Users should generally only interact with scopes, and we should keep clients out of most public facing APIs.
+
+The client is only there to allow an escape hatch when users need to do more complex/special things, like isolating Sentry instances or multiplexing. So client APIs should be designed to _allow_ to do things that cannot be done via `Sentry.init()`, but our main focus should be on making the default experience easy to understand, which includes that users should not have to care about the concept of clients by default.
+
+## What about other Hub references?
+
+While the Hub is mainly exposed via `getCurrentHub()`, it is also used as argument or similar in many places.
```
adinauer, line 438: Wouldn't this need to access all three scopes (global, isolation and current) instead of the single hub?

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+
+```js
+const executionContext1 = getExecutionContext();
+const { scope, isolationScope } = executionContext1.getScopes();
+executionContext1.setScopes({ scope, isolationScope });
+```
+
+In environments where you don't have execution contexts, you may store them in a global place. The exact mechanism to do this is up to the concrete SDK/language.
+
+For each execution context, there will _always_ be exactly one current & isolation scope. When the execution context is forked, these are forked along.
```
adinauer, line 180: I feel like this can easily be misinterpreted as if you fork execution context (e.g. because a new span was started) you always have to fork isolation scope and current scope.

Maybe we should add a chapter about forking scopes as well as passing scopes along when e.g. changing to another thread. I feel like whenever we fork an isolation scope we should also fork the current scope as otherwise we'd open the door for scope bleed and always require users to call both `withIsolationScope` and `withScope` in combination. When moving to another thread we should bring along both isolation and current scope to avoid scope bleed from things that happened on that thread before.

adinauer (commented):
After giving this some more thought, looking at the python implementation currently being done and starting to think about how this would work for Java, I've now given this a more in depth review - sorry it's happening so late.

adinauer (commented):
Today @antonpirker and me discovered some more questions / remarks.

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+APIs that are currently on the hub should instead be called directly on the scope (e.g. `scope.captureException()` etc.), or via a global method (e.g. `Sentry.captureException()`).
+
+The current scope may be kept similar to how we currently keep the current hub, but this is SDK specific and not part of this RFC.
+
+You can find some examples for how the APIs can/should be used [here](https://gist.github.com/mitsuhiko/1bc78d04ea7d08e5b50d27e42676db80).
+
+## Clients
+
+Instad of picking the client from the current hub, going forward a client should always be fetched via `Sentry.getClient()`. 
+This should _always_ return a client. If `Sentry.init()` was not called yet, it should return a Noop Client that does nothing.
```
jamescrosswell, line 94: > Instad of picking the client from the current hub, going forward a client should always be fetched via `Sentry.getClient()`.

This feels odd. If SDK users are only using one client (`Sentry.getClient()` implies a static/singleton reference to a Sentry Client) then this could work. It might be easier to make the Scope Stack a property of the client though and use typical DI techniques to get a reference to the Client... avoiding all the problems with statics/singletons.

If SDK users have multiple clients, getting the client via `Sentry.getClient()` is problematic. It results in weird things like temporarily pushing a different client to the scope stack (so `Sentry.getClient()` can return the top/current client from the stack)... in the case of Global/Shared scope, this could easily lead to problems. 

For SDK users and SDK designers, it would be easier if the ScopeStack was a property of the client. **_If_** SDK users wanted two different clients to share the same Scope, those clients could be configured/initialised with some kind of shared scope stack (like the `GlobalScope` from this RFC)... but events would always end up being sent to the correct Sentry instance/project. 

I'm not sure I understand why/when anyone would want the client to be ambient context (in the same way that scopes are) or why anyone would maintain a stack of clients (except as a workaround to our SDK design - namely that we retrieve the client reference via a static/singleton reference).

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
...
+
+Global scopes will be documented in a new section under "Enriching Events > Global Data", which will explain how to use `getGlobalScope()` and when/how data added there will be applied.
+
+### Isolation Scope
+
+Furthermore, there are also **Isolation Scopes**.
+
+An isolation scope is similar to a global scope, but tied to a specific section of your app/code that should be isolated.
+In a server side language, an isolation scope will commonly be created per request. Each request gets an isolated scope that you can assign data to, which should be applied to all events that are processed in this request. However, an isolation scope is not limited to requests - you could also have an isolation scope for a scheduled job, or for a zone in your UI (in theory).
+
```
jamescrosswell, line 233: Maybe not necessary to implement IsolatedScopes in all SDKs. 

The dotnet SDK has an asynclocal scope manager that achieves this already. I'm not sure how unique that is to .NET though and whether equivalents exist in other languages/runtimes.

text/0122-sdk-hub-scope-merge.md
adinauer: This should just work like other things on static API that interact with the scope by creating a sort of view on top of global, isolation and current scope, then handing back whatever wins according to logic that should be detailed somewhere in this RFC. There shouldn't be a need to change the stack and stack should in theory go away with the proposed changes here.

text/0122-sdk-hub-scope-merge.md
adinauer: Does that mean in dotnet you can have both an isolation scope (via asynclocal scope manager) and a separate current scope?

jamescrosswell: Possibly but it's not really a problem we'd need to solve in .NET. In ASP.NET Core requests get handled by middleware which is async/await all the way through and so we use an [asynclocal scope container](https://github.com/getsentry/sentry-dotnet/blob/cd166964b81b745cbbcda914fc1a601bd3e964dd/src/Sentry/Internal/ScopeStack/AsyncLocalScopeStackContainer.cs) (that has affinity for these requests). 

https://github.com/getsentry/rfcs/blob/c4b1fb4fd2915246760156285393340af093d591/text/0122-sdk-hub-scope-merge.md?plain=1#L304

... this isn't really necessary in .NET as we already get that for free from asynclocal.

adinauer: I was trying to say that parts of static API interact with isolation scope and other parts interact with current scope. I'm guessing you have to propagate both using this asynclocal scope container. Not sure about .NET but there'll probably be some `Sentry.withScope()` and/or `Sentry.pushScope/popScope` API that forks current scope.

jamescrosswell: > Not sure about .NET but there'll probably be some `Sentry.withScope()` and/or `Sentry.pushScope/popScope` API that forks current scope

There is yes. There's the [SentrySdk.ConfigureScope](https://docs.sentry.io/platforms/dotnet/enriching-events/scopes/#configuring-the-scope) which clones the current scope and pops the clone on the stack, then removes the clone once it's done (ie. when it's disposed). 

I guess we could replace that with `SentrySdk.GetIsolationScope` or similar but I don't understand why we would. It doesn't seem to solve a problem that any of our users have and it doesn't get us any closer to using OTEL for spans (at least not for the .NET SDK - I realise it may be a prerequisite to this for the JavaScript SDKs).

text/0122-sdk-hub-scope-merge.md
```diff
@@ -0,0 +1,475 @@
+- Start Date: 2023-11-07
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/122
+- RFC Status: draft
+
+# Summary
```
adinauer, line 6: Since this RFC is no longer being updated, we're writing down how this works and collecting learnings etc. here: https://develop.sentry.dev/sdk/hub_and_scope_refactoring/

PR #123, opened 2023-11-15
Metrics / Span Correlations

mitsuhiko:
This RFC proposes a way in which metrics and spans can be correlated. This is a work in progress.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/metrics-correlation/text/xxxx-metrics-correlation.md)

### DIFF
--- a/text/xxxx-metrics-correlation.md
+++ b/text/xxxx-metrics-correlation.md
@@ -9,7 +9,80 @@ This RFC addresses the high level metrics to span correlation system.
 
 # Motivation
 
-WRITE ME
+We believe the value in a good metrics solution is the correlation to traces and other signals.
+This means that we need to store evidence of metric measurements on the spans.
+
+There are two APIs related to metrics in an SDK.  The low-level `metrics` API which directly
+throws a data point into a local aggregator, and a higher-level span-bound API that more
+explicitly associates measurements with spans directly.  Because there are two APIs provided
+we need to be careful setting customers on the path of success.
+
+# Basics
+
+For correlation purposes, measurements can be stored on the span within the `measurements`
+data bag:
+
+```json
+{
+  "span_id": "deadbeef",
+  "op": "whatever",
+  "metric": "my.span",
+  "measurements": {
+    "duration": { "value": 32.0, "unit": "millisecond" },
+    "block_size": { "value": 1024, "type": "g" }
+  }
+}
+```
+
+The key `duration` here is special in that it is the "trivial" metric name.  Any
+other measurement can be stored there.  The key of in the bag is concatenated to the
+`metric` key of the span except for the `duration` case.  That means that in the above
+case the metrics `custom/my.span` is emitted as distribution (the default) with
+`32.0` as value and `custom/my.span.block_size` for instance as gauge.
+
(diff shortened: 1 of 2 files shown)


text/xxxx-metrics-correlation.md
```diff
@@ -0,0 +1,140 @@
...
+
+However for finding interesting samples, the existence of a span that has that metric
+in it (eg: `count > 0`) on the right tags might be sufficient.
+
+On the other hand if `incr` is incremented multiple times, it might also be worth finding
+the spans sorted by the highest `count` or highest total sum / max value.
+
+# Open Questions
+
+* Metric vs span tags
```
AbhiPrasad, line 137: should hopefully be addressed by https://github.com/getsentry/rfcs/blob/main/text/0116-sentry-semantic-conventions.md - let's just go all in on OpenTelemetry conventions

text/xxxx-metrics-correlation.md
```diff
@@ -0,0 +1,140 @@
...
+of the metrics APIs are span seeking which means that they record a measurement in relation
+to that span.  Some APIs (such as `metrics.timing` when used with a code block) will instead
+create a span and bind it.
+
+```python
+def process_batch(batch):
+    processor = Processor()
+
+    # This creates a span with op `timing_measurement`
+    with metrics.timing("processor.process_batch"):
```
AbhiPrasad, line 28: Let's use the naming scheme of span name/metric name instead of using `op` (aligns with otel, removes educating about op types, aligns with other metrics products).

Op already has a meaning to sentry today (`db.X`, `resource.Y`), and that concept generally has caused us issues, let's not introduce it for any metrics API (we can always make op an optional thing people can set and map it to the same semantics as span op).

text/xxxx-metrics-correlation.md
```diff
@@ -0,0 +1,140 @@
...
+
+The corresponding low-level API (`metrics.timing`) is supposed to automatically create a
+span if tracing is enabled and no explict value is provided.  The following two examples
+are equivalent:
+
+```python
+with metrics.timing("foo"):
+    pass
+
+with start_span(op="timing_measurement", metric="foo"):
```
AbhiPrasad, line 115: how does this work when we are using an otel API under the hood? We'll have to add logic in otel span processors I guess?

AbhiPrasad (commented):
I like this, makes incremental changes super easy.

text/xxxx-metrics-correlation.md
mitsuhiko: Yes, that is an open question I'm not fully sure about yet. OTel Metrics themselves do not suggest any span based metrics changes. They are doing the correlations OOB with exemplars: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#exemplars

But I do think the case of span + timed metric is very common and ideally we can find a way to funnel that out with otel span APIs somehow.

text/xxxx-metrics-correlation.md
mitsuhiko: I mostly wrote this up in the light of what we have today. I agree that the op/description situation is not great so ideally this can recommend something else instead.

PR #125, opened 2023-11-22
ref: Amend transaction and spans RFC 118

philipphofmann:
Amend the decision of RFC 118 to use single-span ingestion.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/ref/ammend-118/text/0118-mobile-transactions-and-spans.md)

### DIFF
--- a/text/0118-mobile-transactions-and-spans.md
+++ b/text/0118-mobile-transactions-and-spans.md
@@ -23,9 +23,20 @@ As of now the way transactions are created has several drawbacks:
 
 4. Transaction durations are sometimes pointless: As they’re based on idle transaction. Any relevant or even non-relevant child span can influence the transaction duration.
 
+## Option Chosen
+
+On 2023-11-21, we decided to move forward with [4. Single Span Ingestion](#option-4), because single span ingestion will be available soon, and we strongly believe it's the future how SDKs will send spans to Sentry. Participants in the decision:
+
+- Philipp Hofmann
+- Karl Heinz Struggl
+- Markus Hintersteiner
+- Nar Saynorath
+- Shruthilaya Jaganathan
+- Pierre Massat
+
 # Options Considered
 
-### 1. Transactions as Carriers for Spans (preferred approach)
+### 1. Transactions as Carriers for Spans
 
 The idea is to use transactions as carriers for measurements on mobile instead of trying to make the backend heavy concept work on mobile. There is no active "open" transaction, but we rather create transactions on-demand, whenever there is span data to send.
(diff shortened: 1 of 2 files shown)


text/0118-mobile-transactions-and-spans.md
```diff
@@ -65,3 +75,18 @@ Have one transaction per screen. This automatic transaction can be fully managed
 ### 3. Leave it as-is
 
 Whilst being the least effort, this option doesn't add any value and we remain with all the drawbacks as outlined in the [background section](#background).
+
+### 4. Single Span Ingestion <a name="option-4"></a>
+
+Use single-span ingestion (PR) and keep screen load transactions. Whenever the SDK creates an auto-generated span, it sends it stand-alone to Sentry. To avoid multiple network requests, SDKs need to batch spans together; for example, send an envelope for every ten spans. The batch logic is still up for definition and is not the goal for this RFC. Instead, we are going to define this in an extra RFC.
```
kahest, line 81:
```suggestion
Keep screen load transactions, and use single-span ingestion ([PR](https://github.com/getsentry/relay/pull/2620)) whenever the SDK creates an auto-generated span and sends it stand-alone to Sentry. To avoid multiple network requests, SDKs need to batch spans together; for example, send an envelope for every ten spans. The batch logic is still up for definition and is not the goal for this RFC. Instead, we are going to define this in an extra RFC.
```

text/0118-mobile-transactions-and-spans.md
```diff
@@ -65,3 +75,18 @@ Have one transaction per screen. This automatic transaction can be fully managed
 ### 3. Leave it as-is
 
 Whilst being the least effort, this option doesn't add any value and we remain with all the drawbacks as outlined in the [background section](#background).
+
+### 4. Single Span Ingestion <a name="option-4"></a>
```
kahest, line 79: sidenote/question: why is it "single" span ingestion? I know we didn't invent this here, maybe you have a source

kahest (approved):
LGTM, just small wording suggestions

text/0118-mobile-transactions-and-spans.md
philipphofmann: That's just the wording people always brought up. I don't have any source.

markushi (approved):

markushi:
Tagging @AbhiPrasad here, as there were similar plans for browser JS.

philipphofmann:
I already mentioned in the TSC that we plan to use span ingestion for mobile starfish. I guess @AbhiPrasad should be aware, but thanks for tagging him here.

PR #129, opened 2024-02-06
rfc(feature): Video replay envelope

### LINKED ISSUE #128: RFC for Session Replay based on video snippets
- A new rrweb event type (number N) that hints to the browser (UI, Player) to handle the video recording
    - Add the caveats: Like no DOM snapshot (can we get around without the empty `<html>` node?
    - @brustolin  @romtsn  to draft an RFC
        - @billyvg  to share some rrweb specific examples/data/docs
        - Document how breadcrumbs live within it. Including touch events, etc.
- Timestamp in milliseconds
- type: N(?)
    - Type 4 and type 2 are required. 4 is metadata and used on the first event. 2 full snapshot of the DOM.
        - List on rrweb of reserved types. How do we get one?

https://github.com/rrweb-io/rrweb/blob/master/packages/types/src/index.ts

vaind:
How should Video-based replay behave:
- sending to Sentry (SDK side of things)
- processing for UI, if necessary
- high-level overview of the UI integration (rrweb?)

[Rendered RFC](https://github.com/vaind/sentry-rfcs/blob/feat/video-replay-envelope/text/0129-video-replay-envelope.md)

closes #128

Relates to: https://github.com/getsentry/develop/issues/1144


text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,49 @@
...
+In order to capture video replays and present them to the user, we need to define/decide:
+
+- how to transport the video data to the server
+- how to integrate the video data into the RRWeb JSON format
+- how to combine multiple replay chunks to a single replay session
+
+All of these influence one another and need to be considered together in a single RFC.
+
+Note: SDK-side implementation that is currently being worked on relies on taking screenshots and encoding them to a video.
+This is based on an evaluation where a video has much smaller size than a sequence of images (**TODO fix these: factor of X for 720p video**).
```
bruno-garcia, line 17: @brustolin @romtsn cc

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,49 @@
...
+We need this to to capture replays on platforms where it's not possible/feasible to produce an HTML DOM (i.e. the native format supported by RRWeb). For example: mobile apps.
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:Attachment
+
+- From the SDK, we should be able to use the existing envelope item type, [`attachment`](https://develop.sentry.dev/sdk/envelopes/#attachment) to transport the video data.
```
brustolin, line 30: I believe we can't use the type 'attachment' because it will impact attachment quota

romtsn: Agree, I guess we need a new type something like `MobileReplayRecording`, similar to [ReplayRecording](https://github.com/getsentry/sentry/issues/63255)

romtsn:
The Relay docs may help @vaind:
https://getsentry.github.io/relay/relay_event_schema/protocol/struct.Replay.html#
https://getsentry.github.io/relay/relay_event_schema/protocol/struct.ReplayContext.html

text/0129-video-replay-envelope.md
vaind: In that case I'd go with something less tied to being "mobile". For example even the Cocoa and Flutter SDKs could eventually work the same on desktop.

Instead, I'd suggest `ReplayVideoRecording`, assuming there's `ReplayRecording` (where does that come from? the linked page doesn't mention it)

romtsn: hm, yeah looks outdate 🤔 here they are https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L112-L115

`ReplayVideoRecording` sounds better indeed!

romtsn: Here's the js sdk envelope creation if you're interested https://github.com/getsentry/sentry-javascript/blob/develop/packages/replay/src/util/createReplayEnvelope.ts#L15

vaind: > hm, yeah looks outdated 🤔 here they are [getsentry/relay@`5fd3969`/relay-server/src/envelope.rs#L112-L115](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L112-L115)

I wonder if we actually need another envelope item type or we can use `ReplayRecording`. This indicates it's just treated as "data": 
```
    /// ReplayRecordings, large blobs sent by the replay sdk
    ReplayRecordings,
```
https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-kafka/src/config.rs#L49

vaind: Similarly, the JS code you've linked seems to take into account that it can be binary.

brustolin: > Here's the js sdk envelope creation if you're interested https://github.com/getsentry/sentry-javascript/blob/develop/packages/replay/src/util/createReplayEnvelope.ts#L15

I believe we're good to go just by copying this.

One item is ReplayEvent, another item is the video binary itself, the type could be an extension of the existing one like `replay_recording.video`.

vaind: The RRWeb JSON should also work fine, with the `Custom` event type, it can have any data JSON object content. On UI end, it just needs a custom handler code that seemed reasonable to do: https://github.com/rrweb-io/rrweb/blob/master/docs/recipes/custom-event.md

The missing bit (for me) is how how reference the video sent in another envelope item. Like do we give it some pre-authenticated link directly to GCS? Or is it going to be streamed through sentry servers?

billyvg: The UI/player side will likely need some additional metadata associated with the video (start/end timestamps) - where can we send this?

also cc @cmanallen for relay info

brustolin: > same envelope, different envelope item (type)

Oh, yes, make sense.

> The UI/player side will likely need some additional metadata associated with the video (start/end timestamps) - where can we send this?

This goes in the RRWeb json new type I believe. Something like:
```json
[
    {
       "type": 4,
       "data": {
          "href": "",
          "width": 390,
          "height": 720
       },
      "timestamp": 1707254438643
    },
    {
       "type": 8,
       "data": {
          "segment_id": 4,
          "duration": 5000
        }
    }
]
 ```

Comment (billyvg) [reply text/0129-video-replay-envelope.md:30]: That sounds good -- the backend will need to ensure we can reference the mp4 via `segment_id`.

I would suggest using `type = 5` so that we don't need to override rrweb's enum:

```
{
   "type": 5,
   "data": {
      "tag": "video",
      "payload": {
        "segment_id": 4,
        "duration": 5000
      }
    }
}
```

Comment (romtsn) [reply text/0129-video-replay-envelope.md:30]: what about the existing `segment_id` on the [ReplayEvent](https://getsentry.github.io/relay/relay_event_schema/protocol/struct.Replay.html#structfield.segment_id) type? Is this something different?

Comment (vaind) [reply text/0129-video-replay-envelope.md]: FYI I've updated the markdown with the changes discussed in this thread.

File: text/0129-video-replay-envelope.md [hunk 3]
```diff
@@ -0,0 +1,62 @@
...
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:Attachment
+
+- From the SDK, we should be able to use the existing envelope item type, [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
+- Additionally, it would be accompanied with replay envelope item type, [`ReplayEvent`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113) in RRWeb JSON format.
```
cmanallen, line 31: Do not send RRWeb formatted JSON on the `ReplayEvent` envelope type.  Please refer to the protocol defined for the `ReplayEvent` here: https://github.com/getsentry/relay/blob/8411b1dfaaee645658cd5e9d5428058f2b19384f/relay-event-schema/src/protocol/replay.rs#L26

The information on this event is required to enable search and sort.

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,62 @@
...
+We need this to to capture replays on platforms where it's not possible/feasible to produce an HTML DOM (i.e. the native format supported by RRWeb). For example: mobile apps.
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:Attachment
+
+- From the SDK, we should be able to use the existing envelope item type, [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
```
cmanallen, line 30: The `ReplayRecording` envelope type should contain RRWeb formatted JSON.  It has the following components: a headers JSON object, a new line character, RRWeb recording JSON.

The headers object must contain a segment_id key but can be extended if your use case requires it:

```json
{"segment_id": 12}
```

If you're going to send binary video data on this event I would recommend extending the headers with a `type` key and a `version` key.  E.g. `type: video` and `version: 2`.  You will also need to track the offsets.  The RRWeb JSON and the video offsets.  E.g. `rrweb-offsets: [22, 3461]` and `video-offsets: [3462, 9555]`.

The video can be appended to the end of the file (immediately after the RRWeb JSON).

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,62 @@
...
+- Additionally, it would be accompanied with replay envelope item type, [`ReplayEvent`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113) in RRWeb JSON format.
+- To link to the video file in the replay JSON, we could use a [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16). The data would contain `segment_id` referencing the replay video recording attachment.
```
cmanallen, line 32: I don't think the segment_id is necessary.  We already store the rrweb JSON by its segment_id so we know what it is internally.

If we imagine two api resources, one for videos and one for rrweb json they would return:

[video, video, video], [rrweb, rrweb, rrweb].

They're order so you can zip them for the video information [(video, rrweb), (video, rrweb), (video, rrweb)].

text/0129-video-replay-envelope.md
cmanallen: Alternatively we can introduce a third envelope type and process the video independently of the RRWeb JSON.  This has pros and cons but it preserves backwards compatibility on the recording endpoint and doesn't overly complicate that consumer with a bunch of forking attributes (type, version).

cmanallen (changes requested):

text/0129-video-replay-envelope.md
vaind: Okay, how **do** we send the actual replay JSON data then? This thing ought to have some docs...

text/0129-video-replay-envelope.md
vaind: OK, I've changed the RFC to propose a new envelope item type

text/0129-video-replay-envelope.md
vaind: OK, changed.

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,69 @@
...
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
+  The envelope item would consist of a single header line (JSON), followed by a new line and the raw video data.
+  - The header should contain at least the following metadata:
```
cmanallen, line 32: Let's also add `segment_id` to the headers.

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,69 @@
...
+- Additionally, it would be accompanied by an item [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113), containing a header, e.g. `{"segment_id": 12}`, followed by a new line and the RRWeb JSON.
+  - The RRWeb JSON should start with an event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), containing the type `video`. If there's other data the UI needs, we can add it alongside the `type` to the `data` field. Because there's only a single `ReplayVideo` sent with a single `ReplayRecording`, there's a one-to-one mapping without further details necessary in the actual RRWeb JSON.
+
+    ```json
+    {
+     "type": 5,
+     "data": {
+        "type": "video",
+      }
+    }
```
cmanallen, line 51: @billyvg Does this format work for the player?  Video and RRWeb information would be ordered so if you're processing the RRWeb at index `1` you would get the video information from index `1` and play it.

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,69 @@
...
+  - The header should contain at least the following metadata:
+    - the raw size of the video data
+    - the duration of the video in milliseconds
+    - the encoder format
+    - the container format
+    - the video resolution
+  - Additionally, it may be useful to include more information, such as:
+    - the number of frames in the video
+    - the frame rate type (constant, variable)
+    - the frame rate value
```
cmanallen, line 41: It might make more sense to put these headers (other than segment_id) into the RRWeb JSON.  For example:

```json
{
  "type": 5,
  "data": {
    "type": "video",
    "size": 3440,
    "duration": 5000,
    "encoding": "whatever",
    "container": "whatever",
    "v-size": 1920,
    "h-size": 1080,
    "frame_count": 50,
    "frame_rate_type": "constant",
    "frame_rate": 10
  }
}
```

This simplifies the implementation on the backend and probably the frontend.  We can keep the video blob in its own file and the RRWeb in its own file.  If we put the video metadata into the headers of the video request we would have to either store it in its own file or encode the video to a b64 string and write it into the headers JSON file.  Which would be inefficient.

cmanallen (changes requested):

text/0129-video-replay-envelope.md
cmanallen: The key names and their values are just placeholders.  Replace them with whatever makes the most sense to the SDK and player implementations.

vaind: > It might make more sense to put these headers (other than segment_id) into the RRWeb JSON. For example:

The idea to put them into the envelope was to allow us to easily work with the data in relay and elsewhere, without relying on RRWeb JSON format, which may change in the future.

text/0129-video-replay-envelope.md
vaind: added

text/0129-video-replay-envelope.md
billyvg: Is it guaranteed that every segment will always contain video?

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
+  The envelope item would consist of a single header line (JSON), followed by a new line and the raw video data.
+  - The header should contain at least the following metadata:
+    - replay segment ID
+    - the raw size of the video data
+    - the duration of the video in milliseconds
+    - the encoder format
```
romtsn, line 36: should we have a predefined set of those (especially the encoder one)? It may vary between platforms

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,78 @@
...
+        "encoding": "whatever",
+        "container": "whatever",
+        "v-size": 1920,
+        "h-size": 1080,
+        "frame_count": 50,
+        "frame_rate_type": "constant|variable",
+        "frame_rate": 10
+      }
+    }
+    ```
```
billyvg, line 61: For `customEvent`: https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L53-L59

This should be

```suggestion
    ```json
    {
      "type": 5,
      "data": {
        "tag": "video",
        "payload": {
          "size": 3440,
          "duration": 5000,
          "encoding": "whatever",
          "container": "whatever",
          "v-size": 1920,
          "h-size": 1080,
          "frame_count": 50,
          "frame_rate_type": "constant|variable",
          "frame_rate": 10
        }
      }
    }
    ```
```

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,78 @@
...
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
+  The envelope item would consist of a single header line (JSON), followed by a new line and the raw video data.
```
romtsn, line 31: Are these values gonna be used, or they are here "just in case"? I think we should spec them out as well, if they are gonna be needed on backend/frontend

text/0129-video-replay-envelope.md
romtsn: there should also be a `timestamp` alongside the `type` for each event, from what I grasped, right?

romtsn: @billyvg would this be sufficient as start_timestamp for you, or we'd have to supply it somewhere else?

text/0129-video-replay-envelope.md
vaind: added it there

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,78 @@
...
+We need this to to capture replays on platforms where it's not possible/feasible to produce an HTML DOM (i.e. the native format supported by RRWeb). For example: mobile apps.
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) to transport the video data.
```
romtsn, line 30: ok, I guess imma little bit lost - I thought we'd keep the [Replay](https://github.com/getsentry/relay/blob/8411b1dfaaee645658cd5e9d5428058f2b19384f/relay-event-schema/src/protocol/replay.rs#L26) type, but it looks like we're replacing it? If we're replacing it, shouldn't we derive most of the fields from there?  (cc @cmanallen). At the very least the `event_id`,`replay_id`,`timestamp` fields should be in here, right?

romtsn:
not sure if this should go into this PR, but what structure the `video_url` field would eventually be part of (for frontend)?

text/0129-video-replay-envelope.md
vaind: added a timestamp to the payload

text/0129-video-replay-envelope.md
cmanallen: I would think it should.  If there's a case where you want to send breadcrumbs without a video the API could backfill by sending empty bytes where it could not find video blob data.  When the client concats them the empty bytes would do nothing.

text/0129-video-replay-envelope.md
vaind: `ReplayVideo` comes in addition to `Replay`

vaind: I'll clarify in the markdown

text/0129-video-replay-envelope.md
vaind: I'd say so but that's an implementation detail (as much as this RFC is concerned), IMO

vaind: in other words, it can be agreed upon later on and changed pretty quickly on each end (SDK, UI) while the rest of the stuff is already implemented.

text/0129-video-replay-envelope.md
billyvg: yeah that'll be sufficient

text/0129-video-replay-envelope.md
cmanallen: Segment-id is needed in the header so I know what to name the file.  All the other head examples should be moved to the RRWeb JSON and can be kept just-in-case.

vaind: i've updated the markdown in the meantime

text/0129-video-replay-envelope.md
cmanallen: We should not remove the `Replay` type or the `ReplayRecording` type.  We should add a `ReplayVideo` type as an additional envelope item.  So three items total in the request.

text/0129-video-replay-envelope.md
vaind: I'd say the contract could be that there's always exactly one element with the video in each `ReplayRecording`. I've updated the proposal

text/0129-video-replay-envelope.md
vaind: It's currently there in the updated json header example

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+  The envelope item would consist of a single header line (JSON), followed by a new line and the raw video data.
+  - The header should contain at least the following metadata:
+
+    ```json
+    {
+      "segment-id": 4,
+      "size": 3440,
+      "duration": 5000,
+      "encoding": "whatever",
+      "container": "whatever",
```
cmanallen, line 41: Only put things in the headers object that are useful for ingestion.  If the player needs it, put it in the RRWeb event.  As it stands only the `segment_id` is planned for use on ingest.  The rest will be discarded.

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+We need this to to capture replays on platforms where it's not possible/feasible to produce an HTML DOM (i.e. the native format supported by RRWeb). For example: mobile apps.
+
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope with the following items: `Replay`, `ReplayVideo` and `ReplayRecording`.
```
cmanallen, line 30: @billyvg How would the player like to be notified that it should download video data?  A `type` value on the replay?  The video events in the RRWeb?  Infer it from the replay's platform?  Something else?

cmanallen (changes requested):

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+      "data": {
+        "tag": "video",
+        "payload": {
+          "segment-id": 4,
+          "size": 3440,
+          "duration": 5000,
+          "encoding": "whatever",
+          "container": "whatever",
+          "v-size": 1920,
+          "h-size": 1080,
```
romtsn, line 63:
```suggestion
          "height": 1920,
          "width": 1080,
```
can those be just like that?

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+      "container": "whatever",
+    }
+    ```
+
+- Additionally, it would be accompanied by an item [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113), containing a header, e.g. `{"segment_id": 12}`, followed by a new line and the RRWeb JSON.
+  - The RRWeb JSON must start with an event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), containing the type `video`.
+  There must only be a single video custom event in the RRWeb JSON.
+  If there's other data the UI needs, we can add it alongside the `type` to the `data` field. Because there's only a single `ReplayVideo` sent with a single `ReplayRecording`, there's a one-to-one mapping without further details necessary in the actual RRWeb JSON.
+
+    ```json
```
romtsn, line 50: should we snake_case everything in here? we usually use it for json props

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send a new envelope with the following items: `Replay`, `ReplayVideo` and `ReplayRecording`.
+- The newly introduced item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20) is used to transport the video data.
+  The envelope item would consist of a single header line (JSON), followed by a new line and the raw video data.
+  - The header should contain at least the following metadata:
+
+    ```json
+    {
+      "segment-id": 4,
```
romtsn, line 37: @cmanallen just to confirm: this `segment_id` and the ones in the `Replay` item and `ReplayRecording`'s payload should be the same one, correct? Shall we also document it here for transparency?

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,89 @@
...
+      "size": 3440,
+      "duration": 5000,
+      "encoding": "whatever",
+      "container": "whatever",
+    }
+    ```
+
+- Additionally, it would be accompanied by an item [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113), containing a header, e.g. `{"segment_id": 12}`, followed by a new line and the RRWeb JSON.
+  - The RRWeb JSON must start with an event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), containing the type `video`.
```
romtsn, line 46: @billyvg from your rrweb explanation, we also need the [metaEvent](https://github.com/rrweb-io/rrweb/blob/master/packages/types/src/index.ts#L44-L51) first, right? So the recording json will look like this:
```json
 [
    {
      "type": 4,
      "data": {
        "href": "",
        "height": 1920,
        "width": 1080
      },
      "timestamp": 1681846559381
    },
    {
      "type": 5,
      "timestamp":1681846559381,
      "data": {
        "tag": "video",
        "payload": {
          "segment-id": 4,
          "size": 3440,
          "duration": 5000,
          "encoding": "whatever",
          "container": "whatever",
          "height": 1920,
          "width": 1080,
          "frame-count": 50,
          "frame-rate-type": "constant|variable",
          "frame-rate": 10,
        }
      }
    },
    // .. breadcrumbs and friends
]
```

and this is for each segment, correct?

text/0129-video-replay-envelope.md
billyvg: @cmanallen Thinking of using the rrweb video event

text/0129-video-replay-envelope.md
billyvg: @romtsn For web replays, it only sends it on the initial DOM snapshot and I think when the view port changes. So only for first segment, but it won't break the player if it's sent with every segment

romtsn: > when the view port changes

ok, that could also be the case on mobile (and also orientation change). We'll actually have to finish the current segment and start a new one in that case since the video size will change. 

So we send metaEvent for first segment and whenever the view port or orientation changes, can we add this pls @vaind?

text/0129-video-replay-envelope.md
vaind: OK, removing for now

text/0129-video-replay-envelope.md
vaind: actually, looking at rrweb stuff it's all camelCased, so changing accordingly.

thanks for the wakup call

text/0129-video-replay-envelope.md
vaind: I've updated the spec. Also added a note that the `meta` dimensions may be different than the video dimensions in case a capture is made only on a part of the view. In such cases, the video event will have non-zero top/left fields. This is relevant for Flutter where we'd be capturing whatever user wrapped in a custom widget. It doesn't have to be the top of the tree.

text/0129-video-replay-envelope.md
romtsn: @vaind you need size in bytes of a video here, right?

text/0129-video-replay-envelope.md
romtsn: thanks! @cmanallen I've checked the item headers once again, e.g. for an attachment they look like that:
```json
{"type":"attachment","length":10,"content_type":"text/plain","filename":"hello.txt"}\n
```
does it make sense to adhere to the common pattern here and do something like this (`type` is actually required afaik):
```json
{"type":"replay_video","length":1000,"content_type":"video/mp4","segment_id":4}\n
```
`filename` is probably not needed if we name it by the segment_id, right? (4.mp4)

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,104 @@
...
+        "height": 1920,
+        "width": 1080
+      }
+    }
+    ```
+
+  > Note: these dimensions may be different than the video dimensions in case only part of the screen is captured.
+    In that case, the following video event will have non-zero `data.payload.left` & `data.payload.top` fields (see below).
+
+- The RRWeb JSON must contain a single event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), with `data.tag == 'vido'`.
```
romtsn, line 59:
```suggestion
- The RRWeb JSON must contain a single event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), with `data.tag == 'video'`.
```

text/0129-video-replay-envelope.md
cmanallen: They _should_ be the same but if you sent a segment 2 video with a segment 3 rrweb event nothing would break.  So long as the rrweb video event on segment 2 accurately reflects the segment 2 video.  The events are not dependent on one another for processing.  They are just associated to the same "segment_id" metadata.

text/0129-video-replay-envelope.md
cmanallen: > filename is probably not needed if we name it by the segment_id, right?

Correct.

> does it make sense to adhere to the common pattern here and do something like this (type is actually required afaik):

No I don't think its necessary to follow that pattern.  They accept those headers because they want to store them.  I have no intention of storing that metadata anywhere except in the RRWeb JSON.  If there's something about these headers you'd like to preserve, you should put it in the RRWeb JSON.  You can put anything you want there; it doesn't concern me at all.  I would recommend putting more than you need in there.  It can always be removed later.

cmanallen (approved):
This looks good.  So long as the player is satisfied I'm happy with this design.  I would give @billyvg final sign off.

text/0129-video-replay-envelope.md
vaind: these are not part of the envelope item content (what is described here), but part of the envelope item specification.

romtsn: > The envelope item would consist of a single header line (JSON)

but this is the item header or am I wrong?

vaind: I don't think that's what the ReplayRecording header does. AFAIU It is part of the envelope item payload. I'm not saying that's the right way to do it or not, just how I understood it is done for that envelope item and proposed to do the same for the new one.

vaind: But I may be wrong, of course :)

romtsn: ah, got it now, thanks. So we just follow the usual pattern for the item header then (I guess we don't need to document it here, or do we?).

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,104 @@
...
+  - The header should contain at least the following metadata: needed to ingest the item.
+
+    ```json
+    {
+      "segment_id": 4,
+    }
+    ```
+
+- Additionally, it would be accompanied by an item [`ReplayRecording`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L113), containing a header, e.g. `{"segment_id": 12}`, followed by a new line and the RRWeb JSON.
+- The RRWeb JSON must start a single event of type [`EventType.Meta`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), with viewport (screen) dimensions.
```
romtsn, line 42:
```suggestion
- The RRWeb JSON may start with a single event of type [`EventType.Meta`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), with viewport (screen) dimensions. The `EventType.Meta` event is usually present in the first segment of a replay, or whenever the view port or screen orientation changes.
```
does this make sense to word like this? Or do we want to send this meta event every time?

cmanallen: From my understanding of the RRWeb specification that makes sense.  I would wait for billy to confirm.

billyvg: Yeah this change makes more sense

text/0129-video-replay-envelope.md
vaind: > I guess we don't need to document it here, or do we?

It evidently is confusing so I'll clarify

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,104 @@
...
+        "width": 1080
+      }
+    }
+    ```
+
+  > Note: these dimensions may be different than the video dimensions in case only part of the screen is captured.
+    In that case, the following video event will have non-zero `data.payload.left` & `data.payload.top` fields (see below).
+
+- The RRWeb JSON must contain a single event of type [`EventType.Custom`](https://github.com/rrweb-io/rrweb/blob/8aea5b00a4dfe5a6f59bd2ae72bb624f45e51e81/packages/types/src/index.ts#L8-L16), with `data.tag == 'video'`.
+  This event must come at the second position in the array, right after the `EventType.Meta` event.
```
romtsn, line 60:
```suggestion
  This event must come at the first position in the array, or at the second position if the `EventType.Meta` event is present.
```

vaind:
@romtsn I think you're the last one to approve this, let me know if there're more changes you'd like to see, otherwise it's good to go

text/0129-video-replay-envelope.md
```diff
@@ -0,0 +1,86 @@
...
+<!-- # Supporting Data -->
+<!-- Metrics to help support your decision (if applicable). -->
+
+# Options Considered
+
+## Using a video, with EnvelopeItem:ReplayVideo
+
+- From the SDK, we would send an envelope with the following items: `Replay`, `ReplayVideo` and `ReplayRecording`.
+- The newly introduced item type, [`ReplayVideo`](https://github.com/getsentry/relay/blob/5fd3969e88d3eea1f2849e55b61678cac6b14e44/relay-server/src/envelope.rs#L115C5-L115C20), is used to transport the video data.
+  The envelope item payload would consist of a single header line (`{"segment_id": 4}`), followed by a new line character (`\n`), followed by the raw video data.
```
bruno-garcia, line 32: don't we need the payload size here too? That's part of the envelope item protocol already though so omitted because of that?

text/0129-video-replay-envelope.md
vaind: thanks, I've updated this with with the numbers from previous tests.

text/0129-video-replay-envelope.md
vaind: It's not necessary. there's nothing more after the binary data, so it's everything after the second `\n` character until the end of stream.

bruno-garcia (approved):

PR #131, opened 2024-02-08
rfc(feature): Pass Native SDKs Spans to Hybrid

philipphofmann:
This RFC aims to find a solution for passing auto-instrumented native SDK spans up to the hybrid SDKs.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/rfcfeature-native-to-hybrid-sdk-spans-passing/text/0131-pass-native-sdk-spans-to-hybrid.md)


text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+
+This RFC aims to find a solution for passing auto-instrumented native SDK spans up to the hybrid SDKs.
+
+# Motivation
+
+The native SDKs instrument file IO operations, DB queries, HTTP requests, and more, but this
+information doesn't end up in the hybrid transactions or anywhere else. Our hybrid SDKs instrument
+most of the mentioned operations on the hybrid layer, but they lack information on how long the
+actual native operations take. It could happen that one DB query from the native layer gets
+converted into multiple native DB queries. Such information would be insightful for our users.
```
brustolin, line 17: Is this correct?
```suggestion
actual native operations take. It could happen that one DB query from the hybrid layer gets
converted into multiple native DB queries. Such information would be insightful for our users.
```

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+1. This option works well with idle transactions and `waitForChildren` because these concepts need to
+know when spans start and finish, but this is a weak argument as with single-span ingestion, this
+concept will become obsolete.
+2. Works well with single-span ingestion, as the hybrid SDKs get callbacks for each span.
+3. Complete flexibility for hybrid SDKs. Hybrid SDKs have full control over how they want to deal with
+native-span information.
+
+### Cons <a name="option-3-cons"></a>
+
+1. Each span requires two calls across the layers, which could negatively impact performance.
```
brustolin, line 163: What prevent us from having only two callbacks and having arguments to control what kind o span was started/finished and its properties?

brustolin (commented):
I like option number 3, it feels more versatile.

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: Good catch thanks

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: I'm sorry, I don't fully understand. Do you maybe mean one callback? 

> arguments to control what kind o span was started/finished and its properties

An interesting idea 🤔. How would that work? What are the benefits of that?

brustolin: You mentioned "each spans requires two" I guess http spans requires two, db spans requires two, file I/O requires two. 

I suggested two callbacks for all of this, along with "hints" to pass information such as "this is an HTTP request to URL X". However, we could even just use one callback and another argument to indicate the start or finish.

philipphofmann: Ah, yes, obviously, this makes sense. Two or one callback makes sense.

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+## Option 3: Native Callbacks<a name="option-3"></a>
+
+The native layers offer callbacks whenever they start and finish auto-instrumented spans and don't
+keep track of spans. For example, the HTTP request instrumentation offers two callbacks: one for
+starting the HTTP request and another for finishing the HTTP request. The hybrid SDKs then subscribe
+to these callbacks and create the spans themselves.
+
+### Pros <a name="option-3-pros"></a>
+
+1. This option works well with idle transactions and `waitForChildren` because these concepts need to
```
kahest, line 154: is this a clear advantage over the aggregate callback when all spans are finished outlined for option 2?

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+know when spans start and finish, but this is a weak argument as with single-span ingestion, this
+concept will become obsolete.
+2. Works well with single-span ingestion, as the hybrid SDKs get callbacks for each span.
```
kahest, line 157: we recently discussed that to calculate exclusive time in single-span ingestion, all child spans need to be finished. does that mean that this option would provide individual callbacks for each span, but they would likely arrive at the same time because we'd need to calculate exclusive time and that's only possible once all (related) spans finished?

kahest (commented):
This is an intriguing topic - I added some clarifying questions, feel free to point out if/when something is not relevant for this RFC

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+1. It should be compatible with both transactions and single-span ingestion. It must be compatible
+with single-span ingestion. Not being compatible with transactions might be acceptable, as we have
+plans to remove them.
+2. It must work with transactions bound to the scope and active spans. We don't have to worry about
+multiple transactions or spans running in parallel, as on mobile, the SDKs can never tell to which
+transaction bound to the scope or active span an auto-instrumented span belongs. Therefore, they
+attach it to the transaction bound to the scope or the active span.
+3. Native auto-instrumented spans must go through the hybrid layer because of span filtering and sampling,
+meaning native SDKs must not send spans themselves.
+4. The native SDKs must support attaching screen frame data to the spans.
```
krystofwoldrich, line 48: Likely better if the frame data are already attached, but since the hybrid SDKs will have to attach the frame data to hybrid spans it would be possible to also attach the frame data to the native span in the hybrid layer.

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+# Options Considered
+
+## Option 1: Self Clearing Cache With Hooks <a name="option-1"></a>
+
+The native SDKs implement a self-clearing cache for their auto-instrumented spans with hooks to
+start and stop the cache. Whenever the native SDKs add a span to their cache, they discard all
+spans older than a configurable threshold. To make cache clearing efficient, the SDKs keep the spans
+in a data structure ordered by span timestamp and use a binary search to find the last index of the
+range of spans to delete, which makes inserting a span `O(log n)`. Furthermore, the native SDKs
+offer a callback when all ongoing spans are finished to support `waitForChildren`.
```
krystofwoldrich, line 59: > to support `waitForChildren`
The `waitForChildren` is from the Hybrid SDKs, is that right?

text/0131-pass-native-sdk-spans-to-hybrid.md
krystofwoldrich: As I understand it, this would be only for the top-level spans, if the native DB query has children those would not be exposed.

kahest: ah that makes sense 👍

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+    And doesn't send its native spans directly to Sentry
+    And the hybrid SDK adds all native spans to its SpansAggregator
+
+Scenario: Active span, native SpansAggregator flushes
+    Given an active span
+    When the native SpansAggregator has to flush all its spans
+    Then native SDK pushes all native spans up to the hybrid SDK
+    And the hybrid SDK filters for spans belonging to the active span 
+    And adds these spans to the active span
+    And adds the non matching spans to the hybrid SpansAggregator
```
krystofwoldrich, line 109: Does option 2 counts with single-span ingestion, if we consider the active span being a transaction, what would happen with the non-matched spans?

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,174 @@
...
+    When the hybrid SpansAggregator has to flush all its spans
+    Then it does not pull spans from the native SpansAggregator
+    And sends all hybrid spans from the SpansAggregator to Sentry
+
+Scenario: No active span, hybrid SpansAggregator flushes
+    Given no active span
+    When the hybrid SpansAggregator has to flush all its spans
+    Then it pulls spans from the native SpansAggregator
+    And adds these to the hybrid SpansAggregator
+    And sends native and hybrid spans to Sentry
```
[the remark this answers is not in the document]
krystofwoldrich: I have the same question, I don't see why would we not pull native spans when there is an active hybrid span?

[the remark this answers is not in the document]
philipphofmann: Good catch. That doesn't make sense. I updated the scenario.

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: The auto instrumented spans on Cocoa we are interested in this RFC aren't nested. DB queries, HTTP requests, and file IO operations don't have nesting. We can nest them on the hybrid layer.

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: I think it's simpler to attach the data on the native layer already as that's the default behavior. Not adding it will be more work.

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: In that case, we would throw them away as we do right now.

text/0131-pass-native-sdk-spans-to-hybrid.md
philipphofmann: Yes

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,181 @@
...
+
+# Motivation
+
+The native SDKs instrument file IO operations, DB queries, HTTP requests, and more, but this
+information doesn't end up in the hybrid transactions or anywhere else. Our hybrid SDKs instrument
+most of the mentioned operations on the hybrid layer, but they lack information on how long the
+actual native operations take. It could happen that one DB query from the hybrid layer gets
+converted into multiple native DB queries. Such information would be insightful for our users.
+Furthermore, there could be native-only operations that our hybrid SDKs currently don't capture. Therefore,
+we want to pass native spans up to the native layers so users know what's happening on the native layer.
```
bruno-garcia, line 19: in a world of ingesting only spans, do we need to pass them across layers? 
Would it dumping straight to the transport be enough while having them on the same trace?

philipphofmann: That's an excellent question. We want to establish a parent-child relationship between some spans, which works best when passing spans from native to RN because, on the backend, it's tricky if the spans don't arrive there in the same envelope. I will update the RFC to explain this.

philipphofmann: I added your suggestion as option 4. Feel free to directly commit pros and cons, @bruno-garcia.

kahest (approved):
Thanks everyone for contributing!

text/0131-pass-native-sdk-spans-to-hybrid.md
```diff
@@ -0,0 +1,256 @@
...
+- RFC Approver: [Karl Heinz Struggl](https://github.com/kahest)
+
+# Summary
+
+This RFC aims to find a solution for passing auto-instrumented native SDK spans up to the hybrid SDKs.
+
+## Option Chosen
+
+On 2024-02-28, we decided to move forward with [Option 2: Spans Aggregator](#option-2) because of
+its the following reasons:
```
kahest, line 15:
```suggestion
the following reasons:
```

kahest (approved):

PR #133, opened 2024-03-04
docs: Clarify span op/description for metric correlation in span creating mode

markushi:
Clarify span op and description, fix example.

Based on 
 * python impl: https://github.com/getsentry/sentry-python/blob/0901953c93071e858f4da67c1e864766ae19c002/sentry_sdk/metrics.py#L800C48-L800C61
 * ruby impl: https://github.com/getsentry/sentry-ruby/pull/2255/files#diff-44d204371288f9a84dd2ed1c054a2b09d85388267ba249b83ca08ea251b1e6f9R40

[Rendered RFC](https://github.com/getsentry/rfcs/blob/markushi/metric-span-correlation-docs/text/0123-metrics-correlation.md)


sl0thentr0py (approved):

philipphofmann (approved):
Good catch 🚀

PR #136, opened 2024-06-04
rfc(decision): Mobile - Tracing Without Performance V2

philipphofmann:
This RFC aims to find a strategy to update the  so traces don’t reference hundreds of unrelated
events.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/mobile-tracing-without-performance-v-2/text/0136-mobile-tracing-without-performance-v-2.md)

### DIFF
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-This RFC aims to find a strategy to update the  so traces don’t reference hundreds of unrelated
+This RFC aims to find a strategy to define a better lifetime for traces so they don’t reference hundreds of unrelated
 events.
 
 # Motivation
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -27,7 +27,7 @@ On mobile, most users interact purely with the static API, which holds a referen
 Hub and Scope. Therefore, mobile SDKs create a `PropagationContext` with `traceId` and `spanId`
 during initialization, and these usually persist for the entire lifetime of the app. Mobile
 SDKs prefer the `traceID` of transactions bound to the scope over the `PropagationContext`. So
-when performance is disabled, or no transition is bound to the scope, mobile SDKs use the same
+when performance is disabled, or no transaction is bound to the scope, mobile SDKs use the same
 `traceId` and `spanId` for all captured events. This can lead to traces with hundreds of
 unrelated events confusing users. JS addressed this recently by updating the `PropagationContext`
 based on routes, see [Ensure browser traceId lifetime works as expected](https://github.com/getsentry/sentry-javascript/issues/11599).
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -34,7 +34,7 @@ based on routes, see [Ensure browser traceId lifetime works as expected](https:/
 
 # Options Considered
 
-## Option 1: Update `PropagationContext` based on screens <a name="option-1"></a>
+## Option 1: Update `PropagationContext` based on screens/routes <a name="option-1"></a>
 
 Mobile SDKs base the lifetime of the `traceId` of the `PropagationContext` on screens/routes,
 which is similar to a route on JavaScript. Mobile SDKs already report the screen name automatically
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -1,0 +1,35 @@
+- Start Date: 2024-06-04
+- RFC Type: decision
+- RFC PR: https://github.com/getsentry/rfcs/pull/136
+- RFC Status: draft
+
+# Summary
+
+One paragraph explanation of the feature or document purpose.
+
+# Motivation
+
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+# Background
+
+The reason this decision or document is required. This section might not always exist.
+
+# Supporting Data
+
+[Metrics to help support your decision (if applicable).]
+
+# Options Considered
+
+If an RFC does not know yet what the options are, it can propose multiple options. The
+preferred model is to propose one option and to provide alternatives.
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented.
+
+# Unresolved questions
+
+- What parts of the design do you expect to resolve through this RFC?
+- What issues are out of scope for this RFC but are known?
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -54,7 +54,11 @@ it based on routes, so it should be easy to implement for React-Native.
 
 ### Cons <a name="option-1-cons"></a>
 
-1. It doesn’t work well for declarative UI frameworks as Jetpack Compose and SwiftUI for which the
+1. For single-screen applications such as social networks, the lifetime of a trace could still be
+long, and multiple unrelated events could be mapped to one trace.
+2. For applications running for a long time in the background, such as running apps, the lifetime of
+a trace could still be long, and multiple unrelated events could be mapped to one trace.
+3. It doesn’t work well for declarative UI frameworks as Jetpack Compose and SwiftUI for which the
 SDKs can’t reliably automatically detect when apps load a new screen.
 
 # Drawbacks
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -48,7 +48,9 @@ foreground after being in the background for longer than 30 seconds, which is th
 mobile SDKs use for determining the end of a session, mobile SDKs renew the `traceId` of the
 PropagationContext. If the app stays in the background for shorter or equal to 30 seconds,
 mobile SDKs must not renew the `traceId` of the PropagationContext when the app moves again to
-the foreground.
+the foreground. When a span starts, SDKs should use the traceID on the PropagationContext, but
+SDKs make an exception to this rule to map all spans related to a screen/route to the same traceID.
+For example, when a span leads to a new screen, SDKs should use the traceID of the new screen.
 
 ### Pros <a name="option-1-pros"></a>
 
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -44,7 +44,11 @@ starfish module uses. Whenever the screen name changes automatically or with a y
 [manual API](https://www.notion.so/sentry/Specs-Screens-API-084d773272f24f57aeb622c07619264e),
 mobile SDKs must renew the `traceId` of the `PropagationContext`. The screen load transaction
 and subsequent events on the same screen must use the same `traceId`. When the app moves to the
-background, mobile SDKs also update the `traceId` of the PropagationContext.
+foreground after being in the background for longer than 30 seconds, which is the same approach
+mobile SDKs use for determining the end of a session, mobile SDKs renew the `traceId` of the
+PropagationContext. If the app stays in the background for shorter or equal to 30 seconds,
+mobile SDKs must not renew the `traceId` of the PropagationContext when the app moves again to
+the foreground.
 
 ### Pros <a name="option-1-pros"></a>
 
--- a/text/0136-mobile-tracing-without-performance-v-2.md
+++ b/text/0136-mobile-tracing-without-performance-v-2.md
@@ -5,31 +5,62 @@
 
 # Summary
 
-One paragraph explanation of the feature or document purpose.
+This RFC aims to find a strategy to update the  so traces don’t reference hundreds of unrelated
+events.
 
 # Motivation
 
-Why are we doing this? What use cases does it support? What is the expected outcome?
+On mobile, traces can have hundreds of unrelated events caused by the possibly never-changing
+required for tracing without performance. This occurs mostly when users don’t have performance
+enabled.
 
 # Background
 
-The reason this decision or document is required. This section might not always exist.
+In the summer of 2023, all mobile SDKs implemented [Tracing without performance](https://www.notion.so/Tracing-without-performance-efab307eb7f64e71a04f09dc72722530?pvs=21),
+see also [team-sdks GH issue](https://github.com/getsentry/team-sdks/issues/5).
+The goal of this endeavor was to
 
-# Supporting Data
+> always have access to a trace and span ID, add a new internal `PropagationContext` property to the
+> scope, an object holding a `traceId` and `spanId`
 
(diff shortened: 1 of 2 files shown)


text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,70 @@
...
+2. Works for spans first, as all spans get added to one trace per screen.
+
+### Cons <a name="option-1-cons"></a>
+
+1. For single-screen applications such as social networks, the lifetime of a trace could still be
+long, and multiple unrelated events could be mapped to one trace.
+2. For applications running for a long time in the background, such as running apps, the lifetime of
+a trace could still be long, and multiple unrelated events could be mapped to one trace.
+3. It doesn’t work well for declarative UI frameworks as Jetpack Compose and SwiftUI for which the
+SDKs can’t reliably automatically detect when apps load a new screen.
```
kahest, line 62: For JPC there might be ways to make the 80% work with some KCP magic I guess (@markushi @romtsn wdyt?). But SwiftUI is a closed box and we'll need to mitigate this by providing clear ways/instructions on where to e.g. wrap something with a SDK method and/or manual API usage

kahest (commented):
Left some first thoughts. In general I don't see a lot of different options here - IMO it's mostly a combination of screen awareness, foreground/background handling and maybe some timeouts. So pretty much exactly your option 1

bitsandfoxes:
What about additionally tying it to foreground/background behaviour like sessions? It doesn't have to be only one rule, right? It could be a set of rules and conditions that lead to the creation of a new `traceId`.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,70 @@
...
+1. Similar to [JavaScript]((https://github.com/getsentry/sentry-javascript/issues/11599)) updating
+it based on routes, so it should be easy to implement for React-Native.
+2. Works for spans first, as all spans get added to one trace per screen.
+
+### Cons <a name="option-1-cons"></a>
+
+1. For single-screen applications such as social networks, the lifetime of a trace could still be
+long, and multiple unrelated events could be mapped to one trace.
+2. For applications running for a long time in the background, such as running apps, the lifetime of
+a trace could still be long, and multiple unrelated events could be mapped to one trace.
```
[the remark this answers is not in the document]
philipphofmann: Hmm, if the user quickly puts the app into the background to do something in another app and navigates back, I guess the traceID should actually stay the same. So, always renewing the traceID when the app goes to the background doesn't work. We could use the same logic as we already do for sessions to only start a new traceID if the app was in the background for 30s or longer.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,74 @@
+- Start Date: 2024-06-04
+- RFC Type: decision
+- RFC PR: https://github.com/getsentry/rfcs/pull/136
+- RFC Status: draft
+
+# Summary
+
+This RFC aims to find a strategy to update the  so traces don’t reference hundreds of unrelated
```
markushi, line 8:
```suggestion
This RFC aims to find a strategy to define a better lifetime for traces so they don’t reference hundreds of unrelated
```

text/0136-mobile-tracing-without-performance-v-2.md
markushi: I think the easiest way to get started is by providing a wrapper ~`SentryScreen() { <Composable> }`, in a later version we could have a `@SentryScreen` method annotation which utilizes KCP to auto-wrap it.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,74 @@
...
+1. For single-screen applications such as social networks, the lifetime of a trace could still be
+long, and multiple unrelated events could be mapped to one trace.
+2. For applications running for a long time in the background, such as running apps, the lifetime of
+a trace could still be long, and multiple unrelated events could be mapped to one trace.
+3. It doesn’t work well for declarative UI frameworks as Jetpack Compose and SwiftUI for which the
+SDKs can’t reliably automatically detect when apps load a new screen.
+
+# Drawbacks
+
+Please add drawbacks here if you can think of any.
```
markushi, line 70: Any UI spans which operate across screens (e.g. a tap interaction triggering a new screen load) need to be assigned to either the old or the new screen `traceId`.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,74 @@
...
+during initialization, and these usually persist for the entire lifetime of the app. Mobile
+SDKs prefer the `traceID` of transactions bound to the scope over the `PropagationContext`. So
+when performance is disabled, or no transition is bound to the scope, mobile SDKs use the same
+`traceId` and `spanId` for all captured events. This can lead to traces with hundreds of
+unrelated events confusing users. JS addressed this recently by updating the `PropagationContext`
+based on routes, see [Ensure browser traceId lifetime works as expected](https://github.com/getsentry/sentry-javascript/issues/11599).
+
+# Options Considered
+
+## Option 1: Update `PropagationContext` based on screens <a name="option-1"></a>
```
markushi, line 37: I'm wondering if we should focus on navigation instead and have screens as a fallback. At least on Android there are a few widely used navigation frameworks (like https://developer.android.com/guide/navigation) which provide an URL-like route.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,74 @@
...
+The goal of this endeavor was to
+
+> always have access to a trace and span ID, add a new internal `PropagationContext` property to the
+> scope, an object holding a `traceId` and `spanId`
+
+On mobile, most users interact purely with the static API, which holds a reference to a global
+Hub and Scope. Therefore, mobile SDKs create a `PropagationContext` with `traceId` and `spanId`
+during initialization, and these usually persist for the entire lifetime of the app. Mobile
+SDKs prefer the `traceID` of transactions bound to the scope over the `PropagationContext`. So
+when performance is disabled, or no transition is bound to the scope, mobile SDKs use the same
```
romtsn, line 30:
```suggestion
when performance is disabled, or no transaction is bound to the scope, mobile SDKs use the same
```

text/0136-mobile-tracing-without-performance-v-2.md
romtsn: we could also assign it to both, but then there has to be a UI change that would show the span spanning (lol) across multiple traces

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,74 @@
...
+foreground after being in the background for longer than 30 seconds, which is the same approach
+mobile SDKs use for determining the end of a session, mobile SDKs renew the `traceId` of the
+PropagationContext. If the app stays in the background for shorter or equal to 30 seconds,
+mobile SDKs must not renew the `traceId` of the PropagationContext when the app moves again to
+the foreground.
+
+### Pros <a name="option-1-pros"></a>
+
+1. Similar to [JavaScript]((https://github.com/getsentry/sentry-javascript/issues/11599)) updating
+it based on routes, so it should be easy to implement for React-Native.
```
buenaflor, line 56: for flutter at the very least we would need the user to use our `SentryNavigatorObserver` and then set a route name, e.g

```dart
navigatorObservers: [
  SentryNavigatorObserver()
],

MaterialPageRoute(
  settings: const RouteSettings(name: 'AutoCloseScreen'),
  builder: (context) => const AutoCloseScreen()),
),
```

Would the default behaviour remain as it is right now if a user didn't use the navigator observer on flutter?

text/0136-mobile-tracing-without-performance-v-2.md
philipphofmann: That makes sense I added an extra rule with https://github.com/getsentry/rfcs/pull/136/commits/764196a99a1516b6b88a26383fd537b4d72d5d8e that should address this problem.

text/0136-mobile-tracing-without-performance-v-2.md
philipphofmann: If you can use routes, use routes, as described below. I updated the title.

text/0136-mobile-tracing-without-performance-v-2.md
philipphofmann: What's the current default behaviour in Flutter, @buenaflor?

krystofwoldrich: There is similar issue with React Native. We don't have the screens/routes information without the performance instrumentation (ReactNavigation or ReactNativeNavigation).

We could update the auto instrumentation work without performance (without creating spans). Or get some signal of change from native.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,70 @@
...
+2. Works for spans first, as all spans get added to one trace per screen.
+
+### Cons <a name="option-1-cons"></a>
+
+1. For single-screen applications such as social networks, the lifetime of a trace could still be
+long, and multiple unrelated events could be mapped to one trace.
```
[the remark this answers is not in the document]
krystofwoldrich: A manual API to renew the `traceId` could help with these edge cases.

text/0136-mobile-tracing-without-performance-v-2.md
krystofwoldrich: The extra rule https://github.com/getsentry/rfcs/commit/764196a99a1516b6b88a26383fd537b4d72d5d8e matched JS V8 and changes in upcoming RN V6 https://github.com/getsentry/sentry-react-native/blob/666cd563b71cf3744fa6dae6fdaa56c79881f3d3/src/js/tracing/reactnativetracing.ts#L568

text/0136-mobile-tracing-without-performance-v-2.md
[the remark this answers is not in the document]
romtsn: fyi, we have [TracingUtils.startNewTrace](https://github.com/getsentry/sentry-java/blob/1e646ca54adc48b77dcf58e74e0c73f3c60cf11c/sentry/src/main/java/io/sentry/util/TracingUtils.java#L17-L25) in Java (should probably be exposed through the static API though)

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,76 @@
...
+2. For applications running for a long time in the background, such as running apps, the lifetime of
+a trace could still be long, and multiple unrelated events could be mapped to one trace.
+3. It doesn’t work well for declarative UI frameworks as Jetpack Compose and SwiftUI for which the
+SDKs can’t reliably automatically detect when apps load a new screen.
+
+# Drawbacks
```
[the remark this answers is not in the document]
philipphofmann: @Lms24, please clarify why this wasn't a problem before. I don't understand how this proposed change here will cause this.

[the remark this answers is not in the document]
Lms24: I might be missing a bit of context around how Tracing without Performance and the PropagationContext is implemented in mobile SDKs. 

If the proposed change is purely intended for TwP scenarios and does not affect the overall trace lifetime in case a root span/transaction _is_ started ("tracing with performance") I think we're good. That is because for TwP, we defer the sampling decision to the downstream service (i.e. send `sentry-trace` headers without a sampled flag).

In JS however, we changed the trace lifetime not just for TwP but in general, leading to scenarios like the one above. To illustrate further, why this is problematic, I'm gonna adjust the example a bit from above

1. Initial Pageload transaction is sampled by rolling the dice
2. PropagationContext stores positive sampling decision
3. application, still on the same page, makes an http request to a downstream service and propagates the `sentry-trace` header _with_ the positive sampling decision, forcing the downstream service to positively sample their transaction.
4. repeat 3 a lot of times (e.g. an application auto-refreshing some state every 5s) and you have a lot of sampled transactions because one initial transaction was sampled positively in the FE.

So even without an active transaction, we'd still propagate a forced sampled flag to downstream services.

Does this make sense?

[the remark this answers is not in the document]
philipphofmann: Thanks for the detailed explanation. Yes, that makes sense, but I guess in the long run, you should have a roughly equal amount of transactions. It shouldn't matter if you roll the dice once for 10 transactions or every time for each transaction. If you roll the dice often enough, an equal amount of transactions should be captured.

text/0136-mobile-tracing-without-performance-v-2.md
```diff
@@ -0,0 +1,76 @@
...
+and use the same information for the name of screen load transactions, which the screen load
+starfish module uses. Whenever the screen name changes automatically or with a yet to be defined
+[manual API](https://www.notion.so/sentry/Specs-Screens-API-084d773272f24f57aeb622c07619264e),
+mobile SDKs must renew the `traceId` of the `PropagationContext`. The screen load transaction
+and subsequent events on the same screen must use the same `traceId`. When the app moves to the
+foreground after being in the background for longer than 30 seconds, which is the same approach
+mobile SDKs use for determining the end of a session, mobile SDKs renew the `traceId` of the
+PropagationContext. If the app stays in the background for shorter or equal to 30 seconds,
+mobile SDKs must not renew the `traceId` of the PropagationContext when the app moves again to
+the foreground. When a span starts, SDKs should use the traceID on the PropagationContext, but
```
stefanosiano, line 51: Wouldn't this break transaction/spans sample rates?
The `sampled` state of the `PropagationContext` would overwrite the calculated sampling decision done through `tracesSampleRate` - or even worse through `tracesSampler`

stefanosiano: or are we just saying to use the traceId of the PropagationContext, retaining the usual sample decision?

text/0136-mobile-tracing-without-performance-v-2.md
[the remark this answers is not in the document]
Lms24: Not necessarily unfortunately. This would only hold up if the sample rates on client and server were the same. If users have lower sample rates on the server, they would send significantly less server-side transactions with the previous implementation.

I tried verifying this with a small script: https://gist.github.com/Lms24/9a631295aef58cf22fb8f5307953335c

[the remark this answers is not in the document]
stefanosiano: > When a span starts, SDKs should use the traceID on the PropagationContext

We should say explicitly to use only the `traceId`, but not the sampling decision of the PropagationContext.
Regardless of client/server side sampling, i think we would break the sampling in general, as it doesn't apply to the PropagationContext. The `sampler` function is particularly problematic imho

text/0136-mobile-tracing-without-performance-v-2.md
stefanosiano: continuing [below](https://github.com/getsentry/rfcs/pull/136/files#r1636312030)

philipphofmann:
The RFC is on hold for now. We found some problems with the suggested approach in JS and need to revisit our approach. After we decided on how to continue we are either going to reopen this RFC or open another one.

PR #137, opened 2024-07-05
rfc(feature): Strict Trage Propagation

mydea:
TODO. [Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/strict-trage-propagation/text/0137-strict-trage-propagation.md)


text/0137-strict-trace-propagation.md
```diff
@@ -0,0 +1,74 @@
...
+
+For now, this will be opt-in, but SDKs are free to switch the default to `true` in a major release. In onboarding for new projects, we should include `strictTracePropagation: true` by default.
+
+In addition to allowing to manually configure the `org`, SDKs may also infer from other places. For example, if a bundler plugin is used we may infer it from there, if possible. 
+
+We may also adjust the DSN structure (which is currently being reworked anyhow) to also include the DSN. In this case, new DSNs may infer the org from there, ensuring that users do not need to manually set the `org` in addition to `dsn`.
+
+## Option B
+
+Another option could be to configure something similar to `tracePropagationTargets` that leads to the SDK only continuing traces for allow-listed URLs. But this requires configuration by the user (and it may also change over time, ...).
```
cleptric, line 66: I would prefer this option. It allows for more granular control and is in line with how other parts of the SDK work.

We could also come up with Option C, which combines A and B. For org wide DSNs, we will already include a service name to the SDK init, so what about we allow a user to choose which of these services should be considered for trace continuation?
The service would then be attached to the baggage as `sentry-service`.

```js
Sentry.init({
    dsn: 'xxx',
    service: 'my-frontend',
});
```

```js
Sentry.init({
    dsn: 'xxx',
    service: 'my-backend',
    tracePropagationOrigins: ['my-frontend'],
});
```

In terms of how the option is evaluated, I propose we stick to  https://develop.sentry.dev/sdk/performance/#sampling, hence this would apply to **3.**. If we get a propagated sampling decision in the `sentry-trace` header (`-1`), we evaluate the service from the baggage against the allow list.

mydea: I think the service is not ideal here, because it may match over orgs. E.g. think that zappier has a `backend` service and sends their requests with `sentry-service=backend`, and in my app I do allow-list `backend` - this would be matched but is exactly what we want to avoid here.

PR #139, opened 2024-08-27
rfc(feature): Versioned docs

chargome:
This RFC proposes the implementation of versioned documentation for [Sentry Docs](https://docs.sentry.io). The goal is to provide users with access to documentation specific to different versions of a Sentry SDK. Currently, users only have access to the documentation of the latest version of a SDK.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/99c564c4244bdae664ef15c5b865556d5a944bdc/text/0139-versioned-docs.md)

### DIFF
--- a/text/0139-versioned-docs.md
+++ b/text/0139-versioned-docs.md
@@ -1,6 +1,6 @@
 - Start Date: 2024-08-27
 - RFC Type: feature
-- RFC PR: TODO
+- RFC PR: https://github.com/getsentry/rfcs/pull/139
 - RFC Status: draft
 
 # Summary
--- a/text/0139-versioned-docs.md
+++ b/text/0139-versioned-docs.md
@@ -15,7 +15,7 @@ This RFC proposes the implementation of versioned documentation for [Sentry Docs
 
 # Options Considered
 
-## 1 (preferred) - Nesting the version by appending the version to the platform path segment
+## 1 Nesting the version by appending the version to the platform path segment
 
 The documentation would include the version number within the URL path (e.g., `docs.sentry.io/platform/javascript/v7/...`, `docs.sentry.io/platform/python/v1/...`). Omitting the version in the path would always point to the latest version. This approach integrates versioning directly into the existing URL structure, allowing users to easily switch between versions while supporting different versions for each SDK. Theoretically this would also allow different folder/file structures under each version. A possible approach would be to have a `versioned_docs` directory under each platform that serves the content for the different versions. This approach potentially introduces content-duplication.
 
--- a/README.md
+++ b/README.md
@@ -65,3 +65,4 @@ This repository contains RFCs and DACIs. Lost?
 - [0129-video-replay-envelope](text/0129-video-replay-envelope.md): Video-based replay envelope format
 - [0131-pass-native-sdk-spans-to-hybrid](text/0131-pass-native-sdk-spans-to-hybrid.md): rfc(feature): Pass Native SDKs Spans to Hybrid
 - [0138-achieving-order-between-pageload-and-srr-spans](text/0138-achieving-order-between-pageload-and-srr-spans.md): Achieving order between Pageload and SRR spans
+- [0139-versioned-docs](text/0139-versioned-docs.md): Versioned documentation for Sentry Docs
--- a/text/XXXX-versioned-docs.md
+++ b/text/XXXX-versioned-docs.md
@@ -1,0 +1,45 @@
+- Start Date: 2024-08-27
+- RFC Type: feature
+- RFC PR: TODO
+- RFC Status: draft
+
+# Summary
+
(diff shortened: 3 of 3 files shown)


rahulchhabria (commented):
this feature only benefits existing customers, should we consider adding the ability to check if someone is logged into sentry first? being able to pull org/project/sdk info would help minimize any confusion around "wait what version am i on, again?" 

i like the idea of versioned docs - but it feels like building this now would introduce a level of complexity that we don't have a good solution for.

HazAT:
I am not sure if I agree with you @rahulchhabria 
Yes, we can combine the logged-in state to pre-select the version. 
But don't think if we don't this right away it's a blocker or too complex for people to understand - so I wouldn't tie those two problems together.

Now, my feedback for the RFC:

I think we should also think about providing versions for not just major version bumps. 
What if we add new features in a minor with a new preferred way we want users to use it? 
Then we should also be able to reflect that somehow. 
If we need to copy an entire folder for every minor change we want to render, it might quickly become overwhelming.

Can we figure out a solution to version individual files/pages, not entire folders?

text/0139-versioned-docs.md
```diff
@@ -0,0 +1,45 @@
...
+- Content duplication
+- User confusion (user might consume outdated information without being aware of it)
+- (Potentially heavily) increased build time due to the number of added pages for each platform.
+- Link management
+- SEO - indexed versioned pages could potentially lead users to outdated information
+
+# Unresolved questions
+
+- How could the build process be optimized to not suffer from introducing versions? (e.g. SSR, ISR, ...)
+- How can we make version switching intuitive from a UX perspective?
```
antonpirker, line 44: In Python land if you end up on an outdated version of the docs, a message is displayed at the top that shows you that you are on old docs. Examples (I like the first one more, because it stands out more):
- https://docs.djangoproject.com/en/3.0/intro/tutorial03/
- https://docs.aiohttp.org/en/v3.8.0/client_advanced.html#custom-request-headers

Those pages have also a version selector at the bottom right corner, where the user can select the version of the docs.

antonpirker:
With duplication of code this will make maintaining the prose text in the documentation harder, and the same page in different versions *will* diverge. This is something we need to keep in mind. Most of the times we want the pages to be different because something changed in the new version. But if we want to update the wording, we need to make sure we do it in all versions.

antonpirker:
Another thing (maybe not fully related to this RFC). 

I would love to render the "[options](https://docs.sentry.io/platforms/javascript/configuration/options/)" sections from source code. We could add this documentation to the source and then render the mark down file from the source. So we would not need to update the code and the docs pages separately, but only update the code and if there are new options or new documentation for an existing option, the documentation is updated automatically.

mwarkentin:
I saw this a while back, wonder if we could use it to render parts of our docs from source: https://til.simonwillison.net/python/cog-to-update-help-in-readme

chargome:
> Another thing (maybe not fully related to this RFC).
> 
> I would love to render the "[options](https://docs.sentry.io/platforms/javascript/configuration/options/)" sections from source code. We could add this documentation to the source and then render the mark down file from the source. So we would not need to update the code and the docs pages separately, but only update the code and if there are new options or new documentation for an existing option, the documentation is updated automatically.

@antonpirker So you would want to render the options version specific from source I assume?

chargome:
> Can we figure out a solution to version individual files/pages, not entire folders?

@HazAT with option 3 we would do basically that. We could for example add the version to filename and use that for displaying different versions of the same page. Or do you think it would make sense to do the versioning within a file (maybe something with a `<Versioned version="1.1.1">...` component?

antonpirker:
> > Another thing (maybe not fully related to this RFC).
> > I would love to render the "[options](https://docs.sentry.io/platforms/javascript/configuration/options/)" sections from source code. We could add this documentation to the source and then render the mark down file from the source. So we would not need to update the code and the docs pages separately, but only update the code and if there are new options or new documentation for an existing option, the documentation is updated automatically.
> 
> @antonpirker So you would want to render the options version specific from source I assume?

Yes exactly. But I think this is not really related with this RFC. However we decide on organizing the versioned docs, we then can add some code or github action or whatever, to render the options page for the right platform and the right version accordingly

text/0139-versioned-docs.md
krystofwoldrich: I would also expect message when not on the latest or when on beta/next version.

And selector somewhere on the page, personally I would expect it to be in the header.

For example like on https://yarnpkg.com/

krystofwoldrich:
Would this let us write docs for next/beta versions, or is that out of scope?

chargome:
@krystofwoldrich depending on the implementation we could do that yes

chargome:
Docs for versioning available [here](https://docs.sentry.io/contributing/platforms/versioning/)

PR #140, opened 2024-09-04
rfc(feature): Record Feature Flags on Error

cmanallen:


text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+We will collect flag evaluations and hold them in-memory.  On error event the flags will be placed into the `contexts` object on the `event` and sent to Sentry.
+
+## Transport
+
+The flags will be represented by this data structure during transport. The `flag` key is the name of the flag and the `result` key is the evaluation result returned by the customer's application.
+
+```json
+{
+    "contexts": {
+        "flags": [
```
bruno-garcia, line 25: I wonder if we risk conflicting with customers that today already use this key. And if a prefix like `sntr_flags` could be beneficial?


```suggestion
        "sntr_flags": [
```

We don't do this with other contexts though so not sure if it's worth it

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+}
+```
+
+## Public Interface
+
+The SDK will expose one new public method `set_flag`.  Similar to `set_tag` or `set_user`, the `set_flag` method sets a key, value pair (representing the flag's name and its evaluation result) into an internal SDK structure. On error, that structure is serialized and appended to the event body as described in the previous section.
+
+## Bounding Memory Usage and Transport Size
+
+We will cap the number of flag evaluations to some fixed capacity (e.g. 100). Duplicate evaluations will update the existing entry rather than insert a new one. New unique evaluations will be appended to the data structure with the least recently accessed evaluation being dropped.
```
bruno-garcia, line 39:
```suggestion
We will cap the number of flag evaluations to some fixed capacity (e.g. 100). Duplicate evaluations will update the existing entry rather than insert a new one. New unique evaluations will be appended to the data structure with the least recently accessed evaluation being dropped. This is the same behavior we have for Breadcrumbs today.
```

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Integrations
+
+Integrations for feature flag SDKs will need to be written. There are many competing offerings and we'll want to provide integrations for. The main providers we want to initially target are: launchdarkly, unleash, split, and OpenFeature. There public SDK interface is available for those wishing to integrate with custom or non-supported vendors.
```
bruno-garcia, line 43:
```suggestion
Integrations for feature flag SDKs will need to be written. There are many competing offerings and we'll want to provide integrations for. The main providers we want to initially target are: launchdarkly, unleash, split, and OpenFeature. The public SDK interface is available for those wishing to integrate with additional vendors.
```

bruno-garcia (commented):
Nice!

text/0140-record-feature-flags-on-error.md
cmanallen: I'm also wondering if a top-level field will be suggested and what, if any, work that entails. I think its fine to name-space as you did.  I'll leave it as is for now and wait for reviews before changing.

text/0140-record-feature-flags-on-error.md
cmanallen: I'm not familiar with breadcrumbs; is it the same or very similar? (just trying to be extra precise with the language)

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+            {"flag": "abc", "result": true},
+            {"flag": "def", "result": false}
+        ]
+    }
+}
+```
+
+## Public Interface
+
+The SDK will expose one new public method `set_flag`.  Similar to `set_tag` or `set_user`, the `set_flag` method sets a key, value pair (representing the flag's name and its evaluation result) into an internal SDK structure. On error, that structure is serialized and appended to the event body as described in the previous section.
```
cleptric, line 35: Not a big fan of another method. With spans, we will unify to one single `setAttributes` method. For flags, using `setContext('flags', ...)` should suffice.

cmanallen: How would someone incrementally add flags as they're evaluated?  How would we manage the capacity on the flags key?

bruno-garcia: I'm not a fan of magic strings tbqh. If we're making flags a first class citizen, it's beneficial to have a direct API for it. Also for discoverability.

Is the concern here bundle size on JS? Or just "too many functions/complexity"?

cleptric: My main concern here is to add new methods before the new feature was fully evaluated.
If we scope the initial implementation to one or two SDKs and trail it for a while behind some feature experimental feature flag, we can proceed. I'm not too worried about the capacity at first, though merging values is a good argument against using `setContext()`.

text/0140-record-feature-flags-on-error.md
bruno-garcia: It's the same but we can keep it out since the description by itself is clear enough.

text/0140-record-feature-flags-on-error.md
bruno-garcia: > My main concern here is to add new methods before the new feature was fully evaluated.

That's fair. No need to push for publish APIs while we're in alpha/beta phase. 

The RFC is proposing a solution long term though, in a world where we're not canceling this project, the final API proposal is what we agree on in this document

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+    "contexts": {
+        "flags": [
+            {"flag": "abc", "result": true},
+            {"flag": "def", "result": false}
+        ]
+    }
+}
+```
+
+## Public Interface
```
cmanallen, line 33: Might need a `reset` or `checkpoint` method for back-end applications.  For example an HTTP server, some flags are called at startup and others are called per request.  It would be wrong to see feature flags from a previous request in the current request's scope.

If we're storing on the `scope` maybe this is handled already?

cleptric: We are moving the SDKs to a new scope model, JavaScript is already there and Python is next. Meaning, we don't have hubs anymore, but instead a global scope and an isolation scope. The isolation scope is "isolated" per request.

cmanallen: Perfect.

bruno-garcia: fwiw that would work with the Hub model too. Since the original flags resolved during startup would be available to the cloned hubs (and its scope) used to serve requests. 

For example, I had to clear breadcrumbs on aspnet because the bootstrapping logs (sometimes days or months old) would show up on all web requests events and got confusing to some folks

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,45 @@
...
+
+The SDK will expose one new public method `set_flag/2`. The method accepts the arguments `flag` (of type string) and `result` (of type boolean).  Similar to `set_tag/2` or `set_user/1`, the `set_flag/2` method stores a flag, result pair on the isolation scope. On error, the isolation scope's flags are serialized and appended to the event body as described in the previous section.
+
+Stateless, multi-tenented applications, such as web servers, must isolate flag evaluations per request.
```
michellewzhang, line 37:
```suggestion
Stateless, multi-tenanted applications, such as web servers, must isolate flag evaluations per request.
```

text/0140-record-feature-flags-on-error.md
cmanallen: @cleptric Should flags exist in the context or should they be given their own top level field?

bruno-garcia, line 25: Just realized that we need an object here, as Sentry won't accept an array.

Could be:

```
"flags": {
 "values": [ ]
```

cmanallen: Thank you!

cleptric: Context

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,47 @@
+- Start Date: 2024-09-04
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull40
```
michellewzhang, line 3: @cmanallen i think you accidentally modified the PR link here
```suggestion
- RFC PR: https://github.com/getsentry/rfcs/pull/140
```

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,47 @@
...
+        { "flag": "def", "result": false }
+      ]
+    }
+  }
+}
+```
+
+## Public Interface
+
+The SDK will expose one new public method `set_flag`. The method accepts the arguments `flag` (of type string) and `result` (of type boolean). Similar to `set_tag` or `set_user`, the `set_flag` method stores a flag, result pair on the isolation scope. On error, the isolation scope's flags are serialized and appended to the event body as described in the previous section.
```
stefanosiano, line 37: What if the user uses a number or a string as feature flag?
[LaunchDarkly](https://docs.launchdarkly.com/sdk/concepts/flag-types) (and probably others) supports that

text/0140-record-feature-flags-on-error.md
```diff
@@ -0,0 +1,47 @@
...
+
+Stateless, multi-tenanted applications, such as web servers, must isolate flag evaluations per request.
+
+## Bounding Memory Usage and Transport Size
+
+The number of flag evaluations must be capped to some fixed capacity (e.g. 100). When the capacity is reached the least recently evaluated flag should be dropped from the set. Duplicate flag evaluations may update their existing entry rather than insert a new one. In most contexts, it will make sense to update the existing entry. An LRU cache is recommended for these cases. If repeat flag evaluations are desired then a ring buffer may be used.
```
Lms24, line 43: I'm wondering if this is strictly necessary? Do we have backend requirements to keep the number of flags capped to e.g. 100? SDK-side, we have normalization and trimming logic to keep event sizes under control, as well as a [maxbreadth attribute](https://github.com/getsentry/sentry-javascript/blob/1e9a1a3c80b10d2aca19b640666af9134d3bf40a/packages/types/src/options.ts#L136-L156) to control how many keys an object can contain. 

Also, duplicated flags should be fairly easy to avoid (at least in JS) via `setFlags` but let's not forget that users could always overwrite the entire content via `setContext('flags', ...)`.

Lms24 (commented):
I'm generally fine with adding the top-level `setFlag` method, given this is tree-shakeable in JS.

My general sentiment from the SDK perspective is that we keep things rather as simple as possible, meaning that for now we don't care too much about a specific max number of flags but treat it like any other context field in the SDK.

I have two asks though if/when we move forward with this:
1. Contexts are a bit arbitrary in the SDK because besides some predefined keys, users can define whatever context they want. We should definitely document that `flags` would be a reserved key in [develop docs](https://develop.sentry.dev/sdk/event-payloads/contexts/) and - wherever applicable - register this key in the type definition
2.  Any integration we potentially add in the future to automatically read flags from other tools/sdks should be opt-in for the JS browser SDK, due to bundle size.

text/0140-record-feature-flags-on-error.md
andreiborza: What happens if users have already defined this key?

sl0thentr0py (approved):

sentrivana (approved):

text/0140-record-feature-flags-on-error.md
cmanallen: Malformed code which inserts unique flags in a loop would present as a memory leak.  Whether or not this is reason for us to put in a defense mechanism I suppose is a matter of Sentry's opinion.

PR #141, opened 2024-11-08
[WIP] rfc(feature): Support Linking Traces

Lms24:
This is still a heavy WIP

### DIFF
--- a/text/0141-linking-traces.md
+++ b/text/0141-linking-traces.md
@@ -1,0 +1,54 @@
+- Start Date: 2024-11-08
+- RFC Type: feature 
+- RFC PR: <[link](https://github.com/getsentry/rfcs/pull/141)>
+- RFC Status: draft
+
+# Summary
+
+This RFC aims to introduce the ability to link or connect traces in Sentry. It involves SDKs, Ingest, Search and Storage, Backend and at a later stage product teams.
+The goal of this initiative is to be able to link multiple traces together so that the Sentry product can show what happened before a trace while preserving the integrity of individual traces.
+This RFC proposes to use "Span Links" as a vehicle specifying relationships between spans and traces. It also discusses other options considered, as well as concrete use cases for this feature.
+
+# Motivation
+
+At the time of writing this RFC, Sentry (the product as well as SDKs) cannot connect multiple traces to deliver a "bigger picture" of what happened before (or after) a specific trace. For example, for frontend applications, we would like to display a user journey (session). Today, we are always limited to the duration of one trace (id), which is handled and kept alive for different times across different SDKs.
+
+TODO: Second problem: child spans outliving their parent spans within a trace
+
+# Background
+
+- TODO: OpenTelemetry Span Links
+- TODO: Browser SDK trace model
+- TODO: Mobile trace model
+
+# Supporting Data
+
+- TODO: Not much data available but we can list specific problems of long-living traces and lack of inter-trace context
+
+# Options Considered
+
+## [Preferred] Span Links
+
+TODO: 
+- SDK implementation (Otel interface, special link attributes)
+- Ingest changes
+- SnS changes (current event storage/EAP?)
+- How to handle sampling
+- How to handle non-transaction/span events: Last traceId
+
+## [Alternative] Previous Trace Id
+TODO
+
+# Drawbacks
+
+TODO:
+- Requires a lot of changes across SDKs, ingest, SnS
+- Sampling implications
+
+# Unresolved questions
+
+- Can we do this with our current event storage? Are we blocked on EAP?
+
+# Tmp: Additional information
+
+ While in backend SDKs most use cases have a clear definition for starting and ending a trace (the request lifecycle), frontend SDKs don't have clear borders when traces start and end. Therefore, over the years and iterations, we created different trace models for frontend SDKs where sometimes traces intentionally span more than one root span. Other times, traces end after a certain idle time or other debouncing mechanism.


text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+At the time of writing this RFC, Sentry (the product as well as SDKs) cannot connect multiple traces to deliver a "bigger picture" of what happened before (or after) a specific trace. 
+A trace is therefore the highest level of connecting signals (errors, traces, profiles, etc.).
+The notable exception is Session Replay, where we are indeed able to provide this bigger picture.
+
+This RFC proposes a method of achieving linkage between (and within) traces in the entire Sentry product, to satisfy a number of currently sub-optimally solved use cases.
+
+## Frontend Traces and User Journeys
+
+The most important application of linking traces are frontend applications are frontend applications. We would like to display a user journey (session) which should make debugging of issues much easier as developers have more context on what happened before a specific issue.
```
mydea, line 22:
```suggestion
The most important application of linking traces are frontend applications. We would like to display a user journey (session) which should make debugging of issues much easier as developers have more context on what happened before a specific issue. 
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+This RFC proposes a method of achieving linkage between (and within) traces in the entire Sentry product, to satisfy a number of currently sub-optimally solved use cases.
+
+## Frontend Traces and User Journeys
+
+The most important application of linking traces are frontend applications are frontend applications. We would like to display a user journey (session) which should make debugging of issues much easier as developers have more context on what happened before a specific issue. 
+Today, we are limited to the duration of one trace (id), which is handled differently and kept alive for different times across our various SDKs. A concrete example are our [JavaScript Browser SDKs](https://develop.sentry.dev/sdk/platform-specifics/javascript-sdks/browser-tracing/#tracing-model) which by default keep a trace alive as long as users
+are on the same page our (URL) route. This was a compromise in which we accepted that a trace would consist of multiple trace root spans (transactions), which is generally discouraged by tracing models like OpenTelemetry.
+Another example of a suboptimal trace model is the one used in most mobile SDKs. In these SDKs we
```
chargome, line 25: something is missing here

text/0141-linking-traces.md
chargome, line 22:
```suggestion
The most important application of linking traces are frontend applications. We would like to display a user journey (session) which should make debugging of issues much easier as developers have more context on what happened before a specific issue. 
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+  links?: Link[];
+}
+```
+
+SDKs should follow the OpenTelemetry spec for the [`Link` interface](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/link.ts) as defined by the platform. Non-Otel SDKs should orient themselves on Otel, resulting in the following interface:
+
+```typescript
+interface Link {
+  // contains the SpanContext of the span to link to
+  context: SpanContext;
```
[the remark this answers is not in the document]
Lms24: It's definitely something to investigate (will do!) but I think for our use cases, span and trace id should be enough.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+# Options Considered
+
+## [Preferred] Span Links
+
+We propose to add links on a span level, as defined and specified by [OpenTelemetry](https://opentelemetry.io/docs/concepts/signals/traces/#span-links). In addition to linking to span ids, a span link also holds meta information about the link, collected via attributes. We'll make use of these attributes by annotating our created span links with a `sentry.link.type` attribute which we can later use in the product to query for linked traces.
+
+### SDK APIs
+
+To support adding span links, SDKs need to expose at least an `addLink` method on their respective `Span` interface. For completeness with OpenTelemetry, ideally they also expose `addLinks`:
```
chargome, line 172: just to mention: for the js sdk this is already exposed on the span interface but basically does nothing

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+
+## Secondary goals
+
+- We can shorten the lifetime of frontend traces because we're now able to link to the previous trace instead. This is something we'll do but it's a follow-up and can only be addressed once the primary goals are achieved. 
+
+## Out of scope/Non-Goals
+
+- Querying across linked traces: Given the "linked list" nature of one root span linking back to its previous root span, we accept that obtaining a complete list is an expensive query operation. Therefore, we consider use cases that require a complete list of all linked traces out of scope. We can address this as a follow-up if technically possible but we're aware that this is likely a sub-optimal data structure for such queries. 
+- Linking transactions started during the rendering of statically generated HTML pages and pageload transactions. A span link is ideal here because we don't want the hard link of a trace id. However, a casual link would still increase context and completeness. Wwe might do it but not at this point as it's a follow-up item.
```
andreiborza, line 57:
```suggestion
- Linking transactions started during the rendering of statically generated HTML pages and pageload transactions. A span link is ideal here because we don't want the hard link of a trace id. However, a casual link would still increase context and completeness. We might do it but not at this point as it's a follow-up item.
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+}
+```
+
+
+## Current Session Models in Sentry SDKs
+
+Sentry SDKs do actually send session data, in fact even two types of sessions. However, neither of the two sessions are used and associated with a trace. 
+
+- SDK error sessions: SDKs currently send a session that counts and describes the error status of such a session. Depending on the error, the session is marked as crashed, abnormal or healthy, which is information that powers our Release (Health) product. These sessions are fundamentally flawed though because (at least in Browser JS) they only last as long as the currently loaded page. Every soft or hard navigation causes the session to be ended and a new one to be started. 
+- Replay Session: Frontend SDKs persist a replayId in the browser's `sessionStorage` which (even though the name does not suggest it) more accurately models a session that the SDKs' error sessions. We cannot have a hard dependency on this session as Replay is an extra product from Sentry, meaning this replay id is not always reliably set. However, the model of persisting the replay id can use as a blueprint on how we'd persist the last traceId.
```
andreiborza, line 130:
```suggestion
- Replay Session: Frontend SDKs persist a replayId in the browser's `sessionStorage` which (even though the name does not suggest it) more accurately models a session than the SDKs' error sessions. We cannot have a hard dependency on this session as Replay is an extra product by Sentry, meaning this replayId is not always reliably set. However, the model of persisting the replay id can be used as a blueprint for how we could persist the last traceId. 
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+- Replay Session: Frontend SDKs persist a replayId in the browser's `sessionStorage` which (even though the name does not suggest it) more accurately models a session that the SDKs' error sessions. We cannot have a hard dependency on this session as Replay is an extra product from Sentry, meaning this replay id is not always reliably set. However, the model of persisting the replay id can use as a blueprint on how we'd persist the last traceId. 
+
+Upon decision from Leadership as well as from it being noted in Sentry's Goal Hierarchy, will not associate spans or traces via any of these sessions. Instead, the sessions will stay as-is and we will link traces in a "linked list"-style approach as described in this RFC.
```
andreiborza, line 132:
```suggestion
Upon decision from Leadership as well as from it being noted in Sentry's Goal Hierarchy, we will not associate spans or traces via any of these sessions. Instead, the sessions will stay as-is and we will link traces in a "linked list"-style approach as described in this RFC. 
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+# Supporting Data
+
+Sentry users would like to get as much insight as possible when inspecting (errors or performance) issues. Right now, with the notable exception of Session Replay, we can only provide answers as to what happened before an error occured up to the point when a trace was started. If we can link from the current to the previous trace, we can significantly widen the insight into the entire user journey, which is often crucial to understand why and how a specific issue occured. 
+
+Given that 32% of ARR can be attributed alone to events sent from Sentry frontend JavaScript SDKs and a further 15% from mobile SDK, a significant portion of sent events would directly benefit from better linkage between events. Considering that frontend applications almost always have a backend counter part, another 44% of ARR can be attributed to backend SDKs that would implicitly also benefit from a better tracing models. Wins all around!
```
andreiborza, line 138: q: Are these ARR numbers public knowledge? 😅

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+
+While linked traces themselves is not identical to user sessions, we're aware that users have been asking fo some kind of connectedness in various forms on GitHub as well as Sales/Solution Engineers. Typically, in a frontend application context, users would expect Sentry to show them a user _session_. In such cases we typically explain that Sentry [does track sessions](#current-session-models-in-sentry-sdks), albeit in a flawed way. Right now, we cannot answer questions like, what happened in a previous trace that might have had an impact on the trace with an error. With linked traces, we can.
```
andreiborza, line 140:
```suggestion
While linked traces themselves are not identical to user sessions, we're aware that users have been asking for some kind of connectedness in various forms on GitHub as well as via Sales/Solution Engineers. Typically, in a frontend application context, users would expect Sentry to show them a user _session_. In such cases we explain that Sentry [does track sessions](#current-session-models-in-sentry-sdks), albeit in a flawed way. Right now, we cannot answer questions like, what happened in a previous trace that might have had an impact on the trace with an error. With linked traces, we can.
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+- Honeycomb.io supports a [session id](https://docs.honeycomb.io/send-data/javascript-browser/honeycomb-distribution/) which is not persisted in storage.
+- [Logrocket sessions](https://docs.logrocket.com/docs/what-defines-a-session) can span multiple tabs and have a lifetime of 30 minutes, with some further heuristics for closing tabs
+
+Again, linked traces are not identical to sessions but they can also help answer questions that sessions could answer.
+
+## Implications of Long-lived traces
+
+As mentioned [above](#frontend-traces-and-user-journeys), our current long-lived traces in Browser JS have some negative implications on the Sentry UX:
+
+- Multiple root spans (transactions): Our product can (and should) handle multiple root spans. However, there are some rough edges which are hard to solve from a UI/UX perspective. Our Trace explorer gives traces a name, for instance. If multiple root spans exist, we can only use a heuristic which span name to use for the "trace name". Furthermore, multiple root spans contradict the conventional concept of a trace, where a trace should only consist of one root span.
```
andreiborza, line 154:
```suggestion
- Multiple root spans (transactions): Our product can (and should) handle multiple root spans. However, there are some rough edges which are hard to solve from a UI/UX perspective. Our Trace explorer gives traces a name, for instance. If multiple root spans exist, we can only use a heuristic for which span name to use for the "trace name". Furthermore, multiple root spans contradict the conventional concept of a trace, where a trace should only consist of one root span.
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+  traceId: string,
+  spanId: string,
+  traceFlags: string,
+}
+
+type Attributes = Record<string, AttributeValues>
+type AttributeValues = string | boolean | number | Array<string> | Array<boolean> | Array<number>
+```
+
+Note: In some platforms, the Otel `Link` interface exposes another optional property: `droppedAttributesCount`. POtel SDKs should support passing in this property as defined by the API but can further ignore it when serializing the span link to Sentry envelopes.  In [JS for example](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/link.ts), the `droppedAttributeCount` can be passed, while [Python](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L120) does not permit it.
```
andreiborza, line 216:
```suggestion
Note: On some platforms, the Otel `Link` interface exposes another optional property: `droppedAttributesCount`. POtel SDKs should support passing in this property as defined by the API but can further ignore it when serializing the span link to Sentry envelopes.  In [JS for example](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/link.ts), the `droppedAttributeCount` can be passed, while [Python](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-api/src/opentelemetry/trace/span.py#L120) does not permit it. 
```

andreiborza:
Great stuff! One thing I was wondering about is how links are affected by span filtering. What if links are actively dropped by users? Is that a concern?

text/0141-linking-traces.md
```diff
@@ -0,0 +1,361 @@
...
+- Linking transactions started during the rendering of statically generated HTML pages and pageload transactions. A span link is ideal here because we don't want the hard link of a trace id. However, a casual link would still increase context and completeness. Wwe might do it but not at this point as it's a follow-up item.
+- Unsampled traces are not linked. Errors across unsampled traces are not linked. For example, if the first trace was sampled negatively, the next trace will not link to it as this would result in a link to a non-existing trace.
+
+# Background
+
+This section provides background information of important concepts for this RFC
+
+## OpenTelemetry Span Links
+
+OpenTelemetry introduced the concept of span links to establish casual relationships between individual spans. Such links can be established by
```
mikejihbe, line 66:
```suggestion
OpenTelemetry introduced the concept of span links to establish causal relationships between individual spans. Such links can be established by
```
🤣

Lms24:
> Great stuff! One thing I was wondering about is how links are affected by span filtering. What if links are actively dropped by users? Is that a concern?

I would argue that this is up to users. If users decide to drop a span link, we just won't be able to link the respective spans together. Meaning, unless there's a strong case for consistent sampling across traces (which is still TBD) I think we can permit mutation of links just like any other span property.

text/0141-linking-traces.md
Lms24: lol good catch 😅 thanks!

text/0141-linking-traces.md
Lms24: thx, went with https://github.com/getsentry/rfcs/pull/141#discussion_r1844089805

text/0141-linking-traces.md
Lms24: thx, went with https://github.com/getsentry/rfcs/pull/141#discussion_r1844089805

text/0141-linking-traces.md
Lms24: went with https://github.com/getsentry/rfcs/pull/141#discussion_r1844089805

text/0141-linking-traces.md
Lms24: Good point, I'll add a note in the RFC. Also, I think this serves as a pro for going with span links: we already expose the API, and it not being functional is confusing for users familiar with pure Otel

text/0141-linking-traces.md
Lms24: I reworded this paragraph to be more generalizing. Good point, thanks!

text/0141-linking-traces.md
Lms24: https://github.com/getsentry/rfcs/pull/141/commits/4cba46c50743a701d6862338b269b73e1fd3b73a

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+
+Upon decision from Leadership as well as from it being noted in Sentry's Goal Hierarchy, we will not associate spans or traces via any of these sessions. Instead, the sessions will stay as-is and we will link traces in a "linked list"-style approach as described in this RFC. 
+
+# Supporting Data
+
+Sentry users would like to get as much insight as possible when inspecting (errors or performance) issues. Right now, with the notable exception of Session Replay, we can only provide answers as to what happened before an error occurred up to the point when a trace was started. If we can link from the current to the previous trace, we can significantly widen the insight into the entire user journey, which is often crucial to understand why and how a specific issue occurred. 
+
+Given that almost half of events sent to Sentry can be attributed alone to events sent from Sentry frontend JavaScript or mobile SDKs, a significant portion of sent events would directly benefit from better linkage between events. Considering that frontend applications almost always have a backend counter events sent from backend SDKs would implicitly also benefit from a better tracing model. Wins all around!
```
s1gr1d, line 141:
```suggestion
Given that almost half of the events sent to Sentry can be attributed solely to events sent from Sentry's frontend JavaScript or mobile SDKs, a significant portion of sent events would directly benefit from better linkage between events. Considering that frontend applications almost always have a backend, counter events sent from backend SDKs would implicitly also benefit from a better tracing model. Wins all around!
```
Some structure improvements for better readability :)
But I am not sure about the phrase "counter events" resp. "backend counter events"

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+}
+```
+
+
+## Current Session Models in Sentry SDKs
+
+Sentry SDKs do actually send session data, in fact even two types of sessions. However, neither of the two sessions are used and associated with a trace. 
+
+- SDK error sessions: SDKs currently send a session that counts and describes the error status of such a session. Depending on the error, the session is marked as crashed, abnormal or healthy, which is information that powers our Release (Health) product. These sessions are fundamentally flawed though because (at least in Browser JS) they only last as long as the currently loaded page. Every soft or hard navigation causes the session to be ended and a new one to be started. 
+- Replay Session: Frontend SDKs persist a replayId in the browser's `sessionStorage` which (even though the name does not suggest it) more accurately models a session than the SDKs' error sessions. We cannot have a hard dependency on this session as Replay is an extra product by Sentry, meaning this replayId is not always reliably set. However, the model of persisting the replay id can be used as a blueprint for how we could persist the last traceId.
```
philipphofmann, line 133: I think we should align on what a session is at Sentry. Having different types of sessions for Session Replay and the rest of the product will not only confuse us but also our users. If we plan on using the sessionStorage as a blueprint, we should instead consider extracting the logic from Session Replay.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+- Replay Session: Frontend SDKs persist a replayId in the browser's `sessionStorage` which (even though the name does not suggest it) more accurately models a session than the SDKs' error sessions. We cannot have a hard dependency on this session as Replay is an extra product by Sentry, meaning this replayId is not always reliably set. However, the model of persisting the replay id can be used as a blueprint for how we could persist the last traceId. 
+
+Upon decision from Leadership as well as from it being noted in Sentry's Goal Hierarchy, we will not associate spans or traces via any of these sessions. Instead, the sessions will stay as-is and we will link traces in a "linked list"-style approach as described in this RFC.
```
philipphofmann, line 135: Yes, for now, but in x years, we might actually do that.

```suggestion
Upon decision from Leadership as well as from it being noted in Sentry's Goal Hierarchy, we will not associate spans or traces via any of these sessions. Instead, the sessions will stay as-is and we will link traces in a "linked list"-style approach as described in this RFC. Still, at some point in the future, we might do so, and the solution in this RFC shouldn't block linking traces and spans to sessions.
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+// 1st trace starts
+const pageloadSpan = startInactiveSpan(...)
+
+// 2nd trace starts
+const navigation1Span = startInactiveSpan({
+  name: '/users', 
+  links: [{
+    context: pageloadSpan.spanContext(), 
+    attributes: {
+      'sentry.link.type': 'previous_trace'
```
philipphofmann, line 238: For strictly typed languages, we must come up with a different API, but I guess this is out of the scope of this RFC. It's mostly a reminder that we align on an API before everybody starts implementing their own.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+      "span_id": string,
+      "trace_id": string, 
+      attributes: Record<string, AttributeValue>,
+    }>
+    // ...
+  }>
+  // ...
+}
+```
+
```
philipphofmann, line 315: #### Pros

1. Allows adding multiple span links to one span
2. Allows linking to other data types, such as errors, session replays, user feedback, sessions, and future data types.
3. Extendable: We can add new types of links without having to change the API.
4. Compatible with OTEL
5. Is compatible with traces, transactions, and spans. We can link everything that has a trace, such as errors.
6. Should be compatible with a potential future session ID. If we at some point add a session ID or use sessions as a session ID to link all data to one session, we should only have to add a `'sentry.link.type':  'session'`.

philipphofmann (commented):
Thanks for all the deep context and background. I added a few comments.

text/0141-linking-traces.md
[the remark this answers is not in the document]
mjq: Based on [the protobuf schema for links](https://github.com/open-telemetry/opentelemetry-proto/blob/90986046c5f0523ca97a3f5933389c5194b5d6e7/opentelemetry/proto/trace/v1/trace.proto#L249-L287) I'm surprised the other properties weren't serialized, but nonetheless it's valid for a SpanContext to only have `spanId` and `traceId` and I can't really imagine any problems arising from that.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+- Can we do this with our current event storage? Are we blocked on EAP?
+  Yes. We will certainly need to adapt our current event storage but it is possible to do this without being blocked on EAP. We'll separately also need to adjust the planned span schema for EAP but this is not a blocker.
+
+
+- How would we handle errors outside of an active span?
+  Errors happening while no span is active, should (as currently is the case in JS Browser SDKs) belong to the last created trace. Therefore, they will store all required fields in their `trace` context.
+
+# Unresolved Questions
+
+- Should we (try to) keep sampling decisions consistent for subsequent (frontend) traces? IOW: If we sample the initial trace, should all subsequent traces be force-sampled as well?
```
mjq, line 362: This seems like the key question for this proposal to me - otherwise, the odds of having links are very low, especially an uninterrupted chain of them, given client and server sampling. I'm not sure if the feature is useful if the traces are still sampled independently.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+  span.addLink({
+    context: navigation1Span.spanContext(), 
+    attributes: {
+      'sentry.link.type': 'previous_trace'
+    }
+  })
+})
+```
+
+In this example, by adding span links, we can link from the last navigation trace all the way back to the initial pageload trace. By passing the `'sentry.link.type': 'previous_trace'` attribute, we can identify the link as a previous trace link in Sentry and display the spans accordingly.
```
mjq, line 254: This example talks about navigation traces, but what about other kinds of traces? Especially once traces are shortened? You can imagine having traces for interactions, web requests, etc - is the idea still to link each of these to the previous one? What about when the previous trace is still active when the new one starts? e.g.

```
[-----------trace 1, span A: interaction-----------]
               [ trace 1, span B: http request ]
   
     [trace 2, span C: http request]
     
                                                         [trace 3, span D: http request]
```
Does trace 2 link to trace 1 as the previous trace, or whatever the previous finished trace was? (Therefore forming a tree rather than a list). Does trace 3 then link to trace 2, even though trace 1 is the most recent span to finish? etc.

text/0141-linking-traces.md
mjq: > Allows linking to other data types, such as errors, session replays, user feedback, sessions, and future data types.

Sorry, I might be missing something but how does this proposal allow you to do that?
Typically everything goes the opposite direction (everything else has traceId and spanId properties) - what would be the use case for linking in the opposite direction?

> Should be compatible with a potential future session ID. If we at some point add a session ID or use sessions as a session ID to link all data to one session, we should only have to add a 'sentry.link.type':  'session'.

FWIW the de facto standard way to handle sessions would be with a `session.id` attribute - not sure what the advantages of a link would be over that? - but that's all down the road if ever.

text/0141-linking-traces.md
mjq: If the crappy ASCII isn't clear, to rephrase in text:

1. Trace 1 starts
2. Trace 2 starts
3. Trace 2 ends
4. Trace 1 ends
5. Trace 3 starts
6. Trace 3 ends

How are these three traces joined by links?

text/0141-linking-traces.md
```diff
@@ -0,0 +1,367 @@
...
+- (Async) data flow operations (e.g. RxJS pipes) where we would be able to link individual operations.
+- Websocket spans where we could link individual web socket messages
+- More generally, whenever we want to "casually" establish a relationship between spans that cannot or should not be achieved via a hard relationship (traceId or parent spanId)
+
+# Concrete Goals of achieving linked traces
+
+Once this RFC is accepted, we have settled on all important details needed to implement connected traces.
+This means, we have all the necessary information to support the following concrete goals:
+
+TODO: List concrete product/UI goals
```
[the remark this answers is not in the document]
Lms24: Yup agreed! I hope to have something more concrete here by tomorrow

text/0141-linking-traces.md
Lms24: I agree 100% and fwiw, we've started talking about making the replay session the main (and only) session in the SDK.

text/0141-linking-traces.md
Lms24: For sure! I think the specific API details (how to model the options, etc) can stay up to specific SDKs and platforms. What's important is that we settle on the top level stuff and agree on the output.

text/0141-linking-traces.md
philipphofmann: > what would be the use case for linking in the opposite direction?

You're right. Most of the time, all the data types link to a trace. Linking a span or multiple spans to an error could be one use case. I guess we should change the pro to:
2. Allows linking to other data types, such as errors and future data types.  

> FWIW the de facto standard way to handle sessions would be with a session.id attribute - not sure what the advantages of a link would be over that? - but that's all down the road if ever.

I'm not saying we have to do it as I described it; we could only use the same approach for sessions if we needed to. Yes, that's down the road.

text/0141-linking-traces.md
philipphofmann: What about changing the relationship of `previous_trace` from 1:1 to 1:n and renaming it to `previous_traces`?

text/0141-linking-traces.md
Lms24: Here's a link to the tracking issue: https://github.com/getsentry/sentry-javascript/issues/14478

text/0141-linking-traces.md
Lms24: I would argue that we should focus on span links as a mechanism to link **only to other spans**. Linking to other events is out of the otel spec as far as I can tell and something that would require some "API bending". This we can't do easily with POTEL SDKs so let's not focus on this too much. Does that sound reasonable to you?

text/0141-linking-traces.md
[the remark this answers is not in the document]
Lms24: Interesting! They might very well be serialized. I will re-check this!

text/0141-linking-traces.md
Lms24: I agree that having a consistent sampling decision across linked traces would make sense. There are some caveats though which we need to consider and deal with. I will discuss this with Ingest folks but will update the RFC ASAP.

* Different sampling mechanisms
  * Assuming, we have a unifom `tracesSampleRate`, quota usage should be consistent, regardless if we continue the sampling decision across traces or sample individually. So the advantage here would obviously be a more consistent user journey vs. only partial signals. 
  * However, if users specify a `tracesSampler` function, they could increase or decrease the sampling of certain routes or spans which means the number of events (i.e. quota usage) could drastically change as opposed to individual sampling. 
  * The `tracesSampler` function can overrule the sampling decision of upstream SDKs in case of continuing a remote trace. Maybe we can leverage the same mechanism to overrule sampling for previous traces? This is TBD though because there are discussions about no longer permitting overruling the remote sampling decision.
* Multi Page Applications
  * Let's assume we have a multi page app where each page navigation triggers a full page reload cycle. This means that for every page navigation, we'd have a new trace **starting on the server** that renders and returns the page. The problem here is that we cannot propagate the sampling decision of the previous trace to the next server-side request. That is, as long as we don't set a cookie (which is a line I don't think we want to cross). So there will be inconsistent sampling in MPAs.
  * We can still connect the individual page load spans on the client side and thereby connect the two traces but we can't enforce consistent sampling
  * We _could_ let users define their own cookie with some mechanism to work around this but I'd argue this is out of scope. 
  * So if we go with consistent sampling across traces, we'll need to accept this limitation.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Queues and Async Operations
+
+Somewhat related, we also face situations in which child spans are not started and finished within the time span of their parent span. As per OpenTelemetry specification, this violates the integrity of a trace. A concrete example for this is that we [currently recommend](https://docs.sentry.io/platforms/javascript/guides/node/tracing/instrumentation/custom-instrumentation/queues-module/) users to create one trace for a producer-consumer (Queue) cycle, where the spans from the consumer likely start after the producer spans finished. OpenTelemetry in fact [recommends](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to start separate traces for the consumer, producer or more generally async operations and to link these traces via a span link.
```
mjq, line 32: > Somewhat related, we also face situations in which child spans are not started and finished within the time span of their parent span. As per OpenTelemetry specification, this violates the integrity of a trace.

It doesn't really affect the proposal but I think that might be stated a bit too strongly - I've never found a place where it's forbidden by the spec for a child to outlive its parent, and the spec even [has an example ASCII doc showing such a case](https://opentelemetry.io/docs/specs/otel/overview/#traces). But, as you say, the convention does seem to be to use span links instead of parent/child relationships for async operations.

(My personal opinion is also that it's better modelling practice to use links for async children and I'd rather we go in that direction, but that's just me 😁)

text/0141-linking-traces.md
```diff
@@ -0,0 +1,368 @@
...
+
+## Out of scope/Non-Goals
+
+- Querying across linked traces: Given the "linked list" nature of one root span linking back to its previous root span, we accept that obtaining a complete list is an expensive query operation. Therefore, we consider use cases that require a complete list of all linked traces out of scope. We can address this as a follow-up if technically possible but we're aware that this is likely a sub-optimal data structure for such queries. 
+- Linking transactions started during the rendering of statically generated HTML pages and pageload transactions. A span link is ideal here because we don't want the hard link of a trace id. However, a casual link would still increase context and completeness. We might do it but not at this point as it's a follow-up item.
+- Unsampled traces are not linked. Errors across unsampled traces are not linked. For example, if the first trace was sampled negatively, the next trace will not link to it as this would result in a link to a non-existing trace.
```
mjq, line 62: The SDK can manage that for client sampling, but how does dynamic sampling affect this? 🤔

text/0141-linking-traces.md
mjq: I think whether turning linked transactions into a graph instead of a list makes sense partly depends on the product goals, since we haven't really outlined exactly what this is supposed to mean in the product. (As Lukas mentioned in another comment, that should be coming soon).

Lms24: I guess IOW, I would err towards accepting that linking back to previous traces might not be ideal for overlapping or parallel running traces. 

We should check though if this is a larger concern in mobile SDKs. @philipphofmann do you have insights on this?

mjq:
Sorry, looks like "starting a review" with both new comments and replies got weird and duplicated a bunch of stuff in the timeline above. Sorry 😬

text/0141-linking-traces.md
mjq: > However, while possible, we don't document overlapping traces today and we don't consider overlapping traces as "specc'd behavior". Setting this up in the browser for instance, would require significant manual intervention by users and I'm not aware of a lot of cases where this might happen.

FWIW this is defined and normal behaviour in OTel, so we will have to cross this bridge under OTLP if not sooner. (Imagine someone who instrumented their HTTP client spans in web, and nothing else: every one would start a trace, overlapping whenever the HTTP requests overlap.)
 
Given this proposal is primarily aimed at Sentry SDK users, and if our SDK won't have separate traces for things like parallel interactions or anything either, then I agree it is not currently a problem and so this alone shouldn't block anything here. 👍

text/0141-linking-traces.md
philipphofmann: Yes, that makes sense. Please put it in the RFC, and let's not add this as a pro.

text/0141-linking-traces.md
Lms24: This is a good point. I think similarly to parent span id's we cannot assume that just because a link to a span exists, said span also exists. Meaning, for a simple previous/next button scenario in the trace view, we'd actually need to search for the trace with the linked span before displaying the button(s). This ties in well with the EPD decision that the product has to handle incomplete traces. Or IOW, SDKs not linking to already SDK-dropped spans is not enough but should reduce the number of non-existing links.

text/0141-linking-traces.md
Lms24: I added a note about this in the RFC.

text/0141-linking-traces.md
Lms24: I addressed the sampling question in the "Answered questions" section.

text/0141-linking-traces.md
Lms24: I reworded this section. Thanks for clarifying! I must have misheard this somewhere then.

text/0141-linking-traces.md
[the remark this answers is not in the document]
Lms24: I rewrote this sections with some more concrete goals for the product

text/0141-linking-traces.md
Lms24: > What about changing the relationship of previous_trace from 1:1 to 1:n and renaming it to previous_traces?

sorry @philipphofmann almost missed this! My thinking is that `"previous_trace"` is fine because a span can have multiple links where each link has a `sentry.link.type: "previous_trace"` attribute. Does this work or am I missing something?

Lms24: I added a section about how SDKs should set the previous trace relation ship span link

text/0141-linking-traces.md
Lms24: I've removed the last point in this section, since we decided in our last call that the SDK will send a span link to a previously negatively sampled root span to indicate that there was a negatively sampled trace (see the diagram I added below).

cleptric (approved):
Let's do it.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+At the time of writing this RFC, Sentry (the product as well as SDKs) cannot connect multiple traces to deliver a "bigger picture" of what happened before (or after) a specific trace. 
+A trace is therefore the highest level of connecting signals (errors, traces, profiles, etc.).
+The notable exception is Session Replay, where we are indeed able to provide this bigger picture.
+Replay is an opt-in, additional product.
+
+This RFC proposes a method of achieving linkage between (and within) traces in the entire Sentry product, to satisfy a number of currently sub-optimally solved use cases.
+
+## Frontend Traces and User Journeys
+
+The most important application of linking traces are frontend applications. We would like to better display a user journey (session) to make debugging of issues easier as developers get more context on what happened before a specific issue.
```
s1gr1d, line 23: Just writing down a thought here: I think we have multiple definitions of "session" in our product that we should be aware of and be careful with in the communication. The ones I have on top of my mind: replay sessions, release health sessions and user journey sessions in tracing.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+
+This RFC proposes a method of achieving linkage between (and within) traces in the entire Sentry product, to satisfy a number of currently sub-optimally solved use cases.
+
+## Frontend Traces and User Journeys
+
+The most important application of linking traces are frontend applications. We would like to better display a user journey (session) to make debugging of issues easier as developers get more context on what happened before a specific issue. 
+Today, we are limited to the duration of one trace (id), which is handled differently and kept alive for different times across our various SDKs. 
+
+A concrete example are our [JavaScript Browser SDKs](https://develop.sentry.dev/sdk/platform-specifics/javascript-sdks/browser-tracing/#tracing-model) which by default keep a trace alive as long as users
+are on the same page or (URL) route. This trace model was a compromise in which we accepted that a trace would consist of multiple trace root spans (transactions) but provides more context. Long-lived traces, with multiple trace root spans, are generally discouraged by tracing models like OpenTelemetry.
```
s1gr1d, line 27: Maybe explain here what the downside or consequence of this was.
```suggestion
are on the same page or (URL) route. This trace model was a compromise in which we accepted that a trace would consist of multiple trace root spans (transactions) but provides more context. This sometimes resulted in very long traces .... Long-lived traces, with multiple trace root spans, are generally discouraged by tracing models like OpenTelemetry.
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+
+## Related: Queues, Async and Batch Operations
+
+Somewhat related, we also face situations in which child spans are not started and finished within the time span of their parent span. While generally supported by OpenTelemetry, such traces pose the question of  A concrete example for this is that we [currently recommend](https://docs.sentry.io/platforms/javascript/guides/node/tracing/instrumentation/custom-instrumentation/queues-module/) users to create one trace for a producer-consumer (Queue) cycle, where the spans from the consumer likely start after the producer spans finished. OpenTelemetry in fact [recommends](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to start separate traces for the consumer, producer or more generally async operations and to link these traces via a span link. Other related use cases are batch processes where one initiator trace triggers a large number of batch processes. Otel also recommends to link these spans instead in favour of grouping them under one trace.
```
s1gr1d, line 33: There is something missing :)
```suggestion
Somewhat related, we also face situations in which child spans are not started and finished within the time span of their parent span. While generally supported by OpenTelemetry, such traces pose the question of ..?..  A concrete example for this is that we [currently recommend](https://docs.sentry.io/platforms/javascript/guides/node/tracing/instrumentation/custom-instrumentation/queues-module/) users to create one trace for a producer-consumer (Queue) cycle, where the spans from the consumer likely start after the producer spans finished. OpenTelemetry in fact [recommends](https://opentelemetry.io/docs/concepts/signals/traces/#span-links) to start separate traces for the consumer, producer or more generally async operations and to link these traces via a span link. Other related use cases are batch processes where one initiator trace triggers a large number of batch processes. Otel also recommends to link these spans instead in favour of grouping them under one trace.
```

text/0141-linking-traces.md
s1gr1d: While reviewing, I found another related comment on this further down in this file: https://github.com/getsentry/rfcs/pull/141/files#r1856179314

text/0141-linking-traces.md
s1gr1d: I see this is explained further down below, maybe this can just be linked.

text/0141-linking-traces.md
Lms24: I removed the "session" reference in this section. As you pointed out, I described the two current sessions in the SDK further below, where we also point out that trace links are not another session but upon leadership decision an alternative to a 3rd session.

text/0141-linking-traces.md
Lms24: linked, thanks!

text/0141-linking-traces.md
Lms24: Good catch, I reworded this section. Thanks! :)

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+- Sampling and quota increase: If a trace is sampled positively, all transactions within this trace will be sampled positively. This also means that while the positively sampled trace is ongoing, we'd always propagate the positive sampling decision via the `sentry-trace` header to downstream services on outgoing requests. Assuming these services are configured to only sample already sampled traces, we'd send more events on an org-wide basis than if we rolled the dice more often (like in short-lived traces). This [increases quota usage](https://github.com/getsentry/sentry-ruby/issues/2318) for our customers without them necessarily being aware of it. 
+ 
+If we can link traces, we can ultimately shorten individual trace lengths and remove most of these disadvantages again.
+
+To be clear, the Sentry product will always need to support long-lived/multiple root span traces. While we can remove them from SDKs supporting this model, older SDK versions will stick around and users can also manually create long-lived traces.
+
+# Options Considered
```
smeubank, line 179: i'm missing somewhere here a short bullet list of titles of options considered. A lot to digest, and wasn't entirely clear from ## vs ### markdown which were options vs sub features like setting previous trace

```suggestion
# Options Considered

1. [Preferred] [Span Links](#preferred-span-links)
2. [Previous Trace Id](#alternative-previous-trace-id)
```

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+- Span links are an established concept in OpenTelemetry. With Sentry increasingly adapting Otel standards, as well as creating an OTLP endpoint in the future, we should embrace Otel-suggested solutions rather than building our own.
+- Attributes are not indexed in our current data storage solution, meaning queries like getting the "next" instead of previous traces are likely impossible or (best case) require the same level of indexing as span links require to show next traces.
+- POtel SDKs, as well as the browser JS SDKs, already expose span link APIs but they currently no-op and are not sent to Sentry. This is currently very opaque behavior and potentially confusing to users familiar with Otel APIs concepts.
+
+# Answered Questions
+
+- **Can we do this with our current event storage? Are we blocked on EAP?**
+
+  We will certainly need to adapt our current event storage but it is possible to do this without being blocked on EAP. We'll separately also need to adjust the planned span schema for EAP but this is not a blocker but a follow-up task after the MVP implementation.
```
smeubank, line 430: any concern to NOT do something like this now? ie if we wait and after EAP is delivered, will there be a lot of complexity to fit in a proper linked traces solution?

smeubank (approved):
I would go with your preferred option. Great write up, thorough and detail. Glad the session topic was address as well.

One option that is not here, is the "do nothing". i suppose, but maybe that is pedantic.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+
+![](./0141-previous-trace-partial-sampling.png)
+
+It's worth noting that we will not link to the previous positively sampled trace if a negatively sampled trace is in-between (see Traces 2-4 in the diagram). Furthermore, we will not show _how many_ traces were negatively sampled in between two trace chains; only that there was at least one trace in between (see Trace 5-8). 
+
+There are some ideas how we could show multiple negatively sampled spans as well as how we could connect traces with gaps due to sampling. For now, we'll disregard this because there's no concrete use case, yet. However, given spans can have multiple links and links can have attributes, we could add this information later on, if we deem it necessary.
```
philipphofmann, line 369: Most customers don't sample at 100%. They will frequently experience previously negatively sampled traces. I can imagine that some customers would like to navigate to the last positively sampled trace when most of their previous traces are negatively sampled. But it might be OK to wait for customers to ask for that feature. The current proposal would allow multiple options to achieve this without changing the APIs and the protocol.

text/0141-linking-traces.md
```diff
@@ -0,0 +1,447 @@
...
+
+### Implementation plan: 
+
+For an MVP and to validate that trace links work, we'll introduce span links on today's transaction events. In a future step, we'll ensure that this also works on our EAP storage system.
+
+The rough MVP implementation order:
+
+1. Ensure Storage can store along span links in `transaction` events today. If not, make necessary changes. Create index/lookup table so that we can query the "next" trace of a given root span/transaction.
+2. Ensure Relay can forward `transaction` events with span links. If not, make necessary changes. Specify types, expected payload structure and discard malformed links.
+3. Implement PoC in SDK, most probably the JS SDK, to send `transaction` envelopes with span links for previous traces
```
philipphofmann, line 408:
```suggestion
3. Implement the PoC in one SDK, most probably the JS SDK, to send `transaction` envelopes with span links for previous traces
```

philipphofmann (approved):
LTGM :shipit:

While this doesn't solve all of our long-lived trace problems, and I would love to see traces linked to sessions, this is a significant step in the right direction. This concept will solve multiple issues you thoroughly described in the RFC. It's also excellent that OTel already has this concept

Thanks for the detailed write-up. I'm looking forward to implementing this.

text/0141-linking-traces.md
Lms24: Yup. I sketched out a solution a while ago how we could add multiple links to indicate missing traces due to sampling: 
image

I didn't include this in the RFC because as I wrote, it's something we can tackle if there's demand for it.

text/0141-linking-traces.md
Lms24: No concerns to do this now, as far as I can tell!

text/0141-linking-traces.md
Lms24: done :)

PR #143, opened 2024-11-12
rfc(feature): SDK Fail-Safe Mode

philipphofmann:
This RFC aims to minimize the damage of a crashing SDK in production. It’s better to have no data than crashes and no data.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/sdk-fail-safe-mode/text/0143-sdk-fail-safe-mode.md)


armcknight:
Exciting to see this being discussed! The scenario analysis is looking good 👍🏻 It might be worth moving options 2 and 3 to a separate RFC as "remote configuration" as that's a different can of worms (there are several docs in Notion about this as a potential feature set) and keep 1 and 4 here.

philipphofmann:
>  It might be worth moving options 2 and 3 to a separate RFC as "remote configuration" as that's a different can of worms (there are several docs in Notion about this as a potential feature set) and keep 1 and 4 here.

Thanks for the input. I still want to keep them in this RFC as it helps us understand how to tackle all different scenarios. If we decide we want to move forward with these, we need to open another RFC to clarify the implementation details because, as you said, that's a different can of worms.

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,302 @@
...
+The SDK implements a retry logic to minimize the risk of wrongly disabling itself. When the app launched x times, the SDK retries if it can launch successfully. If it does, it goes back to normal. If it doesn't, it exponentially increases the number of app launches until it retries.
+
+The specification is written in the [Gherkin syntax](https://cucumber.io/docs/gherkin/reference/). The specification might not work for all edge cases yet, as it can be complicated to get it right. We'll figure out the exact details once we decide to implement it, but it should cover the main scenarios.
+
+```Gherkin
+Scenario: SDK version inits the first time
+    Given there is no launch marker file for the SDK version
+    And there is no success init marker file for the SDK version
+    When the SDK inits
+    Then the SDK creates a launch marker file for the SDK version
```
brustolin, line 64: m: The only thing I want to suggest is to not use more than one file. The SDK is either in a broken state (file exists), or not (file does not exist). After a successful initialization, we just delete the file marker.

armcknight: fwiw the previous implementation of this i built just used user defaults. one key, whatever value you need to represent the last state reached during initialization.

brustolin: Im my tests, `UserDefaults` is no longer (Apple changed something at some point) fast enough. It has a save process that is no synchronous anymore.

philipphofmann: Thanks for pointing that out; I already included this in the RFC: https://github.com/getsentry/rfcs/blob/a66a5b1747712ede252c4222b58ddffe564ee047/text/0143-sdk-fail-safe-mode.md#L167

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+1. Roughly doubles the size of the SDK.
+2. It requires an extra package.
+3. Only a subset of customers might use this, and only high-quality aware customers might accept the tradeoff of a double-sized SDK. In fact, most high-quality aware customers most likely care about app size and will use the stable release channel.
+
+## C: Knowing When the SDK is Disabled <a name="c-knowing-when-the-sdk-is-disabled"></a>
+
+### Option C1: Failing SDK Endpoint <a name="option-c1"></a>
+
+Add a unique endpoint for sending a simple HTTP request with only the SDK version and a bit of meta-data, such as the DSN, to notify Sentry about failed SDKs. We must keep this logic as simple as possible, and it should hardly ever change to drastically minimize the risk of causing more damage. The HTTP request must not use other parts of the SDK, such as client, hub, or transport. The SDKs must only send this request once. As we can’t have any logic running, such as rate-limiting or client reports, it’s good to have a specific endpoint for this to reduce the potential impact on the rest of the infrastructure.
```
kahest, line 433: > As we can’t have any logic running, such as rate-limiting

I know there are risks and limitations, but I'm thinking even a hard-coded e.g. 0.1 sample rate could help to reduce load by 90% and still give a useful signal

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+#### Pros <a name="option-c1-pros"></a>
+
+1. We know when a SDK disables itself.
+
+#### Cons <a name="option-c1-cons"></a>
+
+1. Potential risk of crashing while performing this action.
+2. It requires extra infrastructure.
+3. We don't know why the SDK disabled itself.
```
kahest, line 443: do you think there's a reasonable chance that there will be crash reports in Sentry that (at least via manual investigation and comparing timelines) could help find out more about the root cause?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+#### Pros <a name="option-c2-pros"></a>
+
+1. We don't fly completely blind when the SDK is in NoOp Mode.
+2. We reduce the risk of falsely staying in the NoOp or Safe Mode.
+
+#### Cons <a name="option-c2-cons"></a>
+
+1. The SDK might cause extra crashes, which could confuse users.
+
+### Option C3: Anomaly Detection <a name="option-c3"></a>
```
kahest, line 458: much to the pros mentioned below, this could be **very** useful as a signal, independently of this RFC and whether it's caused by SDK crashes or other 🔥

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented:
+
+1. False positives: Potentially wrongly disabling the SDK.
+2. Introducing new crashes with the new logic.
+
+# Unresolved questions
+
+- How do we test and rollout the new logic?
```
kahest, line 510: 💯

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+# Drawbacks
+
+Why should we not do this? What are the drawbacks of this RFC or a particular option if
+multiple options are presented:
+
+1. False positives: Potentially wrongly disabling the SDK.
+2. Introducing new crashes with the new logic.
+
+# Unresolved questions
```
kahest, line 508: one question that remained open for me is - do we allow configuration of any of this? to what degree? e.g. number of retries after reverting back from safe mode, etc.
IMO we will definitely want a config option to completely disable all of this, at least in the beginning (will at first probably be opt-in). we'll likely figure out if there's more we want to make configurable along the way

kahest (commented):
this is extremely thorough and well thought-out, thank you @philipphofmann 

I left some notes/thoughts/questions and minor corrections, but no blockers for aligning on a decision

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+#### Cons <a name="option-b2-cons"></a>
+
+1. We stop sending data due to a false positive.
+
+### Option B3: [Discarded] Remote Kill Switch <a name="option-b3"></a>
+
+> 🚫 **Discard reason:** It makes sense to consider this option once we implement a remote config for all SDKs, but it's too much effort as a standalone feature. 🚫
+
+There might be scenarios where the SDK can’t detect it’s crashing. We might be able to detect via the SDK crash detection that the SDK causes many crashes, and we could manually or, based on some criteria, disable the SDK. We could also allow our customers to disable the SDK remotely if they see many crashes in the Google Play Console or App Store Connect. It
```
buenaflor, line 390:
```suggestion
There might be scenarios where the SDK can’t detect it’s crashing. We might be able to detect via the SDK crash detection that the SDK causes many crashes, and we could manually or, based on some criteria, disable the SDK. We could also allow our customers to disable the SDK remotely if they see many crashes in the Google Play Console or App Store Connect.
```

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+multiple options are presented:
+
+1. False positives: Potentially wrongly disabling the SDK.
+2. Introducing new crashes with the new logic.
+
+# Unresolved questions
+
+- How do we test and rollout the new logic?
+- Should we surface the info of a failing SDK information to the user?
+- How does the failing SDK endpoint look like? Maybe we wan't to replace it with a special type of envelope with specific metadata so the load balancer can discard envelopes in the worst case.
```
buenaflor, line 512:
```suggestion
- How does the failing SDK endpoint look like? Maybe we want to replace it with a special type of envelope with specific metadata so the load balancer can discard envelopes in the worst case.
```

buenaflor (commented):
looks good! added some comments

JoshuaMoelans (commented):
LGTM! I added some small nits & have one question. Thanks for writing this up!

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+Scenario: New App version inits with previous failed init
+    Given there is a start init marker file for a previous app version
+    And there is no success init marker file for the previous app version
+    And there is no success init marker file for the current app version
+    When the new app version inits
+    Then the SDK deletes marker files from the previous app version
+    And the SDK creates a start init marker file for the current app version
+```
+
+As we must access checkpoint information during the application launch, we must choose an efficient way to read and write this information to not slow down our users' apps. Depending on the platform, we can use marker files or key-value stores such as [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) on Apple or [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) on Android. Marker files are more efficient than reading file contents because, for these, the OS only needs to check the file's existence, which is usually a system metadata look-up. We still need to determine which approach is the most efficient.
```
markushi, line 244:
```suggestion
As we must access checkpoint information during the application launch, we must choose an efficient way to read and write this information to not slow down our users' apps. Depending on the platform, we can use marker files, key-value stores such as [UserDefaults](https://developer.apple.com/documentation/foundation/userdefaults) on Apple or [SharedPreferences](https://developer.android.com/reference/android/content/SharedPreferences) on Android or System APIs like [ApplicationExitInfo on Android](https://developer.android.com/reference/android/app/ApplicationExitInfo). Marker files are more efficient than reading file contents because, for these, the OS only needs to check the file's existence, which is usually a system metadata look-up. We still need to determine which approach is the most efficient.
```

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+This section looks at the three different problems we need to solve:
+
+1. [Detecting Continuous SDK Crashes](#a-detecting-continuous-sdk-crashes)
+2. [Minimizing the Damage](#b-minimizing-the-damage)
+3. [Knowing When the SDK is Disabled](#c-knowing-when-the-sdk-is-disabled)
+
+## A: Detecting Continuous SDK Crashes <a name="a-detecting-continuous-sdk-crashes"></a>
+
+First, we need to know when our SDKS continuously crash our customers. Only then can we act accordingly. We categorize crashes into four different time categories:
```
romtsn, line 120:
```suggestion
First, we need to know when our SDKSs continuously crash our customers. Only then can we act accordingly. We categorize crashes into four different time categories:
```

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+    Then the SDK creates a start init marker file for the app version
+
+Scenario: App version inits the first time with successful init
+    Given there is no start init marker file for the app version
+    And there is no success init marker file for the app version
+    When the SDK reaches the successful init checkpoint
+    Then the SDK creates a success init marker file for the app version
+
+Scenario: App version inits the first time with failed init
+    Given there is no start init marker file for the app version
```
romtsn, line 207:
```suggestion
    Given there is a start init marker file for the app version
```
was this supposed to be like that? Or is this a different scenario?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+3. It is the most reliable option to detect if the SDK crashes.
+
+#### Cons <a name="option-a3-cons"></a>
+
+1. __Doesn't work with static linking:__ This approach doesn’t work with static linking, as the Sentry SDKs end up in the same binary as the main app. As we don’t have symbolication in release builds, we can’t reliably detect if the memory address stems from the Sentry SDK or the app. We might be able to compare addresses with known addresses of specific methods or classes, but this won’t work reliably. As with iOS, many apps use static linking, so we must use an alternative approach.
+2. __Doesn't work for obfuscated code:__ For obfuscated code, detecting if a frame in the stacktrace stems from the Sentry SDK or the app can be difficult or even impossible.
+3. __Wrongly disabling the SDK:__ We frequently see wrongly reported SDK crashes in the SDK crash detection. As SDKs use bytecode manipulation, swizzling, or monkey patching, the stacktraces sometimes contain Sentry frames in the crashing thread, but the root cause isn't Sentry but the user's code.
+4. It doesn't work when the SDK crashes during or before sending the crash report.
```
romtsn, line 355: just an idea crossed my mind: we could easily solve 4 and 5 cons here by installing another internal crash handler (if possible) as the first thing when initializing the SDK, whose sole purpose would be to identify SDK crashes and disable the SDK/let us know about/etc.

This way we could also checkmark the 1st scenario here and ultimately be the best solution? 😅 but the problem with static linking still remains ofc.

Maybe we could think of this as another potential solution or apply it to some other one.

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+2. We reduce the risk of falsely staying in the NoOp or Safe Mode.
+
+#### Cons <a name="option-c2-cons"></a>
+
+1. The SDK might cause extra crashes, which could confuse users.
+
+### Option C3: Anomaly Detection <a name="option-c3"></a>
+
+The backend detects anomalies in our customers' session data. If there is a significant drop, we can assume that the SDK works in NoOp mode. The logic has to correctly detect debug and staging releases and take sampling into account.
+
```
romtsn, line 461: Another idea crossed my mind (idk if we want to add it as an idea or an option here): we could query our customers' google play console/appstore connect via a public API (like an integration) and then look up for our own crashes in the dataset and ingest those into the SDK crash reporting projects and get notified, etc etc. I know it's [possible](https://developers.google.com/play/developer/reporting/reference/rest/v1beta1/vitals.errors.issues) with the google play console API, not sure about the [apple](https://developer.apple.com/documentation/appstoreconnectapi/downloading-analytics-reports) one though.

kahest: 💯 similar to the existing option C3 this could provide a lot of value, even independently of the current RFC. I looked into this a bit for Apple though, and don't think there's a way (at least not a remotely official one). somehow Xcode needs to access this information from the Apple servers, but that's as closed as can get I believe

philprime: I've worked with the [official App Store Connect API](https://developer.apple.com/documentation/appstoreconnectapi/) quite a lot, and also with [spaceship](https://github.com/fastlane/fastlane/blob/master/spaceship/README.md) which uses unofficial API endpoints used in the App Store Connect web app.

In general, if we need data that is not accessible via public unauthenticated endpoints, the ASC API requires an API key created by the customer, or we need to have access to ASC via an Apple Account (2FA is required).

It's quite complicated to do, but it can be done.

kahest: maybe a holiday read for me 😅 https://developer.apple.com/documentation/appstoreconnectapi/power-and-performance-metrics-and-logs

philprime: I have not used the Power and Performance Metrics API before, but it seems to have been available since API version v1.2. 

The last time I have checked crash reports reported to Testflight are still not available in the ASC API, but it can be downloaded with [spaceship.get_beta_feedback](https://github.com/fastlane/fastlane/blob/400f21481aa1b0331b5407e5c4f125ce4b217874/spaceship/lib/spaceship/connect_api/testflight/testflight.rb#L510-L513).

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+1. The SDK still works for crashes if there is a critical bug in most areas of the SDK. For example, if a bug in session replay continuously crashes the app shortly after SDK initialization, the SDK will still report crashes.
+
+#### Cons <a name="option-b1-cons"></a>
+
+1. If the crash occurs in the safe mode, the SDK causes an extra crash before disabling itself.
+
+### Option B2: NoOp SDK Init <a name="option-b2"></a>
+
+The SDK makes the SDK init a NoOp (no operation) when it detects a continuous crash with one of the options of [A](#options-for-detecting-sdk-crashes). This is the last resort and we should try to avoid this as much as we can, but it's better to have no data than crashes and no data. To minimize the risk of staying wrongly in NoOp mode and to avoid flying completely blind on the root cause, the SDK keeps track of how often the SDK was started in NoOp mode and after x times it retries to initialize.
```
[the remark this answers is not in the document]
philipphofmann: Yes, makes sense. I updated the RFC.

text/0143-sdk-fail-safe-mode.md
philipphofmann: I updated this. I have now changed it to a special envelope item type. As the infrastructure can already handle session init updates and SDKs only send this info once, it might be OK to skip even rate limiting and not have an extra endpoint. We still need to clarify this with the ingest team.

text/0143-sdk-fail-safe-mode.md
philipphofmann: Good point. There is a slight chance that the SDK can send a crash report on a previous attempt. We could add the user ID to the envelope to investigate related crashes.

text/0143-sdk-fail-safe-mode.md
philipphofmann: A configuration adds complexity, and we must avoid complexity. I would only allow a boolean flag to turn it on or off.

philipphofmann: I added this to a new FAQ section.

text/0143-sdk-fail-safe-mode.md
philipphofmann: That scenario doesn't make any sense. I removed it, thanks for pointing it out.

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+#### Scenario 3: User's Application Crashes Shortly After SDK Initialization
+
+The user's application crashes shortly after the SDK initialization. We have to ensure that we're not wrongly disabling the Sentry SDK. Still, suppose we detect that the app continuously crashes after x seconds of the Sentry SDK initialization. In that case, switching to the SDK Safe Mode might be acceptable to minimize the risk of the Sentry SDK being the root cause. Furthermore, our users will mainly be interested in the crash events, not other data such as performance or session replay.
+
+#### Scenario 4: User's Application Crashes After SDK Initialization
+
+This happens frequently, and we must ensure that the SDK correctly ignores this scenario.
+
+### Option A1: [Preferred] Checkpoints <a name="option-a1"></a>
```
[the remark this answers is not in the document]
philipphofmann: I thought that maybe we could solve that later, but I think we must consider it. I quickly gave it some thought, but as it's not trivial I will add this later.

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
...
+
+#### Scenario 3: SDK Continuously Crashing Shortly After SDK Init <a name="continuous-crash-scenario-3"></a>
+
+Finally, a bad-case scenario is our SDK crashing continuously shortly after the app start. The app might be unusable as it constantly crashes at a specific area, or only certain features stop working. The Sentry SDK should still be able to send a crash report, so the SDK crash detection should surface this, and our users can see the crashes in their data. If the SDK can't send a crash report, this scenario can either turn into scenario 1, where sending a crash report continuously crashes, or the SDK can't send crash reports, which is also bad but out of the scope of this RFC. Similar to scenario 2, our users must release a hotfix, and they could lose trust, but it is better than [scenario 1](#continuous-crash-scenario-1).
+
+#### Scenario 4: SDK Continuously Crashing After SDK Init
+
+Similar to [scenario 3](#continuous-crash-scenario-3), but the SDK crash happens after the SDK init.
```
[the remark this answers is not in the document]
philipphofmann: I updated the scenario. Is it clear now, @JoshuaMoelans?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,512 @@
+- Start Date: 2024-11-12
+- RFC Type: feature
+- RFC PR: https://github.com/getsentry/rfcs/pull/143
+- RFC Status: draft
+
+# Summary
```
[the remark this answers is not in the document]
philipphofmann: Yes, it reduces the problem space. Thanks for pointing this out. 💯

text/0143-sdk-fail-safe-mode.md
philipphofmann: Thanks for the feedback Roman 😃. I added to the description of this option that the SDK has to install the crash handlers as soon as possible. I changed con 4 to `before writing the crash report` not `sending`, and I removed con 5. Does that work for you @romtsn?

text/0143-sdk-fail-safe-mode.md
philipphofmann: I added this as idea 4. Thanks for the input @kahest, @romtsn and @philprime 😃.

text/0143-sdk-fail-safe-mode.md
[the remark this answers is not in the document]
JoshuaMoelans: yes perfect, thanks for clarifying @philipphofmann !

text/0143-sdk-fail-safe-mode.md
romtsn: yes, thanks!

cleptric:
Can't this be a regular event envelope with a special context?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,524 @@
...
+
+We want to keep the logic as simple as possible to reduce the risk of bugs, and it could be that we disable the wrong integration, which would cause more unnecessary crashes.
+
+**2. Why do we focus mostly on crashes during and after SDK init?**
+
+As described in the [motivation](#motivation), continuous SDK crashes during the app start are more critical because they cause the most friction by making the application unusable.
+
+**3. Do we allow users to opt-out of this feature?**
+
+Yes, we can add a flag to the SDK to opt-out of this feature.
```
adinauer, line 505: I presume opting out means the SDK behaves the same way it currently does (without this RFC), correct?
```suggestion
Yes, we can add a flag to the SDK to opt-out of this feature, meaning the SDK will behave as it did before changes discussed in this RFC were implemented.
```

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,524 @@
...
+
+**4. Do we allow a configuration for number of retries for our users?**
+
+No, a configuration adds complexity and we must avoid it. We only allow to enable or disable the feature.
```
adinauer, line 509: Do we want to allow configuration by customers on whether Safe Mode should be used or whether the SDK should go into No-Op directly?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,524 @@
...
+    decision-sdk-mode-- failed safe init --> sdk-init-no-op
+
+
+    decision-sdk-init-x-times-safe-mode-- yes --> sdk-init-normal
+    decision-sdk-init-x-times-safe-mode-- no --> sdk-init-safe-mode 
+
+    sdk-init-no-op --> sdk-fail-envelope
+```
+
+**What about Hybrid SDKs?** This logic works independently for every SDK and doesn't work across multiple SDKs. Implementing such a logic across different hybrid SDK layers would significantly increase the complexity. We have to remember that there are multiple variations of hybrid SDK setups. Some users manually initialize the native SDKs before or after initializing the hybrid SDKs. Getting the fail-safe logic correct logic across all variations of hybrid SDK layers is a nightmare we don't want to deal with. It is permitted to have a hybrid SDK running with a normal SDK init and the underlying native SDK in safe mode or vice versa.
```
JoshuaMoelans, line 110: (nit) typo:
```suggestion
**What about Hybrid SDKs?** This logic works independently for every SDK and doesn't work across multiple SDKs. Implementing such a logic across different hybrid SDK layers would significantly increase the complexity. We have to remember that there are multiple variations of hybrid SDK setups. Some users manually initialize the native SDKs before or after initializing the hybrid SDKs. Getting the fail-safe logic correct across all variations of hybrid SDK layers is a nightmare we don't want to deal with. It is permitted to have a hybrid SDK running with a normal SDK init and the underlying native SDK in safe mode or vice versa.
```

text/0143-sdk-fail-safe-mode.md
philipphofmann: No, I don't have that planned to keep it simple. It's all or nothing. Do you think we need that? If yes, what is the use case?

adinauer: I was just thinking about how many chances I'd give an app that crashes on startup and it might be nice to let customers choose their priority (i.e. disable Sentry on a startup crash). However we can revisit if customers actually ask for this.

romtsn:
> > Can't this be a regular event envelope with a special context?
> 
> Yes, I actually like that more than an extra envelope item. A simple plain event with some context or something should be enough. Then users would see it in Sentry, and we can filter in the SDK crash detection for it and even create alerts. Thanks for the suggestion @cleptric, I added this in [7afaf13](https://github.com/getsentry/rfcs/commit/7afaf1359f48806edd95c37f12f76729c88c751c).

The incentive was to not go through our envelope serialization pipeline at all. IIRC that was the problem that triggered this initiative in the first place, right? If we have a bug in the envelope serilization logic this will do no good.

cleptric:
@romtsn adding a new event type to our ingestion pipeline to by-pass certain logic in an SDK is bit overkill imo. You can still by-pass it in the SDK and create a normal event (error) envelope manually.

philipphofmann:
@romtsn and @cleptric I added this to the option in https://github.com/getsentry/rfcs/pull/143/commits/066479570ae7ec17a9637a41051ad37f8f1ace74
```
The event envelope doesn't go through the scope/hub or client. Instead, the payload for the event envelope is hard-coded, and the SDK directly sends the HTTP payload to Sentry.
```
Is it OK now @romtsn?

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,543 @@
...
+#### Pros <a name="option-a2-pros"></a>
+
+1. It already exists.
+2. It correctly ignores all potential false positives.
+3. It works for all SDK crash scenarios when the SDK can send a crash report.
+
+#### Cons <a name="option-a2-cons"></a>
+
+1. It doesn't work for SDK init crashes.
+2. It doesn't only works for single tenants or self-hosted.
```
philprime, line 312:
```suggestion
2. It doesn't work for single tenants or self-hosted.
```

philprime (approved):
Left one comment, other than that seems fine to me

markushi (approved):

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,543 @@
...
+
+1. Detecting SDK bugs such as missing data or broken functionality.
+2. Implementing some health checks for parts of the SDK, such as integrations.
+
+# Recommended Approach
+
+The recommended approach solves the three problems mentioned [above](#background):
+
+1. [A: Detecting a continuously crashing SDK](#a-detecting-continuous-sdk-crashes) with [Checkpoints](#option-a1).
+2. [B: Minimizing the damage of a continuously crashing SDK](b-minimizing-the-damage) with [Safe Mode](#option-b1) and [NoOp Init](#option-b2).
```
JoshuaMoelans, line 54:
```suggestion
2. [B: Minimizing the damage of a continuously crashing SDK](#b-minimizing-the-damage) with [Safe Mode](#option-b1) and [NoOp Init](#option-b2).
```

text/0143-sdk-fail-safe-mode.md
```diff
@@ -0,0 +1,543 @@
...
+
+Why didn't the [SDK crash detection](https://github.com/getsentry/sentry/tree/master/src/sentry/utils/sdk_crashes) help us to identify the issue? For those who do not know, the SDK crash detection allows us to view SDK crashes in Sentry. It checks every ingested event in the processing pipeline for a potential SDK crash. When detecting one, it only keeps the most essential data due to PII and copies the crash event to dedicated Sentry projects. The SDK crash detection helps fix crashing SDK bugs. Still, it only works when the SDK crash events make it to Sentry, which wasn't the case in the abovementioned incident.
+
+To identify and minimize the damage of such incidents, we need to develop extra strategies. To get there, we have to solve three different problems:
+
+1. [**A: Detecting a continuously crashing SDK**](#a-detecting-continuous-sdk-crashes): First, we need to know when our SDKS continuously crash our customers.
+2. [**B: Minimizing the damage of a continuously crashing SDK**](b-minimizing-the-damage): It's better to have no data than crashes and no data.
```
JoshuaMoelans, line 23:
```suggestion
2. [**B: Minimizing the damage of a continuously crashing SDK**](#b-minimizing-the-damage): It's better to have no data than crashes and no data.
```

JoshuaMoelans (approved):
small nit on in-doc links, but LGTM content-wise!

kahest (approved):

krystofwoldrich (approved):

buenaflor (approved):

text/0143-sdk-fail-safe-mode.md
philipphofmann: > However we can revisit if customers actually ask for this.

Yes indeed.

PR #144, opened 2025-01-19
fix: View Hierarchy positioning is relative

timfish:
It was not mentioned in the RFC but View Hierarchy positioning is expected to be relative to the parent window. 

In Sentry, flutter projects were hard-coded to use absolute positioning but it was not possible to otherwise override this. 

I opened a PR to allow overriding this by including `absolute` in the `rendering_system`:
- https://github.com/getsentry/sentry/pull/83292


timfish:
Another alternative might be to add a top level `positioning` property which can be `relative` or `absolute` and defaults to `relative` if not set?

markushi (approved):
Sounds good to me, thanks!

PR #147, opened 2025-05-08
rfc(decision): Make Sentry Semantic Conversions Mandatory in SDKs

antonpirker:
We should make sure, that it is hard to add span attributes that are not listed in our Semantic Conversions.
This should make it impossible to just randomly add attributes that are not documented anywhere.

[Rendered RFC](https://github.com/getsentry/rfcs/blob/rfc/make-sentry-semantic-conversions-mandatory-in-sd-ks/text/0147-make-sentry-semantic-conversions-mandatory-in-sd-ks.md)


antonpirker:
As `set_attribute()` is a public API and designed to be used by the user to add arbitrary attributes to spans, this RFC is not going to work. I withdraw my request.

PR #152, opened 2026-02-27
rfc(feature): SDK-symbolicated stack-frames

supervacuus:
This is a feature RFC to find a solution for SDK/client-side symbolicated stack frames:

[Rendered RFC](https://github.com/supervacuus/rfcs/blob/rfc/sdk-symbolicated-frames/text/0152-sdk-symbolicated-frames.md)


Dav1dde (commented):
Overall makes sense. 

We definitely shouldn't re-Use `symbolicator_status`, this field is not part of the Event ingestion, it's purely a symbolicator/sentry field (like you call out).

Inferring from other data seems not too bad, but an explicit signal is probably preferred and much less confusing.

There is also an option of re-using the platform for this, since that already controls symbolication. Either by extending it, replacing it or setting a magic `noop` (not really a fan of this).

supervacuus:
Thanks for taking the time, @Dav1dde!

> There is also an option of re-using the platform for this, since that already controls symbolication. Either by extending it, replacing it or setting a magic noop (not really a fan of this).

Could I ask you to elaborate on what this would look like in practice? 

Do you mean in the sense that the `platform` for _each frame_ carries the logic for handling already symbolicated frames that failed symbolicator resolution in `lang/native/processing.py`? Because that would fall into the category of Option B for me ("Infer from existing frame data"  is long for: handle in the backend).

Or do you mean the `platform` event payload entry should be extended to signal from the SDK to processing how particular frames would be handled? What would be the benefit of that vs per-frame properties?

Dav1dde:
So as an example, I think if you were to set the `platform` on a frame (overriding the default platform) to `noop` it would already skip symbolication of that frame.

Now obviously this isn't the greatest idea, but maybe there is a world where we go away from `platform` and instead define a field `symbolicate_as`.

Examples:
- `symbolicate_as: native` <- Apply native symbolication
- `symboliicate_as: js` <- This frame must have source maps (e.g. react native stack traces)
- `symbolicate_as: jvm` <- Proguard etc.
- `symbolicate_as: none` <- Skips symbolication

We currently infer the Symbolication decision from the platform, but we could also make it explicit. This is a pretty unfiltered thought and definitely needs a lot more thought and bikeshedding. I think I'd prefer the flag over this, but possibly worth a discussion.

supervacuus:
> We currently infer the Symbolication decision from the platform, but we could also make it explicit.

Yes, this is a relevant point I didn't explicitly highlight (edit: [added](https://github.com/supervacuus/rfcs/blob/rfc/sdk-symbolicated-frames/text/0152-sdk-symbolicated-frames.md#the-platform-fields-dual-role)): we also tie the rendering of each frame to the `platform`, which in turn leads to entirely different outcomes wrt "missing symbol/image" signalling.

That is also why I thought RFCing this makes sense: there are good reasons to do it entirely server-side as well as SDK-side.

The fact that the SDK-set `platform` _currently_ implicitly carries how a frame is both **rendered** and **symbolicated** might be a good hint for the general direction this RFC could take: do we want to keep this implicit marker approach for both, or break it (partially/fully) open towards the SDK? And of course, what are the ups/downs of each?

> This is a pretty unfiltered thought and definitely needs a lot more thought and bikeshedding. I think I'd prefer the flag over this, but possibly worth a discussion.

Agree and matches the format. Right now, I am looking for clarity in the proposal (and counterproposals, thx) and for issues I haven't thought of (or incorrectly framed). So, this is welcome feedback at that level of filtering.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,167 @@
...
+
+See also: [getsentry/sentry#46955](https://github.com/getsentry/sentry/issues/46955): "Remove error banner for non-app symbols"
+
+### Third-party library detection
+
+There is a `is_known_third_party()` check that suppresses missing-symbol errors for recognized system libraries (e.g., iOS system frameworks). This is a denylist approach and does not scale to arbitrary platforms or deployments.
+
+### Passing `symbolicator_status` from the SDK
+
+In theory, an SDK could set `symbolicator_status` directly on the frame's `data` dict. Since `data` is part of the stacktrace `Frame` and falls into relay's untyped "other" catch-all ([`relay-event-schema/.../stacktrace.rs` L200-202](https://github.com/getsentry/relay/blob/55c59cf75d3c35bbbb66df14072d147eca056bd7/relay-event-schema/src/protocol/stacktrace.rs#L200-L202)), it would technically be forwarded. However, using untyped catch-all fields for regular SDK usage is explicitly frowned upon and for good reason.
```
jjbayer: Currently it would be dropped by the processing relay because the catch-all does not have `retain = true`.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,167 @@
...
+
+Let SDKs explicitly set `symbolicator_status: "symbolicated"` (or a new value) in `frame.data`. Relay would need to explicitly allow this field from SDKs rather than treating it as backend-only.
+
+**Testing has shown that SDK-set values in `frame.data.symbolicator_status` are discarded or overwritten by the backend during processing.** This means Option C is not viable without changes to either relay (to stop stripping/ignoring the field) or processing (to check for an existing value before overwriting).
+
+**Changes required (if pursued despite the above):**
+
+- **relay-event-schema**: Make `symbolicator_status` a recognized field that SDKs may set, and ensure it is not stripped during ingestion.
+- **processing.py**: Check if `symbolicator_status` is already set before overwriting it with symbolicator's result.
+- **SDKs**: Set `data.symbolicator_status` on symbolicated frames.
```
jjbayer, line 377: I see an Option D where The SDK and symbolicator use a common format to annotate the symbolication status. This would be future-proof towards what software component does the symbolication. Something like

```json
"symbolication": {
    "by": "symbolicator", // or "client"
    "status": "success"
}
```

Later we could even extend the protocol to log multiple unsuccessful attempts, if needed:

```json
"symbolication": {
    "by": "symbolicator",
    "status": "success",
    "previous_attempts": [
        {"by": "client", "status": "failed", "reason": "unknown_image"}
    ]
}
```

text/0152-sdk-symbolicated-frames.md
supervacuus: Thank you, that backs the experimental outcome. I will rectify the description.

text/0152-sdk-symbolicated-frames.md
supervacuus: I can see how this could be useful from a diagnostic/audit perspective, and I also agree that it would be more in line with what `symbolicator_status` currently does: it records the outcome of symbolicator's handling of the frame.

However, I think this is almost orthogonal to my intention for this RFC (though it could still be in scope if we decide to implement it), and I might have communicated it poorly by introducing `symbolicator_status` as a potential solution to the problem (because it acts more as a counter pattern).

**These are the core questions for which I seek a design proposal from this RFC:**

* Can we let the SDK decide whether processing/symbolicator should handle a particular stack frame or not? 
* How far do we want to parameterize this (is this an on/off switch per frame, or do we want to open the encoding vs `platform` as described here: https://github.com/getsentry/rfcs/pull/152#issuecomment-3991710125 and what about symbolicator responsibilities that go beyond symbols, like source-code and line numbers, etc.)? 
* Should SDKs have that power or not? While the two questions above are very much "does the shoe fit now?", this one is about a trajectory: do we already have capabilities lined up that would be inhibited by pushing this towards the SDK?

So, purely from an intent perspective, I see your proposal as a record after the fact, whereas I am currently looking for a mechanism to control that outcome. Maybe I misunderstood. Also, if anything in this comment wasn't clear from the RFC, let me know whether and where I should extend it. Thanks!

supervacuus: Or, would you argue that this is also intended as a mechanism for control, because native processing could implement not to send a frame to `symbolicator` (or overwrite the client symbol if a symbolicaton request is attempted anyway) if a stack frame contains:

```JSON
 "symbolication": {
    "by": "client",
    "status": "success"
}
```

Then it could be both diagnostic and control. But we should decide on the control meaning getting such a symbolication tag from the SDK.

supervacuus: In any case, I integrated the feedback I got until now: https://github.com/getsentry/rfcs/pull/152/commits/bbd301cec0e18c5ce1239a194e69b7558b99a77e

As you can see, I just massaged your "Option D" into a control mechanism. Let me know if that aligns with what you had in mind.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+**Control semantics:** When processing encounters a frame with `symbolication.by == "client"` and `symbolication.status == "success"`, it skips symbolicator for that frame (analogous to the early-exit in Option A). If the status is `"failed"`, processing may still attempt server-side symbolication and record the outcome, optionally preserving the client attempt in `previous_attempts`.
+
+**Changes required:**
+
+- **relay-event-schema**: Add a typed `symbolication` object to `Frame` with fields `by: enum("client", "symbolicator")`, `status: enum("success", "failed", "missing", "unknown")`, and optionally `reason: string` and `previous_attempts: list`.
+- **SDKs**: Set `symbolication: { "by": "client", "status": "success" }` on frames the SDK has symbolicated.
+- **processing.py**: Check for `symbolication.by == "client" && symbolication.status == "success"` early in frame handling; if present, skip symbolicator. Otherwise, proceed normally and write the symbolicator result into the same `symbolication` field.
+- **symbolicator**: Write results into the `symbolication` object instead of (or in addition to) the current `symbolicator_status` field in `frame.data`.
```
jjbayer: Symbolicator could double-write into both fields for a while, until the sentry backend is updated to use the new field for a while.

text/0152-sdk-symbolicated-frames.md
jjbayer: > Or, would you argue that this is also intended as a mechanism for control

Yes, I would intend it mainly as a control mechanism which can also serve as a diagnostic tool or to give the user more visibility of what happened to their stack traces. Thank you for adding the option!

jjbayer (commented):
FWIW I think Option A would work perfectly fine. Option D is just an attempt to foresee potential future use-cases / unify the newly proposed protocol with the existing one.

chargome:
For JS this this would not have implications I think as there is no client-side symbolication apart from nextjs dev server. We can adapt out stackframe type definition if we land on a type here but would likely leave that undefined in all cases that we currently cover. 👍

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,272 @@
...
+- The dual role (control + diagnostic) may conflate concerns: the decision to skip symbolication and the record of what happened are conceptually distinct, and coupling them in one object may complicate future changes to either.
+
+# Unresolved Questions
+
+- How are the associated debug-meta images being resolved? If a frame is marked as "client-symbolicated" and thus doesn't show a "missing symbol" warning, images that are missing, but whose associated frames are entirely "client-symbolicated", shouldn't be marked as an error either (but should still be listed). Should this be resolved on the client (with a tag similar to the frame)? Or should this be resolved in native processing?
+- Are there workflow interactions between symbolication, line-number, and source lookup that would be prevented by the presented approach?
+- What is the interaction with demangling? If an SDK provides a mangled C++ symbol, should the backend still demangle it even if the frame is marked as client-symbolicated?
+- Should the UI distinguish between server-symbolicated and client-symbolicated frames (e.g., a subtle indicator), or treat them identically?
```
philprime: I believe surfacing the origin of the symbolication could be beneficial for debugging absent symbols - which happens easily in case CI workflows uploading debug symbols soft-fail to do so. By comparing symbolicated issues to issues without symbols with indication of the source could indicate if it's an SDK issue or not.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,272 @@
...
+- **.NET SDK**: The .NET SDK resolves function names, file paths, and line numbers locally using portable PDB metadata. When these frames arrive at the backend, symbolicator still attempts to process them. Because the user hasn't uploaded PDB files to Sentry (the SDK already handled this), every frame returns `symbolicator_status: "missing"` and the UI shows a misleading symbolication error banner. This is tracked in [getsentry/sentry#97054](https://github.com/getsentry/sentry/issues/97054).
```
Flash0ver, line 25: note: applies to JIT-compiled apps ... for Native AOT compiled apps, we only got function names, but I believe file names and line numbers are missing ... but Native AOT maybe a whole different story in this context

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,272 @@
...
+
+This is a cross-cutting concern that touches:
+
+- **SDKs**: All native code handling SDKs (sentry-native, sentry-java/Android NDK, sentry-dotnet, (maybe sentry-cocoa), and downstream dependants + any future SDK performing client-side native symbolication).
+- **Relay / ingestion**: The frame schema needs to either gain a new typed field or explicitly allow SDKs to set existing fields.
+- **Processing (monolith)**: `sentry/lang/native/processing.py` needs to honor the new signal and skip symbolication for marked frames.
+- **Symbolicator**: May need awareness of the flag if processing delegates the decision.
+- **UI**: Should stop showing misleading "missing symbols" errors for frames that are intentionally client-symbolicated.
+
+# Options Considered
```
Flash0ver: question: Do I understand it correctly that when both sides are given (uploaded debug files & SDK-runtime-resolved metadata), the Client-provided metadata is preferred over the previously uploaded symbols?

supervacuus: As part of this proposal's options, or how it is currently implemented?

text/0152-sdk-symbolicated-frames.md
supervacuus: > or Native AOT compiled apps, we only got function names, but I believe file names and line numbers are missing

Okay, but is the assumption that those filenames and line numbers can be resolved later because the user will upload them as debug info?

> but Native AOT maybe a whole different story in this context

How so?

jamescrosswell: I think it's pretty unlikely we'd be symbolicating AOT compiled code on the client (the subject of this RFC)... but we should be able to resolve symbols later.

AFAIK during AOT compilation, when converting the IL into assembly, the linker can create a map from the native instructions back to the original IL... and this map gets stored in the symbol file. So even though the app is AOT compiled and you no longer have the original IL, you can still symbolicate if you have the maps in the symbol files.

Testing this out by running `dotnet publish -c Release -f net10.0 samples/Sentry.Samples.Console.Basic/Sentry.Samples.Console.Basic.csproj` to get an AOT complied app I see that we _sometimes_ get line numbers and _sometimes_ don't. 

For [this exception](https://github.com/getsentry/sentry-dotnet/blob/48b744fd3e3996c1c425aae8a8607c69d2dfb62d/samples/Sentry.Samples.Console.Basic/Program.cs#L159-L160) I [see](https://sentry-sdks.sentry.io/issues/7331516266/events/931cd69fde0d411a85a532e6013376c6/) full symbolication of the system frames but not for the InApp frames:
```
System.InvalidOperationException: Something happened that crashed the app!
  at Program.<>c__DisplayClass0_0.<<<Main>$>g__ThirdFunction|4>d.MoveNext()()
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()() in ExceptionDispatchInfo.cs:line 53
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)() in TaskAwaiter.cs:line 154
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task, ConfigureAwaitOptions)() in TaskAwaiter.cs:line 118
  at Program.<<Main>$>d__0.MoveNext()()
  at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()() in ExceptionDispatchInfo.cs:line 53
  at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task)() in TaskAwaiter.cs:line 154
  at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task, ConfigureAwaitOptions)() in TaskAwaiter.cs:line 118
  at Program.<Main>(String[] args)()
  at ?()
```

That will probably be an issue with not resolving the symbols I uploaded for that app... none the less, even the symbolication of the system frames demonstrates that we can symbolicate (file names and line numbers) AOT compiled code.

supervacuus: Thanks for the input from both of you. Just to clarify: this proposal is only relevant for scenarios where full symbol coverage via a backend store is not an option (the classic example being Android, because there are more platform OEMs than we can realistically collect from). So, if you know that the uploaded debug info can fully recover every frame, there is no need for this feature.

The .NET-related [issue](https://github.com/getsentry/sentry/issues/97054) I linked in the RFC just highlights a problem that aligns with what we are trying to solve here: 

* The UI signals to the user that something is missing, and while that is technically true, it is not necessarily actionable. 
* In that case, the client was able to symbolicate frames from `ConsoleOnboarding.dll`, but the signal to the user is still that something is not right and needs to be fixed. 
* With a tag from the SDK indicating that symbolication happened on the client for a particular frame, the UI could check that all frames associated with that module have already been correctly symbolicated and, as such, lower the severity of the warning and call to action. 
* Of course, in this particular case, there might still be a realistic chance that this is an okay warning (but the issue was created in response, so I guess even here the error is overstating the fact and confusing to users)
* In scenarios where the referenced module will _never_ be uploaded, the warning is just factually wrong (and in the case of native, it would not only affect the module being marked as missing, but each associated frame would be marked, and the event would also get an error at the top, as shown in the screenshots)

supervacuus:
Thanks to all reviewers for the feedback.

Since I haven't seen any immediate pushback, I went ahead and transformed the previous document to better center on the approach using the `symbolication` object and to push the role of the arbiter toward `symbolicator`. 

Specifically, the RFC proposes that symbolicator honors the `symbolication` annotation by not overriding client-provided symbols, while continuing to apply demangling, source context, and line-number enrichment (see [Control semantics](https://github.com/supervacuus/rfcs/blob/rfc/sdk-symbolicated-frames/text/0152-sdk-symbolicated-frames.md#control-semantics)).

Is this the right granularity, or are there stages that should also be skipped or handled differently for client-symbolicated frames?

While I invite everyone to reread (if you have time), these changes are particularly addressing the ingestion team (@Dav1dde, @jjbayer, and now also @loewenheim). I really hope I managed to synthesize the aspects coherently enough for us to push this over the line soon.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,299 @@
...
+- **`by`**: Identifies who performed the symbolication. An enum of `"client"` (SDK) or `"symbolicator"` (backend).
+- **`status`**: The outcome. An enum of `"success"`, `"failed"`, `"missing"`, or `"unknown"`.
+
+This minimal shape already provides the control mechanism needed to solve the motivating problems, while the object structure allows future extension without breaking changes.
+
+## Control semantics
+
+Processing passes the `symbolication` object through to symbolicator as part of each frame. Symbolicator interprets the annotation and decides how to handle each stage.
+
+When symbolicator encounters a frame with `symbolication.by == "client"` and `symbolication.status == "success"`:
```
jjbayer: What should happen when symbolicator encounters a frame with `symbolication.by == "symbolicator"`. This "should not happen" but I think it makes sense to define the behavior.

text/0152-sdk-symbolicated-frames.md
```diff
@@ -0,0 +1,299 @@
...
+
+- **Symbol resolution** is skipped. The client-provided `function`, `symbol`, `filename`, `lineno`, `colno`, and `abs_path`, if provided, are not overwritten. The frame is not reported as `"missing"` if the corresponding debug file is unavailable.
+- **Demangling** is still applied. If the SDK sends a mangled C++ symbol, symbolicator can demangle it.
```
jjbayer: Do we want the client to be able to mark symbols as mangled? Something like

```json
"symbolication": {
    "by": "client",
    "status": "success",
    "demangled": false
}
```

This could be a later addition to the protocol.

jjbayer (approved):
LGTM, I'll let @loewenheim comment on the symbolication details (demangling, source context, etc.).

text/0152-sdk-symbolicated-frames.md
supervacuus: We can add flags for all other `symbolicator` stages in future extensions.

text/0152-sdk-symbolicated-frames.md
supervacuus: My position is that anything but `symbolication.by == "client" && symbolication.status == "success"` follows current behavior. But we could also only look for `symbolication.status == "success"` and ignore who did it.

loewenheim:
I think the proposal as it currently stands is very good. I just noticed a small ambiguity: if a frame is pre-symbolicated, but Symbolicator makes an addition (e.g. source context), should that be recorded somehow? Should the value of the `"by"` field change from `"client"` to `"symbolicator"`, or should it be left alone?

supervacuus:
> I think the proposal as it currently stands is very good. I just noticed a small ambiguity: if a frame is pre-symbolicated, but Symbolicator makes an addition (e.g. source context), should that be recorded somehow? Should the value of the `"by"` field change from `"client"` to `"symbolicator"`, or should it be left alone?

This essentially touches on the same asymmetry @jjbayer raises [here](https://github.com/getsentry/rfcs/pull/152#discussion_r2993025622):

> Do we want the client to be able to mark symbols as mangled? Something like
> 
> ```json
> "symbolication": {
>     "by": "client",
>     "status": "success",
>     "demangled": false
> }
> ```

The RFC proposes a top-level object `symbolication`, but we actually touch on multiple stages of per-frame enrichment:

* symbols
* demangling
* source-context
* line-numbers

All could have some level of client-side resolution. Symbols are the most urgent right now, but each one accepts at least a "by" and "status" field, conceptually, even if "symbols" are currently the only ones coming from the clients, and the others are only filled in for provenance by `symbolicator`.

So we could have something like this coming in from the SDK

```json
"symbolication": {
    "symbols": {
        "by": "client",
        "status": "success",
    }
}
```

which implies that all other enrichments have not been attempted on the client, and then after symbolicator processed it, the frame would get this (bikeshedding ignored):

```json
"symbolication": {
    "symbols": {
        "by": "client",
        "status": "success",
    },
    "demangling": {
        "by": "symbolicator",
        "status": "success",
    },
    "source_context": {
        "by": "symbolicator",
        "status": "missing",
    },
    "lineno": {
        "by": "symbolicator",
        "status": "success",
    }
}
```

Each of those frame enrichments could then receive common extensions such as `previous_attempts` or enrichment-specific fields. I mean, it does have the downside of further nesting, but aside from that, I think something like this is the solution for both items. What do you say, @jjbayer and @loewenheim, does it make sense?

jjbayer:
> I mean, it does have the downside of further nesting, but aside from that, I think something like this is the solution for both items

@supervacuus I agree, this would make the protocol even more powerful, and inform symbolicator exactly about what it should do.

What would be the condition for sentry to send the event to symbolicator? I assume the number of sub-entries in `"symbolication"` depend on the platform, so I guess sentry would determine on a platform-specific bases which "success" entries it needs to see to skip server-side symbolication entirely?

supervacuus:
> What would be the condition for sentry to send the event to symbolicator? I assume the number of sub-entries in `"symbolication"` depend on the platform, so I guess sentry would determine on a platform-specific bases which "success" entries it needs to see to skip server-side symbolication entirely?

The RFC is written from the POV of frames where `platform == "native"`. I am aware that other platforms also use symbolication infra, so I would likely consider `symbolication` cross-platform (not just `"native"`), but I can't make that call for other platforms, and the feedback suggests limited non-default usage even inside `"native"`.

We can add checks in the platform-specific processing modules in `sentry` for this as `SHOULD` in the RFC, if that is what you are interested in, to reduce unnecessary load on `symbolicator` instances. For instance, the Android SDK detects whether native frames come from modules inside the app package or not (mostly for `inApp` marking), but for native system frames, the SDK could set all these frames as client sucesses (since there is no realistic way to further enrich them in the backend), and "processing" no longer would need to search for markers to create heuristics of what to exclude from the frame-set being sent to symbolicator or not. This functionality could be shared across platforms because the success filter is the same; the only thing that changes between platforms is the set of filter fields.

Is that what you meant?  Does it sound sensible?

supervacuus:
I also wonder if it would make sense to have 

```JSON
"symbolication": {
    "by": "client",
    "status": "success"
}
```

mean

```JSON
"symbolication": {
    "symbols": {
        "by": "client",
        "status": "success",
    },
    "demangling": {
        "by": "client",
        "status": "success",
    },
    "source_context": {
        "by": "client",
        "status": "success",
    },
    "lineno": {
        "by": "client",
        "status": "success",
    }
}
```

for whatever set a specific platform uses, since, for instance, in the Android case, the native system frames will likely represent the largest frame population overall thread stack traces in any event.

I totally understand if you downvote this, but it would significantly reduce the payload.

jjbayer:
> We can add checks in the platform-specific processing modules in `sentry` for this as `SHOULD` in the RFC, if that is what you are interested in, to reduce unnecessary load on `symbolicator` instances. For instance, the Android SDK detects whether native frames come from modules inside the app package or not (mostly for `inApp` marking), but for native system frames, the SDK could set all these frames as client sucesses (since there is no realistic way to further enrich them in the backend), and "processing" no longer would need to search for markers to create heuristics of what to exclude from the frame-set being sent to symbolicator or not. This functionality could be shared across platforms because the success filter is the same; the only thing that changes between platforms is the set of filter fields.
> 
> Is that what you meant? Does it sound sensible?

Yes this sounds sensible. It does not need to be part of the RFC to be honest, let's just see it as an optimization that the sentry processing tasks may do.

> I totally understand if you downvote this, but it would significantly reduce the payload.

The shorthand makes sense to me 👍

supervacuus:
I added the enrichment stages to the RFC.

supervacuus:
@jjbayer and @loewenheim, can I ask you to quickly fly over [the diff](https://github.com/getsentry/rfcs/pull/152/changes/69873ccbd43ecad037e41b8d49d09731cc69f769) and check whether it captures our exchange? Likely, the changes to the introduction are enough, since the other changes only detail that.

In a way, it is a considerable design change to expose the stages to the client; thus, it would help to validate my understanding of the conversation we had.

loewenheim:
The only thing that stands out to me is this.
>When symbolicator encounters
>
>```
>"symbolication": {
>  "symbols": {
>    "by": "client",
>    "status": "success"
>  }
>}
>```
>it treats symbol resolution for that frame as already authoritative on the client side:
>
> - Symbol resolution is skipped. The client-provided function and symbol, if present, are not overwritten.
> - The frame is not reported as "missing" if the corresponding debug file is unavailable.
> - The associated module/image should not be attributed as missing solely because server-side symbol resolution was skipped.
> - Other enrichments remain open by default: demangling, source context, and location continue to follow current behavior unless and until the corresponding stage is also claimed in the normalized long form.

Specifically the second bullet point ('The frame is not reported as "missing" if the corresponding debug file is unavailable.'): I assume the intention is that if the frame's function name is filled in, but we can't resolve line numbers because of a missing debug file, the frame would be considered missing as it does now?

supervacuus:
> Specifically the second bullet point ('The frame is not reported as "missing" if the corresponding debug file is unavailable.'): I assume the intention is that if the frame's function name is filled in, but we can't resolve line numbers because of a missing debug file, the frame would be considered missing as it does now?

Yeah, that is a bug in the spec: since this symbolication spec from the client is `undefined` for all stages except `symbols`, `symbolicator` should attempt those, and, as a result, if the lookup fails, the frame and its module should be highlighted as missing. Only if all stages have been marked as `"success"` by the client, should there be no further lookup and no missing marker in the UI.

I am going to fix this; that was likely a copy/paste error.

loewenheim:
Thank you for clarifying. Everything else looks good to me!

PR #156, opened 2026-03-19
rfc(decision): scope model RFC for Go

giortzisg:
This RFC evaluates how `sentry-go` should model scope state so it can align with Sentry’s three-scope model 

[Rendered RFC](https://github.com/getsentry/rfcs/blob/7880acc7dee3030c52f6f7d48d56291345c26d34/text/0156-scope-model-for-go.md)

### Links
- resolves GO-127
- links to https://github.com/getsentry/sentry-go/issues/1228


text/0156-scope-model-for-go.md
```diff
@@ -0,0 +1,296 @@
...
+- Easier to reason and user friendly.
+- Removes scope level locking. Race conditions are impossible, simplifies SDK maintainance.
+- Users don’t need to manage hidden mutable shared state.
+- In general `context.Context` is meant to be immutable, so this option makes the most sense in the Go ecosystem. We simplify the `Scope` API for users.
+- This aligns much more naturally with OTel’s immutable `context` model.
+- Goroutine propagation becomes safe by default as long as the caller passes `ctx` and each goroutine receives an immutable scope snapshot instead of a shared mutable scope pointer.
+
+### Cons
+
+- Major change from current SDK architecture, both for us and the users.
```
sl0thentr0py, line 209: we need concrete examples here on what migration would mean for the users and how much code they need to change

sl0thentr0py (changes requested):
While I mostly understand the reasoning behind this RFC, and also partially agree with the arguments set forth, some considerations and comments:

### Why 3 Scopes?

* the whole point of the other 3 scope RFC was to make our scope more in line with otel copy-on-write context (which is the same design as go context) for POTEL (= bidirectional flow between sentry <-> otel)
* if we already make our scope context like in this RFC suggests, we don't even need 3 scopes - just a global and a context-like scope suffices
* the `currentScope` especially is really weird here because all it needs to hold is a `span` reference but we make it hold all the other stuff like tags and data and if we already have a copy-on-write structure we don't need the added complexity
* honestly, since POTEL is now a dead project and we are already considering reverting it in JS, I'm not sure this whole 3-scope thing is even necessary. I'm definitely postpoining it and might not do it at all in Ruby, for instance.

### Breaking Changes and Migration

We need a proper pathway for migration if we do this, so please add more clarification on this in the RFC.

### Divergence

If we go forward with this in go, it will certainly create short term (if not permanent) divergence between go and other SDKs, something we have to be aware of and have buy-in from relevant people.

giortzisg:
The three scope model mapping section is mostly there for associating how the proposal would work when compared to the old approach.

I also agree that we don't really need the three scopes model (or at least limit it to only two scopes) for it to actually work, so let me rework the format a bit to be more clear.

cleptric:
As discussed offline, I'm in general happy with this approach, but we should include how a new span API interacts with the new scope model.

text/0156-scope-model-for-go.md
```diff
@@ -350,6 +350,59 @@ func middleware(next http.Handler) http.Handler {
...
+- the active span is stored in `context.Context`.
+- the active span is also mirrored onto the current scope.
+
+This means the active span is not represented by a single source of truth. Child spans are created from the active span in `ctx`, but tracing logic also mutates scope state to attach the active span and restore parent spans when spans finish.
+
+#### Target direction
+
+The final tracing API should move toward a span-first model similar to OTel:
+
+- `context.Context` is the canonical carrier of the active span.
```
lcian, line 376: Is it correct to say that `Context` just holds a reference to the active span (could be just the span ID I think) (plus the DSC)?
Because `Context` is immutable but, judging by the examples below, you want `Span` to be mutable, so I don't think you can just store the whole `Span` on the `Context`.

Or maybe you can store the whole `Span` in `Context` and always return a `&Span` to the user. But that way you will have interior mutability.

text/0156-scope-model-for-go.md
```diff
@@ -350,6 +350,59 @@ func middleware(next http.Handler) http.Handler {
...
+```go
+ctx, span := sentry.StartSpan(ctx, "validate-cart",
+ 	sentry.WithAttributes(
+		attribute.String("user.id", "123"),
+ 	),
+)
+defer span.End()
+
+span.SetAttribute("valid", attribute.BoolValue(true))
+span.SetStatus(sentry.SpanStatusOK)
```
lcian, line 394: `Span` is mutable, so it will still need to lock internally for methods like `SetAttribute`, right?

text/0156-scope-model-for-go.md
```diff
@@ -350,6 +350,59 @@ func middleware(next http.Handler) http.Handler {
...
+```
+
+The main behavior should be:
+
+- StartSpan(ctx, name, ...) returns (context.Context, Span).
+- the returned ctx carries the started span as the active span.
+- if no explicit parent option is passed, the started span should use the currently active span from ctx as parent.
+- if an explicit WithParent(span) option is passed, that span becomes the parent.
+- if an explicit WithNoParent() option is passed, the new span becomes a root/segment span.
+
```
lcian, line 406: I think we also need something like `sentry.GetSpan(context.Context)` if the user wants to get the current span and modify it.

text/0156-scope-model-for-go.md
giortzisg: With just the span context we wouldn't be able to have any mutation methods on the span itself (like span.End and span.SetAttributes). 

We need to expose a `GetSpanFromContext` method so that users and integrations can get the span and add span attributes and finish. 

> Or maybe you can store the whole Span in Context and always return a &Span to the user. But that way you will have interior mutability.

That's what we will do, since we can't really get away with also span immutability. Let me add a segment for that to make it more clear.

text/0156-scope-model-for-go.md
giortzisg: we already expose a `SpanFromContext` method, so this will stay as is in the API.

text/0156-scope-model-for-go.md
```diff
@@ -211,6 +211,122 @@ This fits existing Go APIs well. `otel`, `grpc/metadata`, and similar packages a
...
+    - scope.SetAttributes(...)
+- Then replace the mutation site with the ctx-returning equivalent:
+    - ctx = sentry.SetTag(ctx, ...)
+    - ctx = sentry.SetUser(ctx, ...)
+    - ctx = sentry.SetAttributes(ctx, ...)
+
+#### Hub-in-context APIs
+
+- SetHubOnContext(ctx, hub) and GetHubFromContext(ctx) stop being the primary user-facing propagation mechanism.
+- Users should no longer manually clone hubs for request/task isolation, just pass the correct context.Context.
```
sl0thentr0py, line 288: this assumes that that go context forking semantics for concurrency/isolation are the same as what we were doing before. Does this hold for all cases where we manually forked before?

text/0156-scope-model-for-go.md
```diff
@@ -350,6 +350,59 @@ func middleware(next http.Handler) http.Handler {
...
+- StartSpan(ctx, name, ...) returns (context.Context, Span).
+- the returned ctx carries the started span as the active span.
+- if no explicit parent option is passed, the started span should use the currently active span from ctx as parent.
+- if an explicit WithParent(span) option is passed, that span becomes the parent.
+- if an explicit WithNoParent() option is passed, the new span becomes a root/segment span.
+
+#### Interaction with scope model
+
+Under the proposed scope model, scope if only an infromation carrier and is no longer responsible for owning tracing state. The `context.Context` carries both the scope and active span state.
+This means that `StartSpan` is also a scope-boundary operation and derived scopes already contain the previously inherited local state. Under this, the SDK does not need to merge separate scopes when a span ends. Semanticaly, the model gets simplified to:
```
sl0thentr0py: 💯

sl0thentr0py (approved):

dingsdax (approved):

text/0156-scope-model-for-go.md
giortzisg: That's a nice observation. I would say yes, since each scope.fork snapshot is essentially a different context. For forks that prevented mutability CoW guarantees that and for forks that are a snapshot of scope state, the mechanism now would be the ctx.

The only thing I can think of is `CurrentHub.Clone()` which becomes completely redundant under our model, since ctx and global scope would merge anyways on send.

PR #158, opened 2026-05-01
rfc(decision): Migrate Browser Web Vitals to Metrics

logaretm:
Web vitals are measurements, not execution traces. Metrics fit them better as a structure and future proofs them for soft navigation web vitals that will roll out in browsers.

The current span model carries structural overhead that doesn't serve the data. The exact overhead depends on the SDK code path: current default (v1) only sends INP as a standalone web vital span, while streamed spans (v11 default) also send LCP and CLS as standalone spans. The pageload span remains in both cases. Trace metrics cost less per emission and capture 100% of data (no SDK sampling).

The migration is phased in a way that hinges on two critical aspects: 

- **_No billing change for customers at any point_**
- **_We only end up maintaining one data source for web vitals_**

**Suggested plan is:**

1. **Double-write**: Relay converts web vital spans to metrics; both are written, derived metrics are free
2. **Dashboard cutover**: dashboards switch to metrics-only queries after sufficient metrics history
3. **Single-write:** Relay stops writing the original span, converts old-SDK spans to metrics billed as spans
4. **SDK v11**: SDK emits web vitals as metrics natively, billed as metrics
  
[Rendered RFC](https://github.com/getsentry/rfcs/blob/awad/webvitals-as-metrics/text/0154-web-vitals-as-trace-metrics.md)


text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+- RFC PR: TBD
+- RFC Status: draft
+- RFC Driver: [Abdelrahman Awad](https://github.com/logaretm), [Dominik Buszowiecki](https://github.com/DominikB2014)
+- RFC Approver: TBD
+
+# Summary
+
+Migrate web vitals (LCP, CLS, INP, FCP, TTFB) from spans to trace metrics in the Sentry JavaScript SDK. It's a multi-phase approach to ensure we end up with single data source for web vitals and avoid introducing any billing changes for our customers.
+
+Relay double-writes web vital spans as trace metrics (free) for a period of time (detailed below), then dashboards cut over to metrics-only queries. New major SDK releases emit web vitals as trace metrics natively, billed as metrics. Finally, Relay stops double writing and only converts spans from older SDKs to metrics, billed as spans, no customer sees a billing change at any phase.
```
Dav1dde: I don't think this is a good idea, we should first figure out what it takes to make that billing change and then make a decision. Maintaining this double billing pipeline is not something what I would commit to easily, this is an absolute first in Sentry with a wide range of implications (rate limits, protocol, ...).

Dav1dde: Why double write in Relay instead of figure it out on the query side, how was that decision made?

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Suggested timeline:**
+
+The timeline here isn't sequential, as the work items can happen in parallel. But the over all goals to be met are dependent on the double-writing period duration.
+
+Having said that, these are the work items with their own expected timelines:
+
+- Double-write: 6 months, should be enough to build sufficient metrics history before dashboard cutover.
```
Dav1dde, line 20: Retention period is 13 months, 6 months seems quite a long time frame to not then solve the 13 months down-sampled retention. If we already compromise on the 13 months retention, can we do it faster than 6 months?

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+- Dashboard and product cutover: Given enough metrics has been collected from the double-write period, we can cut over to metrics-only queries.
```
Dav1dde, line 21: As noted above, if the query side supports both storages, we don't need double writing or cutovers.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+
+# Motivation
+
+Web vitals are measurements, not execution traces. The current span-based implementation carries structural overhead that doesn't serve the data:
+
+- **Zero-duration spans.** LCP, CLS, and INP spans have zero durations. The span model's timing, parent/child relationships, and segment tracking add ~300-500 bytes per emission for fields that carry no information.
+- **Cost to customers.** At confirmed post-GA metrics pricing ($0.50/GB, log parity), web vitals as metrics cost ~85-95% less per emission than as spans. At realistic sample rates, metrics-unsampled costs roughly the same as spans-sampled but captures 10x more data.
+- **Metrics use-cases.** The browser does not have any OOTB metrics use-cases. By moving web vitals to metrics, it encourages metric adoption by customers.
+- **Standalone Span Deprecation.** The v2 span protocol is in the works and will deprecate standalone spans. But this is also an opportunity to change protocols and align with better telemetry model. We only have one type of spans to match against here, but if span-first ships first then we have two sets of spans to match against.
```
Dav1dde: Just a small nit, all of span v2 is "standalone". What is being deprecated is the transaction+span schema.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+- Double-write: 6 months, should be enough to build sufficient metrics history before dashboard cutover.
+- Dashboard and product cutover: Given enough metrics has been collected from the double-write period, we can cut over to metrics-only queries.
+- Single-write conversion: After the double-write period, we can stop double-writing and only convert spans from older SDKs to metrics.
+- JS SDK emits web vitals as metrics natively: It can happen in the interim of the double-write, and it would reduce the cost of the double-write period.
+
+# Motivation
+
+Web vitals are measurements, not execution traces. The current span-based implementation carries structural overhead that doesn't serve the data:
+
+- **Zero-duration spans.** LCP, CLS, and INP spans have zero durations. The span model's timing, parent/child relationships, and segment tracking add ~300-500 bytes per emission for fields that carry no information.
```
Dav1dde: That's a large estimate, what fields would that be? I think it boils down to only a few fields which are required on a span but not a trace metric (`timestamp` and `span_id`?), all other information added to a span is optional. This should come out to a few bytes not `~300-500`.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+| SDK major | % of volume |
+|---|---|
+| v10 (current) | 46% |
+| v9 | 19% |
+| v8 | 19% |
+| v7 | 10% |
+| Unknown | ~5% |
+| v6 and older | ~1% |
+
+**Legacy tail implication:** 48% of web vital span volume comes from pre-v10 SDKs. v7 at 10% is non-trivial and likely sticky (pinned SDKs, older customers). Relay's span-to-metric conversion is not a short-term bridge and needs to be robust and long-lived. That said, this is not net-new maintenance. Relay already maintains span-matching logic for performance score computation (deriving `score.lcp`, `score.cls`, etc. from span attributes). The metric converter is a similar pattern on a different telemetry type, not additional complexity on top of what already exists.
```
Dav1dde: Who made that assessment? This conversion would be nothing like performance score calculation.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,324 @@
...
+When spans stop being written, anything querying the spans dataset for web vitals will silently stop returning data:
+
+A non-trivial number of active alerts, saved Discover queries, and custom dashboard widgets reference web vital span measurements across a significant number of orgs. These won't error, they'll just quietly stop producing data. Customers may not notice immediately. We have a couple of options here:
+
+- **Migrate alerts and Discover queries to the metrics dataset.** Auto-migrate known patterns (e.g. `measurements.lcp` → `browser.lcp`) or provide migration tooling.
+- **Gate the single-write conversion on a feature flag.** Keep double-write running longer to give customers time to migrate their alerts and Discover queries to the metrics dataset before spans are dropped.
+
+The longer the double-write period is, the more time we have to migrate the alerts and Discover queries to the metrics dataset.
+
+### Provenance-based billing
```
Dav1dde, line 276: How does this work for AM2 where we don't bill by spans but transactions/performance units?

Dav1dde (commented):
This document lays out a plan without discussing alternatives and downsides, I think this would be better suited using the DACI process to collect different approaches as well as up/down sides and considering the noop option.

logaretm:
> This document lays out a plan without discussing alternatives and downsides, I think this would be better suited using the DACI process to collect different approaches as well as up/down sides and considering the noop option.

My bad, I had assumed the thread already navigated the noop and the query adjustment options given we didn't want to have two data sources co-existing. I drafted this within those parameters. I will add them.

text/0154-web-vitals-as-trace-metrics.md
logaretm: Maybe I misunderstood, I believe this was discussed in the slack thread already. My understanding is that we would only consider this idea if it meant we don't maintain two data sources and eventually spans are phased out.

If query/dashboard side adjustments are an option, then the only change needed is from the SDK side and no fancy double-writing is needed. Double writing was only suggested to help migrate the data source to metrics.

text/0154-web-vitals-as-trace-metrics.md
logaretm: We considered 14 days first then 30 days. There was concerns from product that we needed more historical data given that the absorbed costs are justifiable.

I suppose we can do a shorter period but will need to find out how to fill the gaps or how to transition the customer properly.

text/0154-web-vitals-as-trace-metrics.md
logaretm: @DominikB2014 Would be in a better position to answer this. I think it isn't possible which is why double/single writes and dashboard fallbacks were being suggested in the first place.

But worth-rexploring as it would simplify things considerably.

text/0154-web-vitals-as-trace-metrics.md
logaretm: Sorry, that is correct. It's just how we refer to them in the SDK:

- Standalone: Those web vital spans
- v1: Transaction + Span impl
- v2: Streamed spans / span first

Will clarify that up.

text/0154-web-vitals-as-trace-metrics.md
logaretm: It's an estimation at best, I built payloads for each span and averaged them against their metric counterparts.

**Fixed overhead (195):**

- `start_timestamp` (duplicated, equals `timestamp` for zero-duration spans) 32B
- `exclusive_time` at top level 18B
- `sentry.exclusive_time` in `data` (duplicates top-level) 25B
- `parent_span_id` 35B
- `segment_id` 31B
- `op` at top level (duplicates `sentry.op` in `data`) 22B
- `origin` at top level (duplicates `sentry.origin` in `data`) 32B
   
**Variable overhead per vital:**

- `description` (element name, not needed on metric) ~47-70B
- `events[]` array (v1 only, wraps `sentry.measurement_value` + `sentry.measurement_unit` in a
  nested object) ~105-114B
- `measurements` mirror (v1, duplicates the value already in `events[]`) ~47-60B
- `sentry.report_event` (LCP/CLS only) 32B  
  
**Total overhead per vital:**                                                               

- LCP: ~471B                                                          
- CLS: ~431B
- INP: ~410B

Some vitals are heavier than others, like CLS

text/0154-web-vitals-as-trace-metrics.md
Dav1dde: I think there are 2 options which essentially lead to the same outcomes:

1. Double write for X months, after X months only query metrics data set
2. Query metrics and spans data sets for X months and convert all data to metrics, after X months only query metrics data set

It seems like 2) would be easier to do, but that's best confirmed by someone from the databrowsing team.

text/0154-web-vitals-as-trace-metrics.md
Dav1dde: Thanks for sharing the numbers, looks like I missed some attributes. 

I think if the goal was to optimize size, we would have some wiggle room to make the spans a bit slimmer.

Dav1dde:
@logaretm For me interesting questions (possibly options) to explore would be:

- What if we do nothing?
- Double write on ingestion vs double read.
- What options do we have for billing? 
  - Do we really need to count metrics as spans (really not a good idea with how Relay associates item types with data categories today)? 
  - How does it work for AM2 (billing based on transactions)? 
  - Can we just switch all billing over to metrics, with all customers also having 5 GIB free metrics quota now? 

I think the billing questions are the important ones, depending on how we can make this happen on the billing side, we can guide the implementation. Counting metrics as spans, I would really only consider if it is for a limited time-frame and we'll have to accept a bunch of caveats during that time.

text/0154-web-vitals-as-trace-metrics.md
logaretm: Definitely not the main goal at all here. Mostly a "pro" to mention is all.

text/0154-web-vitals-as-trace-metrics.md
logaretm: That's fair, and I guess it is a bit of a stretch. The intent was to point out that we have a precedent for deriving data out of web vital spans (so we must have already some logic of matching/scoring them). Converting to a metric involves similar logic (which spans, what's the metric value).

This is a bad assumption from my part, and it would certainly be different in terms of complexity. I will remove that completely as we should focus on the proposal's own merits instead.

text/0154-web-vitals-as-trace-metrics.md
DominikB2014: Even if it's possible to create a query to join the metrics and spans data, this feels like it would become a strange product experience. If I'm a user who creates a dashboard widget, selected spans as the dataset, and then aggregated a web vitals span, it just query metrics and spans under the hood? 

Imo data that you query for should match directly what's stored, even if it's more painful to migrate it feels like the right solution.

Dav1dde: Fair enough, figured our main problem may be the "insights module"

k-fish:
@Dav1dde I was talking with Awad a bit offline, we discussed some of your points

> What if we do nothing?

imo the major upside to doing this is that today it IS quite annoying to collect metrics on your frontend for two reasons:
1. Aggressive `traceSampleRate` really reduces the number of vitals you collect, especially since pageload spans aren't that prevalent in the first place
2. Vitals sent outside of `idleTimeout` were previously impossible, are were only solved with 'standalone' spans for INP etc. but really vitals can be sent at any time, and sometimes multiple times per pageload.

So in the long term having these supported on metrics with (possibly) their own separate higher sample rate is quite beneficial for users, without tying things to tracing.

> Do we really need to count metrics as spans (really not a good idea with how Relay associates item types with data categories today)?

Existing spans that currently have vitals attached would still be sent **even after double-write or any other scenario**, since vitals are in measurements on the root span today, that doesn't change after this since the root span is still sent afterwards. W.r.t. billing, if we need to make a new outcome for metrics and have complex billing, it should be a new datacategory `SPAN_RELATED_METRIC` etc. that we can simply not show (if we aren't charging it), or also bill on (if we are). 

Removing double-write later I don't think actually reduces the size on the pageload span a huge amount since attributes are quite compressed in storage (tens of x).

> How does it work for AM2 (billing based on transactions)?

Billing for spans/transactions would remain completely unchanged since even after double write we wouldn't remove any spans (maybe with the exception of standalone INP, which is pretty low #).

> Do we really need to count metrics as spans (really not a good idea with how Relay associates item types with data categories today)?

For the double write period we definitely shouldn't charge for both since that would be overcharging the user, once we move to single write we can just charge for metrics. 


**Billing**

Essentially, since our goal is to not charge users anything extra, we'd eat the cost of metrics double writing while we're working on switching over. If the users manually enabled the metrics integration early, they'd be charged for it ofc since the sample rate is 100% not the traceSampleRate (they get way more vitals data points). Then when a new major version of js sdk is released we can default the metrics vitals being sent to `true` and those should be charged to the user (since they will be a higher % above the previously sent free ones on spans), so there will be some additional cost on the new version on the sdks, but ONLY when the user passes the free metrics SKU (5gb) which is millions of metrics events.

 Thoughts?

Dav1dde:
> Existing spans that currently have vitals attached would still be sent even after double-write or any other scenario, since vitals are in measurements on the root span today, that doesn't change after this since the root span is still sent afterwards.

Is my assumption wrong that we can convert these spans entirely to metrics and from that point on the span 'never existed'?

> , if we need to make a new outcome for metrics and have complex billing, it should be a new datacategory SPAN_RELATED_METRIC etc. that we can simply not show (if we aren't charging it), or also bill on (if we are).

This just doesn't work well, because of unclear relationships between rate limits, and envelope items, and data categories. This is not a path we should go down, we explored it with continuous profiling and it's not great.

> Billing for spans/transactions would remain completely unchanged since even after double write we wouldn't remove any spans (maybe with the exception of standalone INP, which is pretty low #).

I see, I think I missed the part where user sending trace metrics are billed as such immediately.

> (maybe with the exception of standalone INP, which is pretty low #)

Curious, why keep them around?

> Billing [...]

That makes sense, for me important is really just the end state and we don't have to carry some legacy form of billing for webvitals spans forever.
If the end state is, all web vital spans are converted to metrics and billed as metrics, that sounds like a good place to be.

logaretm:
> Is my assumption wrong that we can convert these spans entirely to metrics and from that point on the span 'never existed'?

So the conversion isn't uniform, because of how messy it is today. Depending on user configuration, all 5 vitals can be represented with 2 spans or 4 spans (if they turn on `lcp, cls` span reporting, otherwise they ride on pageload spans).

| Vital | Current source                                           | Current representation          |                                                                                                                                                                                                             
  |-------|----------------------------------------------------------|---------------------------------|                                                                                                                                                                                                             
  | LCP   | Standalone span (`ui.webvital.lcp`)                      | 1 span (opt-in, experimental, otherwise rides on pageload)   |                                                                                                                                                                                                             
  | CLS   | Standalone span (`ui.webvital.cls`)                      | 1 span (opt-in, experimental, otherwise rides on pageload)   |                                                                                                                                                                                                             
  | INP   | Standalone span (`ui.interaction.{click,keypress,drag}`) | 1 span (always)                 |                                                                                                                                                                                                             
  | FCP   | `measurements.fcp` on pageload span                      | Bundled on pageload span (free) |                                                                                                                                                                                                             
  | TTFB  | `measurements.ttfb` on pageload span                     | Bundled on pageload span (free) |       

So if we were to just convert, we would eliminate `INP` spans only, and `LCP/CLS` if the user was sending them. The `pageload` span would still exist as it is, we just added the measurements to it. INP is the exception because it comes in at any point and usually after `pageload` has eneded.

tldr; singel conversion would remove 1 span, or 3 at best to emit 5 metrics. Still cheaper for the user tho but not factoring in sampling.

> Curious, why keep them around?

INP is one of the "core" metrics today so it has very high visibility right now, the problem with it is it gets reported when the user tabs away. So it comes in way later, sometimes sampled away, and sometimes dropped because there is no longer any active root spans. This would get fixed by switching to metrics, but in this RFC case, only when user switch to the new SDK.

Dav1dde:
If we were to do the conversion (final state) in Relay:

- LCP, CLS, INP span converted to Metric
- FCP, TTFB, a new metric is extracted for each measurement on the span

For the double writing case, the span (and measurements) would always stick around + the metrics are extracted.

Does that sound about right?

logaretm:
> If we were to do the conversion (final state) in Relay [...]

I would adjust it slightly since `CLS` and `LCP` can either exist as a span or as a measurement.

So:

- `LCP`, `CLS`, `INP` spans converted to Metric
- `FCP`, `TTFB`, `LCP`, `CLS` a new metric is extracted for each measurement on the span

The good news is `CLS` and `LCP` won't be duplicated because it exists as either a span, OR a measurement but never both.

https://github.com/getsentry/sentry-javascript/blob/748d8cc02483e853baa259c56a5bda1c5d686556/packages/browser-utils/src/metrics/browserMetrics.ts#L113-L123

text/0154-web-vitals-as-trace-metrics.md
k-fish: I think we need to also consider that old sdk usage means they'll continue to only send vitals on spans for a very long time probably 1 year +

text/0154-web-vitals-as-trace-metrics.md
k-fish: I would also adjust the bytes emission point here downwards, 300-500 is too high an estimate.

`parent_span_id` would be included on the metric as well, and possibly `origin` now that it's a metric generated via an integration, no?

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,375 @@
...
+
+
+| Mode (100M pageloads/month) | Items | Monthly cost (PAYG) | Coverage |
+|---|---|---|---|
+| Spans, `tracesSampleRate: 1.0` | 300M standalone spans | ~$560 | 100% |
+| Spans, `tracesSampleRate: 0.1` | 30M standalone spans | ~$60 | 10% |
+| Metrics (unsampled), ~400B/item | 500M metrics (300M LCP/CLS/INP + 200M FCP/TTFB) | ~$100 | 100% |
+| Metrics (unsampled), ~200B/item | 500M metrics | ~$50 | 100% |
+
+FCP and TTFB currently ride for free on the pageload transaction span (~3 standalone vital spans + 2 bundled measurements per pageload). As metrics, all 5 vitals become separate items. Even so, at metric pricing the 500M items still cost less than 30M sampled spans while capturing 100% of emissions.
```
k-fish: Doesn't LCP also get sent on the pageload?

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,375 @@
...
+    "sentry.pageload.span_id": "...",
+    "sentry.report_event": "pagehide",
+  },
+  events: [{
+    name: "lcp",
+    attributes: {
+      "sentry.measurement_value": 1234.5,   // the actual LCP value
+      "sentry.measurement_unit": "millisecond",
+    }
+  }]
```
k-fish, line 68: Again I'd point out that when double-write is over the ONLY thing that gets dropped for the regular pageloads is line 71 to line 74 here, the event. The cost reduction is very small for the bundled vitals

text/0154-web-vitals-as-trace-metrics.md
k-fish: I think this section and table above is a bit unclear, 100 million pageloads = 100 million spans  at traceSampleRate 1.0 which _don't go away_, the 300 million standalone spans (is this true on earlier versions of js sdk? LCP was included previously) are primarily what are being replaced

> Even so, at metric pricing the 500M items still 
cost less than 30M sampled spans while capturing 100% of emissions.

They're still paying for 100M spans (or 10million pageload spans) even after single write, removing the 'double' write only removes the event from those, not the spans. It should be more clear that the standalone spans are the ones being removed.

gggritso:
Only a few things to add to David and Kevan's points:

1. The "SDK Only" option doesn't even need to include the query layer aspect, IMO. We can have a project flag that checks if the project received vitals-as-metrics, and show them metrics-backed widgets on the Web Vitals dashboards. We can take their current Web Vitals dashboard and "freeze it" so they can check historic data
2. As Kevan pointed out, some Dashboard Widgets might be plotting Web Vital measurements on top of other span attributes. Since a single widget can only plot one dataset, it might be impossible to migrate those widgets, they'd have to be split into two widgets, or something else wacky. That said, it would become possible to plot Web Vitals alongside other _metrics_ which is cool
3. The Performance Score section needs a bit more eyes, I think. At least some aspect of performance score calculation is done at query time now and that would need a plan, too
4. I'm not sure if Web Vitals spans go through any enrichment in the buffer. As far as I can tell they do not, if that's true then no problem 👍🏻 
5. What's the plan for Mobile Vitals? Will they also become metrics?

Overall using metrics for Web Vitals makes a lot of sense to me, especially because of the sampling.

text/0154-web-vitals-as-trace-metrics.md
logaretm: So the complication here is we have a fork:

- Right now, yes 4 vitals ride on the pageload span
- Next major with span v2, the default will be only FCP and TTFB, since LCP and CLS will be their own spans. So this changes the math, I will try to calculate for both. But for v11, doing them as metric will cost less than doing them as v2 spans.

So as things currently stand, customers will pay more with v11 anyways. I will get some numbers on that as we discussed.

text/0154-web-vitals-as-trace-metrics.md
logaretm: I'm going to do some manual analysis for this, and inspect the payloads and update the numbers. AI cannot be trusted here 🤦‍♂️

text/0154-web-vitals-as-trace-metrics.md
logaretm: I will clarify this up, the math here is for the v11 span streaming defaults which will send out (INP, CLS, LCP) as spans, which is +2 more than what we have today. So each 1 pageload span will have 3 more in v11, in today's math you are correct. Only INP is standalone.

The RFC currently assumes v11 future behavior rather than today's actual behavior. I will clear both cases up wherever it is relevant.

text/0154-web-vitals-as-trace-metrics.md
logaretm: I re-ran the numbers again against actual payloads and no AI estimation. Per telemetry type it is true that metrics are lighter than spans. But here, given that TTFB/FCP will be converted to metrics as well, it doesn't become a win per pageload.

I have updated the numbers in a new section, and no longer framing it as a win.  Sorry for the confusion on this.

logaretm:
@gggritso I added your point (1) to the RFC section. Could be a good alternate plan if we want to just do it for newer projects.

> What's the plan for Mobile Vitals? Will they also become metrics?

No plans for that, I raised this RFC in the SDK meeting roundtable and nothing was discussed about that. So this is isolated to browser SDK only at the moment, and only web vitals.

text/0154-web-vitals-as-trace-metrics.md
bcoe: If @logaretm's guesstimates on cost were correct, it seemed like we could potentially consider a year of double writing, would need to double check all our napkin math.

logaretm: I can run that and update the numbers for the absorbed listing price costs.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,384 @@
...
+| LCP   | `span.op == "pageload"` + `browser.web_vital.lcp.value` attribute present                                | `browser.web_vital.lcp`  | `span.attributes["browser.web_vital.lcp.value"]`  | `millisecond` | `browser.web_vital.lcp.element`, `.id`, `.url`, `.size`, `.load_time`, `.render_time` |
+| CLS   | `span.op == "pageload"` + `browser.web_vital.cls.value` attribute present                                | `browser.web_vital.cls`  | `span.attributes["browser.web_vital.cls.value"]`  | `none`        | `browser.web_vital.cls.source.1`, `.source.2`                                         |
+| INP   | `span.op == "ui.interaction.{click,hover,drag,press}"` + `browser.web_vital.inp.value` attribute present | `browser.web_vital.inp`  | `span.attributes["browser.web_vital.inp.value"]`  | `millisecond` | _(none)_                                                                              |
```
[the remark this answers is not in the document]
logaretm: No longer the case in the SDK, LCP/CLS will be purely attributes.

text/0154-web-vitals-as-trace-metrics.md
```diff
@@ -0,0 +1,384 @@
...
+
+One question to answer is: is it possible to **Migrate alerts and Discover queries to the metrics dataset?** We can provide auto-migrate known patterns (e.g. `measurements.lcp` -> `browser.web_vital.lcp`) or provide migration tooling. This depends if users make use of custom attributes on those spans or if they combine filters on other attributes on their saved queries.
+
+### Provenance-based billing
+
+Each converted metric carries a provenance or a meta attribute that determines how it's billed:
+
+| Attribute              | Value              | Meaning                                    | Billed as                                                |
+| ---------------------- | ------------------ | ------------------------------------------ | -------------------------------------------------------- |
+| `sentry.metric.source` | `"span"`           | Relay converted this from a web vital span | Free during double-write, as Metric when cutover happens |
```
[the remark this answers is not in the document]
logaretm: Addressed in the next paragraph...

